PR #1, opened 2020-06-07
Create UI_switch_to_selected_part.lua

CJGarciaMusic:
This is a script that will switch to the first linked part that contains the top staff in the selected region. If you are viewing a part and run the script, you will be returned to the score.


rpatters1:
The script should use part:ViewInDocument() rather than SwitchTo(). SwitchTo is for switching efficiently under the covers to a different part and then switching back. It does not fully update the U.I.

Another enhancement (which I could work on myself at some point) is scrolling the current selection into view. Finale is supposed to be doing it for, but somehow it fails about half the time when switching from part to score view, unless update layout happens. This script might be able to address that.

PR #2, opened 2020-06-07
Use ViewInDocument instead of SwitchTo

rpatters1:
Here's an update for ViewInDocument. (I only know about this stuff because I know what how it's has to call the PDK. I am a complete newbie with both lua and the JW Lua class framework.)

### DIFF
--- a/UI_switch_to_selected_part.lua
+++ b/UI_switch_to_selected_part.lua
@@ -2,7 +2,7 @@ function plugindef()
     finaleplugin.RequireSelection = true
     finaleplugin.Author = "CJ Garcia"
     finaleplugin.Copyright = "© 2020 CJ Garcia Music"
-    finaleplugin.Version = "1.0"
+    finaleplugin.Version = "1.1"
     finaleplugin.Date = "6/5/2020"
     finaleplugin.CategoryTags = "UI"
     return "Switch To Selected Part", "Switch To Selected Part", "Switches to the first part of the top staff in a selected region in a score. Switches back to the score if viewing a part."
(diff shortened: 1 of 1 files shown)


rpatters1:
I went ahead and added the scroll-into-view feature. I wish I had thought of this script myself about 2 months ago. Woulda save a *ton* of time and frustration.

rpatters1:
I removed the commented print statements. On comments in general, I think we should be putting more, not less. I guess I don't agree that comments are bad style. In fact, my experience has been the opposite, at least on other platforms like c++.

Nick-Mazuk:
I wouldn't necessarily say comments are bad style for production code, but code should have as few comments as possible. Ideally, the code itself is readable so there's no confusion on what's going on. Therefore, you won't need comments.

The main issue for comments is if you edit the code later, you need to also update the comment. If you don't, the comment says something different than the code. In my opinion, comments should only be used as a last resort when it comes to readability.

rpatters1:
Okay. I don't believe in self-commenting code. In 40 years of coding, I've
never encountered a comment I regretted, especially in my own code. But
ymmv.

The comments I left in this script explain why the code has to do what it's
doing. I don't think the code itself explains it.

rpatters1:
I would agree that overuse of comments is probably not good form. Certainly not replicating documentation that exists elsewhere (such as how to use the JW Lua framework). But where I think they are useful is in documenting Finale's idiosyncrasies and outright bugs that cause plugins to have to do extra work. (In this case it was the need to use ViewInDocument instead of SwitchTo, or the fact that Finale fails to automatically convert the selected region to the part when switching to the part.)

Nick-Mazuk:
I think it's sounding like perhaps the best thing will be to not regulate commenting at all. If there seems like there's something with comments that's causing a problem in the future, we can deal with it then.

But for now, you have good reasons for commenting and I'm not particularly adamant about eliminating all comments.

PR #3, opened 2020-06-08
replicate BeamSelection plugin in lua

rpatters1:
These two script replicate the functionality of Patterson Plugins Beam Selection. One of them beams everything in the selection. The other unbeams everything in the selection.

It would be nice to move the entry_get_next_same_v() function into a common file and import it into each. I need some Lua guidance on the best way to do that. (I'm still a Lua newbie.)


rpatters1:
I just realized I used camel case for some of my local variables. Dang, old
habits die hard. I'll do better next time.

PR #4, opened 2020-06-08
Added a Few Scripts + A Library Script

Nick-Mazuk:
This pull request has two parts:

- 9 scripts
- Start of a reusable function library

Specifically, please check that `pitch_entry_double_octave_down.lua` and `pitch_entry_double_octave_up.lua` work.

These two scripts use `library.lua` for the helper function `change_octave`. These scripts work fine on my setup and theoretically should work wherever the local git repo is. But if you could check to see if the scripts work, that would validate that the import process isn't specific to my machine.

P.S. The function library is a JW Lua file that can house many common functions. That way, we can reuse previously written code for newer JW Lua scripts, speeding up development time, removing duplicate code, and lowering the barrier to entry for JW Lua.

### DIFF
--- a/chord_toggle_visibility.lua
+++ b/chord_toggle_visibility.lua
@@ -1,0 +1,20 @@
+function plugindef()
+    finaleplugin.RequireSelection = true
+    finaleplugin.Author = "Nick Mazuk"
+    finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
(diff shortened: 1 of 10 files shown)


rpatters1:
An alternative would be to create a library folder and add library functions to it in individual .lua script files. That would allow us to pick and choose the library functions we needed.

Nick-Mazuk:
I like the idea of having a `library` folder with several individual scripts. Let's do that!

PR #13, opened 2020-06-12
Organized Standard Library

Nick-Mazuk:
Also:

- Added library use to README.md
- Updated scripts already using library functions
- Added a few more library modules to flesh out (e.g., transposition—I have many functions to go there)
- Maybe a few more scripts?


rpatters1:
Do we know for sure that forward slash path names work on Windows?

Nick-Mazuk:
I don't know, nor do I know how to test it.

Bit it looks like there's a Lua function AssureEndingPathDelimiter() for finale.FCString() that adds the correct path delimiter for the OS. Perhaps we can use that?

Like perhaps:

local path = finale.FCString()
path:SetRunningLuaFolderPath()
path:AssureEndingPathDelimiter()
path.LuaString = path.LuaString .. library
path:AssureEndingPathDelimiter()
package.path = package.path .. ";" .. path.LuaString .. "?.lua"
Seems like that's more complicated than it can be.

rpatters1:
Before we add all the clunky code everywhere, let me test it on windows. I
think it will work, but we need to be certain. Also, there's a (slightly)
cleaner way to get the delimiter if it doesn't work.

Nick-Mazuk:
Sounds good.

If there's a more elegant way that works, I'd opt for that.

rpatters1:
I just tested on Windows, and it works fine. Apparently Windows has
accepted '/' as a path delimiter since DOS 2.0, as long as it wasn't on the
command line.

rpatters1:
TIL. I worked on DOS and Windows as a day job from 1984-2011 and never
knew. Dang.

Nick-Mazuk:
I never knew that either!

Interesting how Windows accepts different delimiters depending on where you use them, though.

rpatters1:
According to the Google, Windows using '\' was an accident of history due
to the DOS command line treating '/' as an option flag. (Instead of the
more familiar '-'.) So only the command line parser ever cared, and then
only when it was ambiguous. Google says you could always "CD ./subdir" for
example, because that isn't ambiguous.

PR #23, opened 2020-06-20
new script to adjust measure numbers for key, time, and open repeat bars

rpatters1:
New function to shift measure to the left as needed to compensate for key signatures, time signatures, and open repeat bars. It works best if you select only the staff or staves that actually have measure numbers. There are comments included to help users refine it to their taste if they so desire.


rpatters1:
Hold off approving this for a minute. I've discovered some issues. (Also, I don't understand the build checks well enough to know why they are failing.)

rpatters1:
Closing this for now. It needs more work.

PR #37, opened 2021-03-01
Standalone Hairpin adjustments added

CJGarciaMusic:
Just pulled the hairpin alignment/dynamic adjustments out for a stand alone script.

### DIFF
--- a/Standalone_Hairpin_Adjustment.lua
+++ b/Standalone_Hairpin_Adjustment.lua
@@ -1,0 +1,350 @@
+function plugindef()
+    finaleplugin.RequireSelection = true
+    finaleplugin.Author = "CJ Garcia"
+    finaleplugin.Copyright = "© 2021 CJ Garcia Music"
+    finaleplugin.Version = "1.0"
+    finaleplugin.Date = "2/29/2021"
+    return "Hairpin and Dynamic Adjustments", "Hairpin and Dynamic Adjustments", "Adjusts hairpins to remove collisions with dynamics and aligns hairpins with dynamics."
+end
+
+function vertical_dynamic_adjustment(region, direction)
+    local lowest_item = {}
+    local staff_pos = {}
+    local has_dynamics = false
+    local has_hairpins = false
+    local arg_point = finale.FCPoint(0, 0)
+
+    local expressions = finale.FCExpressions()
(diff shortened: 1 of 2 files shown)


CJGarciaMusic:
I'm feeling a little silly, but I'm not sure what this failed merge means/is about. Anything I can do to fix it?

rpatters1:
Don't worry about the merge fail. I think it is an incomplete test setup in the repo. This script is great. It's covering all the bases. I have two small nitpicks.

1. Should the filename be all lowercase? That seems to be the convention most of the rest of them are using.
2. I don't think we should be finding Dynamics by comparing the category name to a string. This is not compatible with any language other than English. Instead, should it not compare the category id to the  predefined category id `finale. DEFAULTCATID_DYNAMICS`? As far as I can tell, all Finale files are guaranteed to have this category.

rpatters1:
I had another thought, because it occurred to me you may be comparing the category name to a string to include other categories besides just the default Dynamics category. In the PDK the category struct has a field called `categoryType` which (if I recall correctly) is the default category ID from which the user created the category. This would be the ideal way to identify categories created from the Dynamics category. Unfortunately (and surprisingly) JW Lua does not appear to export this field. :-/

So if the design goal is to include user-defined categories with dynamics, for now we could find the default category by id and also find any user-defined categories containing the string. (At least until—he says hopefully—we can update JW Lua.)

Nick-Mazuk:
> Don't worry about the merge fail. I think it is an incomplete test setup in the repo. This script is great. It's covering all the bases.

This is spot on. The test setup was to help create the "sitemap" for the installer script (https://github.com/Nick-Mazuk/jw-lua-scripts/issues/17), but it's not finished. I'm going to go ahead and remove the test because there are better ways to do the same thing now, and then it will prevent further confusion.

I do agree with @rpatters1's suggestions. Other than that, it looks great!

CJGarciaMusic:
Thanks for the input all! I'll do an "or" statement for both options and commit a new version shortly!

CJGarciaMusic:
Okay, script and filename updated!

Nick-Mazuk:
@CJGarciaMusic, looks good to me! @rpatters1, I'm going to go ahead and merge it since I believe it uses your suggestions as you noted.

PR #38, opened 2021-03-01
Remove workflow that indexes the scripts

Nick-Mazuk:
Removing the workflow script as mentioned in https://github.com/Nick-Mazuk/jw-lua-scripts/pull/37#issuecomment-788050390


rpatters1 (approved):
looks good

PR #40, opened 2021-03-01
right hairpin now aligns with end of the note

CJGarciaMusic:
If there is not a dynamic attached to the right end segment, the hairpin will extend to the end of the notehead instead of the beginning of the notehead.
Also, I applied Roberts suggestion of single bool expressions, all seems to work still.


rpatters1 (approved):
looks good

PR #41, opened 2021-03-03
Rgp/hairpin adjustments support justification

rpatters1:
I modified C.J.'s script to take into account horizontal alignment and justification of the expr. def. This entailed creating two new libraries to support figuring all that stuff out.


rpatters1:
I just discovered there is more work to be done, so this pull request is a no-go

PR #45, opened 2021-03-03
refinements to gitignore

rpatters1:
I got a little hasty with the review, so I just fixed the .gitignore myself.


Nick-Mazuk (approved):
Looks good. Definitely much clearer

PR #46, opened 2021-03-03
Rgp/hairpin adjustments support justification

rpatters1:
Added support for expression justification and alignment and extensive debugging.

I'll do the text-file script settings as a separate PR.


Nick-Mazuk (approved):
Not familiar with the hairpins algorithm, but it seems like it should work. Approving, with the assumption that CJ will also find it good.

PR #47, opened 2021-03-05
Rgp/parameters from text file

rpatters1:
Here is the configuration file feature. It has the following features:

- It is not dependent on format. Each parameter is named with the same name as in the script. If the script changes the parameter name, that value in the config file will be ignored.
- Parameters may be listed in any order in the config file, and they may also be omitted.
- Config files must be in a sub-directory of the running Lua path called `script_settings`.
- Config files may contain comments using `--` just as in the Lua language.
- Currently only numbers and boolean values (`true` or `false`) are supported. Strings can be a future enhancement.
- A config file is always optional. In the case when it is missing, the default values are used.

Here is a sample config file for the `standalone_hairpin_adjustments` lua script:
```

-- Comments can be here
--
--

left_dynamic_cushion = 24		--evpus
right_dynamic_cushion = -24		--evpus
extend_to_end_of_right_entry = false

```

### DIFF
--- a/library/configuration.lua
+++ b/library/configuration.lua
@@ -1,0 +1,83 @@
+-- A collection of helpful JW Lua scripts to retrieve parameters from a text file
+-- Simply import this file to another Lua script to use any of these scripts
+--
+-- Author: Robert Patterson
+-- Date: March 5, 2021
+--
+-- This library implements a text file scheme as follows:
+-- Comments start with "--"
+-- Leading and trailing whitespace is ignored
+-- Each parameter is named and delimited by a colon as follows:
+--
+-- <parameter-name> = <parameter-value>
+--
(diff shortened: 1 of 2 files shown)


CJGarciaMusic (approved):
It'll take a little bit to untangle this for JetStream, but it looks like it will function well with the configuration library.

rpatters1:
If you decide to port this approach to JetStream, it has the feature that you could combine all your parameters for all JetStream script into a single file. Or you could put them in multiple files. It's extremely flexible. The library matches the the values based on the parameter names. The script can have names that aren't in the file, and the file can have names that aren't in the script. The function doesn't care.

This is all possible due to how lua makes pseudo structs that are actually tables with string keys. I wasn't expecting it to be as straightforward as it turned out to be.

PR #49, opened 2021-03-08
Rgp/notehead delete duplicates mmrest adj

rpatters1:
This PR adds two new scripts. One that removes duplicate notes from a chord and another that lengthens multimeasure rests to allow horizontal room for tempo markings.


Nick-Mazuk (approved):
Looks good

PR #50, opened 2021-03-10
Rgp/articulation autoposition rolled chords

rpatters1:
These changes were part of a failed attempt to create a script to autoposition rolled chord articulations. The attempt failed because the PDK Framework does not provide access to the PDK fields that specify the horz and vert extensions for cloned marks. (I am going to do this as a plugin.)

However, I would like to keep the changes to the libraries I made. I added some useful functions and fixed some other issues as well.


Nick-Mazuk (commented):
If I understand correctly, you were working on a script for rolled chords and created several useful libraries for that script. Though that script didn't turn workout because limitations of the PDK Framework, you're leaving the libraries for future use. Then, you used the libraries to abstract some of the logic from `standalone_hairpin_adjustment.lua` and `mmrest_widen_to_tempo_mark.lua`.

That's also what the code looks like to me.

I'm not a fan of leaving unused code in a repo because it can make maintaining the repo more time consuming in the future. But I can see the future utility of several of the unused libraries. So perhaps we could add a comment on the currently unused functions saying "this function is unused. It can be removed in the future. If you use this function, remove this comment" or similar. What are your thoughts?

These are the specific functions I'm seeing that are unused:

- note_entry.get_music_region
- note_entry.get_evpu_notehead_height
- note_entry.get_top_note_position
- note_entry.get_bottom_note_position

rpatters1:
You have correctly analyzed the situation. I would add that the unused functions are fully debugged, including optional parameters (if any), which is why I would hate to abandon them. The only issue I have with the comment is that there is no way to enforce its removal if the function becomes used.

Also, I do hope to finish implementing the rolled chord script some day, but that will require rebuilding JW Lua with new features in the PDK Framework.

rpatters1:
P.S. Perhaps we could soften the comment to "may not be used" and "you may remove after confirming that it isn't used".

Nick-Mazuk:
> P.S. Perhaps we could soften the comment to "may not be used" and "you may remove after confirming that it isn't used".

I like this suggestion.

This is one of those times, though, that I wish Lua had better developer tools. At least in the JavaScript world, there are tons of developer tools to help with things like this, including ones that automatically find unused code.

rpatters1:
Okay, I added the comments.

Nick-Mazuk (approved):
Looks good

PR #51, opened 2021-03-15
Rgp/automatic jete

rpatters1:
This script implements automatic jeté notation as given on page 404 of Elaine Gould, "Behind Bars".

I ended up using a couple of the "unused" functions in note_entry.lua, so I changed the comments accordingly.


Nick-Mazuk (approved):
Looks good! Seems like a useful script as well

PR #52, opened 2021-03-17
Rgp/multifont support for text width calc

rpatters1:
Added multifont support to enigma_string.calc_advance_text_width function. This will allow correct width calculation for expressions that contain font type, size, style changes such as **pp** _subito_.


Nick-Mazuk (approved):
This seems really useful

PR #53, opened 2021-03-20
Expressions assignment revisions

rpatters1:
- Revised expression assignment scripts to recognize score-only/part-only assignment.
- Added scripts for score-only and part-only assignment.


Nick-Mazuk (approved):
Looks good

PR #54, opened 2021-03-21
Add fix indent from doc settings script

rpatters1:
Recreate the `Fix Indent` function of the `JW New Piece` plugin. The reason for this script is that `JW New Piece` uses the indentation of system 1 for the other first systems, and it assumes 0 for non-first systems. This script gets those values out of Page Format For Score or Page Format For Parts, depending on whether we are currently viewing score or part.


Nick-Mazuk (approved):
Looks great

PR #57, opened 2021-03-24
Rgp/transposition rewrite

rpatters1:
I have gone ahead and implemented math-based transposition in the transposition library. This pull request is fully backwards compatible with the previous note-name-based transposition. However, it also fixes at least one bug I found in the old transposition. In C Major, transposing F up a perfect fourth resulted in B# rather than Bb. The new version correctly produces Bb.

I envision further enhancements to

- Provide an option to simply the note spelling
- Attempt to compensate if the alteration exceeds +/- 7. (This issue is quite common in larger EDO microtonal systems.)

I also intend to provide a U.I. front end. It will be mainly useful for users of microtonal key signatures, but it will include a simplified spelling option and a U.I. to transpose arbitrarily by number of steps with simplified spelling. (For standard key signatures this means by number of half steps.) Some of this should be of interest to users of standard key sigs as well.

### DIFF
--- a/library/configuration.lua
+++ b/library/configuration.lua
@@ -11,7 +11,13 @@
 --
 -- <parameter-name> = <parameter-value>
 --
--- Parameter values must be numbers or booleans or strings.
+-- Parameter values may be:
+--      Strings delimited with either single- or double-quotes)
+--      Tables delimited with {}
+--      Booleans (true or false)
+--      Integers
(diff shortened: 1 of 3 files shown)


rpatters1:
I'm canceling this pull request and will submit another with the feature-complete transposition including new scripts that exploit it.

PR #59, opened 2021-03-27
Rgp/transposition extended

rpatters1:
This pull request includes a rewrite of the transposition library that includes the ability to simplify spelling and also that supports custom key signatures (microtone systems) by means of a config.txt file. (Unfortunately, the PDK Framework does not provide the necessary APIs to read this information directly from the Finale document.) All of the existing scripts that use the transposition library have been re-tested with the new version.

In addition, it provides new scripts for
- Transpose Enharmonic Up
- Transpose Enharmonic Down
- Transpose By Step (half step in standard keys, microtone step in microtone systems)
- Transpose Chromatic

All of these support the microtone system supplied in the config.txt file if there is one (and if the key of the document is a custom key signature.)

Finally, it fixes the following pre-existing bugs:
- Transposition up a 4th from F (in C Major) was producing B# instead of Bb.
- String Harmonics 5th was transposing to the incorrect octave for the original sounding pitch.


rpatters1:
Here is a sample custom_key_sig.config.txt file for the most common microtone systems.

[custom_key_sig.config.txt](https://github.com/Nick-Mazuk/jw-lua-scripts/files/6215935/custom_key_sig.config.txt)

Nick-Mazuk (approved):
Looks good!

PR #60, opened 2021-03-31
Bug Fixes and Configs

rpatters1:
This pull request incorporates various bug fixes and minor enhancements and code cleanup.

**String Harmonics**
- Add support for breve noteheads.
- Use CalcStemUp instead of CalcStaffPosition to determine direction of offsets for whole and breve noteheads (much more accurate).
- Fixed issue where non-floating rests could cause a runtime error.
- Add config file support for open and closed harmonic noteheads (but default is all-open, per Elaine Gould).
- Add config file support for whole and breve offsets and resize percentage.

**Double Octave Up/Down**
- Now supports chords in addition to single notes.

**Rotate Chords Up/Down**
- Replace MIDI note-based logic with logic that takes into account note spelling.
- Add support for octave spanning chords (such as the extremely common 4-note triad voicings found in keyboard music).

Other minor cleanup.

### DIFF
--- a/library/note_entry.lua
+++ b/library/note_entry.lua
@@ -199,4 +199,15 @@ function note_entry.duplicate_note(note)
     return new_note
 end
 
+function note_entry.spans_number_of_octaves(entry)
+    local top_note = entry:CalcHighestNote(nil)
+    local bottom_note = entry:CalcLowestNote(nil)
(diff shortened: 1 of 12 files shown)


Nick-Mazuk (approved):
Looks good

PR #61, opened 2021-04-04
Add automated library documentation

Nick-Mazuk:
This PR does 3 things.

1. Adds a GitHub Action to automatically create documentation for the library scripts on each push to main
2. Adds Dependabot, so a PR will be created each time the GitHub action is updated
3. Converts current documentation comments in the library scripts into ExpLua

And merging this PR into the master branch will start creating the documentation. The docs will be added to `docs/library`. This way we can add more documentation about general development in the repo to hopefully onboard new users (coming tomorrow, hopefully).

Related issue: https://github.com/Nick-Mazuk/jw-lua-scripts/issues/55

Keeping the issue open as there's still more to do.

---

I know we were talking about using [this GitHub Action](https://github.com/marketplace/actions/lua-docs-generator) to create the documentation, however, I could not figure out how to update the Python code to parse ExpLua. And since there were no tests, I wasn't confident that any changes I'd make would result in regressions for current users. So I created a new GitHub action that's used in this PR: https://github.com/marketplace/actions/generate-lua-docs

There are a few, slight differences between this action and the Python-based action. The main differences are that this action assumes "one documentation file per Lua file", and this action parses ExpLua. It's also much better (uhmm…) documented. 

You can see exactly how comments are parsed [here](https://github.com/Nick-Mazuk/lua-docs-generator/blob/main/documentation.md). There are also a few [examples](https://github.com/Nick-Mazuk/lua-docs-generator/tree/main/test-files).

The action is new, but it is tested. If bugs are found, I can easily fix them. And since there are tests for every feature, there should not be regressions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  # Enable version updates for npm
+  - package-ecosystem: "github-actions"
+    # Look for `package.json` and `lock` files in the `root` directory
+    directory: "/"
+    # Check the npm registry for updates every day (weekdays)
+    schedule:
+      interval: "daily"
--- a/.github/workflows/generate-docs.yml
+++ b/.github/workflows/generate-docs.yml
@@ -1,0 +1,23 @@
+name: Generate docs
+
+on:
+  push:
+    branches:
+      - "master"
+
+jobs:
+  test-push:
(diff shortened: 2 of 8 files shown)


rpatters1 (approved):
This looks great.

PR #72, opened 2021-04-20
Move scripts into `src` directory [breaking]

Nick-Mazuk:
This PR moves all the scripts into a `/src` directory. This should lead to better organization in the future, and will make it easier to create downloadable versions of each script. Discussed in https://github.com/Nick-Mazuk/jw-lua-scripts/issues/70

**Breaking change**: Anyone who currently uses this repo to run these scripts will have to "reregister" them with JW Lua.


rpatters1 (commented):
Does `.gitignore` need to be changed as well (now that `script_settings` will be in `src`)?

(I'm not sure it does, but it's worth checking.)

Nick-Mazuk:
@rpatters1  Just checked, and no, the `.gitignore` does not need to be changed.

rpatters1 (approved):
Looks good.

rpatters1 (approved):
looks good

PR #74, opened 2021-04-21
Rgp/add function decorations to lib files

rpatters1:
I believe this takes care of all the library functions I created. (Plus a few more.)


Nick-Mazuk (approved):
Looks good!

PR #79, opened 2021-05-17
Add files via upload

jwink75:
dynamics_move_above_staff.lua moves dynamics (expressions whose category is named "dynamics" and hairpins) above a staff, for vocal music. The script attempt to place all dynamics on a system so that they will clear the highest entry for that staff on that system.


Nick-Mazuk:
Looks great, Jake!

PR #82, opened 2021-06-18
Workaround for bug in Finale json files

rpatters1:
It turns out there's a bug in the Finale json files. One of the fields is incorrectly implemented. When I finally tested with Bravura I ran into it. This pull request works around the bug, but I expect Finale will fix the json files at some point.


Nick-Mazuk (commented):
Looks good!

Nick-Mazuk (approved):

PR #83, opened 2021-06-22
more SMuFL support

rpatters1:
This pull request
* moves smufl font checking to general_library
* adds support for the user-level directory for json files
* adds support for SMuFL diamond noteheads without the need of using a config file
* adds support for SMuFL staccato dot in automatic_jete.lua (without the need of a config file)


Nick-Mazuk (approved):
Looks good to me

PR #84, opened 2021-06-23
add lyrics baseline nudgers

rpatters1:
add lyrics baseline nudgers similar to Nick's expression baseline nudgers. These nudge all verse/chorus/section lyrics at once.


Nick-Mazuk (approved):
Looks good to me

PR #86, opened 2021-07-08
change staff nudgers to iterate on slots instead of staff numbers

### LINKED ISSUE #85: [BUG] Move Staff Up and Move Staff Down do not work if staff numbers are out of sequence
The Move Staff Up and Move Staff Down scripts appear to make a flawed assumption that staff numbers occur in order and without gaps. Neither is the case. Staff numbers can occur in any order and not every number has to be there. (The baseline nudger scripts also make this flawed assumption, but I am already preparing a fix for those, among other bug fixes and enhancements.)

It appears that the correct way to iterate staves with JW Lua is by means of slots in FCMusicRegion or FCSystemStaves per system (using a counter for the slot number). I am willing to fix the Staff nudger scripts, but I can't quite make out how they are supposed to function. Are they supposed to work like drag (where all the staves below move) or option-drag (where staves on the system below the selection stay put)?

I am attaching a sample file with the staff numbers out of sequence.

[staves-out-of-sync.musx.zip](https://github.com/Nick-Mazuk/jw-lua-scripts/files/6780586/staves-out-of-sync.musx.zip)

rpatters1:
This pull request contains the following changes:

- Changes staff nudgers to iterate from FCSystemStaves instead of by staff number
- Iterates all selected systems
- Renames an enclosing function to match its file name (per our conventions)
- Ignore top staff in each system. The Finale UI appears to keep those at zero, so we should probably do the same.

The behavior of the functions (other than being multisystem) is compatible with the previous versions for documents where the staff numbers were in order and consecutively numbered.

resolves #85

### DIFF
--- a/src/staff_move_down.lua
+++ b/src/staff_move_down.lua
@@ -2,7 +2,7 @@ function plugindef()
     finaleplugin.RequireSelection = true
     finaleplugin.Author = "Nick Mazuk"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "1.0"
+    finaleplugin.Version = "1.0.1"
     finaleplugin.Date = "June 20, 2020"
     finaleplugin.CategoryTags = "Staff"
     finaleplugin.AuthorURL = "https://nickmazuk.com"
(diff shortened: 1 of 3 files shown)


rpatters1:
If there is an issue with this pull request, let me know. I've got a couple lined up behind this one, but I thought I'd wait for merging before proceeding with them. (Though it isn't strictly necessary.)

Nick-Mazuk (approved):
Looks good to me, sorry for the delay

PR #88, opened 2021-07-22
Rgp/nudger bug fixes and enhancements

rpatters1:
Adds baseline nudgers for lyrics and expression below (as well as above, which was already there). Also changes the existing ones to use slots to iterated on instead of staff numbers.


Nick-Mazuk (approved):
Looks good to me

PR #89, opened 2021-07-26
Bundle scripts with library dependencies

Nick-Mazuk:
This PR adds a workflow to create a downloadable version of every script. This means that all library files are inlined into the script—even if one library imports another. These scripts will automatically be built on every commit to the master branch. No extra work needed on the developer end.

The downloadable versions of the scripts will automatically be commited to the `/dist` directory. This means we can keep the same developer workflow, but script users will now have the option to download a single file instead of the entire repo.

I did not test the final output on all ~90 scripts in the repo, but I tested several and they seem to work without issue.

Related: https://github.com/Nick-Mazuk/jw-lua-scripts/issues/70

P.S. Once this is merged, I plan on creating a new workflow that parses the metadata from the `pluginDef` to make the scripts searchable on and downloadable from the JW Lua Scripts site.


rpatters1 (approved):
It looks good to me. Is there a way to prevent checkins that modify files in the /dist directory? Also, have you checked how it will work with script_settings config files? I suppose there will need to be a script_settings folder in the /dist directory on the local machine.

Nick-Mazuk:
> Is there a way to prevent checkins that modify files in the /dist directory?

The `/dist` directory is cleared then recreated after every commit, so if anyone does modify it, their modification will be undone once they commit to master. This way, the `/dist` directory is always a reflection of the corresponding code in the `/src` folder. Does this do what you're looking for?

> Also, have you checked how it will work with script_settings config files? I suppose there will need to be a script_settings folder in the /dist directory on the local machine.

That is correct.

rpatters1:
That should be fine with respect to /dist.

After looking at some of the generated files, I realize there could
potentially be name conflicts between local variables inside libraries. I
am not aware of any that actually exist at the moment, though.

Nick-Mazuk:
> After looking at some of the generated files, I realize there could potentially be name conflicts between local variables inside libraries. I am not aware of any that actually exist at the moment, though.

Yeah, there could be, though it should be minimal. Right now, when you import a library, the original variable name is used. So for instance:

```lua
local library_variable = require("library.file_name")
```

The output will use this same `library_variable` name. This shouldn't produce naming conflicts unless the original script had a naming conflict.

However, if a library imports another file, that could cause naming conflicts. I could update it, though, to give these indirect dependencies a random name to reduce potential naming conflicts significantly. Though that would significantly increase the chance for bugs as doing a simple find-and-replace for the variable names could be problematic.

This also means that it's possible for multiple versions of the same library to be bundled into the script, but that shouldn't be an issue since file size isn't a concern at this point (I think).

PR #92, opened 2021-07-30
add cluster scripts

Nick-Mazuk:
Adds the cluster scripts created by @jwink75. These are the updated versions he just sent me via email.


rpatters1 (commented):
I notice a number of the functions use Pascal case instead of snake case. Shouldn't they be consistent with the rest of the repo?
Also, I'd like to see the font and chars for the special noteheads be configurable, but that isn't a show stopper for me.

Nick-Mazuk:
> I notice a number of the functions use Pascal case instead of snake case. Shouldn't they be consistent with the rest of the repo?

You're totally right. I totally missed that when reviewing the scripts… guess I just got used to writing JavaScript for the actions this week.

Will fix later this evening.

rpatters1 (approved):
Looks good

PR #95, opened 2021-08-02
Rgp/bug fixes 1

rpatters1:
This is a set of cumulative bug fixes I've found over the last month:

- get the correct visible measure number region info
- work around Lua bug in StaffGroup.DrawBarlineMode (see comment in code)
- include beginning of system when adjusting measure numbers for time/key/repeat
- do not attempt to adjust the measure number if no measure number region exists for that measure
- when looking for the SMuFL json file, search in the order prescribed by the spec (user then system)


Nick-Mazuk (approved):
Looks good to me

PR #96, opened 2021-08-17
add chord accidental adjustment scripts

Nick-Mazuk:
Adds two scripts that Michael McClennan email me that adjust accidentals on chord symbols vertically.

For instance, if with the chord tool, you add a Db7 chord then run the `chord_accidental_adjust_up` script, the `b` will move up a tiny bit.

According to Michael, it works well adjusting these accidentals after batch-replacing fonts.

### DIFF
--- a/src/chord_accidental_adjust_down.lua
+++ b/src/chord_accidental_adjust_down.lua
@@ -1,0 +1,30 @@
+function plugindef()
+    finaleplugin.RequireSelection = true
+    finaleplugin.Author = "Michael McClennan"
+    finaleplugin.Version = "1.0"
+    finaleplugin.Date = "August 14, 2021"
+    finaleplugin.AuthorURL = "www.michaelmcclennan.com"
+    finaleplugin.AuthorEmail = "info@michaelmcclennan.com"
+    finaleplugin.CategoryTags = "Chord"
+    return "Chord Accidental - Move Down", "Adjust Chord Accidental Down", "Adjust the accidental of chord symbol down"
+end
+
+function chord_accidental_adjust_down()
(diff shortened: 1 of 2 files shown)


rpatters1 (approved):
It would be nice if the `my_distance` value were configurable in each of these scripts. I'm going to go ahead and approve this, but I wish Michael would consider become a direct contributor. This could be an easy task for him to learn about how to make scripts configurable.

Nick-Mazuk:
> It would be nice if the `my_distance` value were configurable in each of these scripts.

Just pushed an update to make it configurable. It is my first time using the configuration library, so if you could double-check my work that would be great. Feel free to merge if it looks good to you.

> I wish Michael would consider becoming a direct contributor. This could be an easy task for him to learn about how to make scripts configurable.

Totally agree. Michael also sent me another script, so I'm going to see if he could try adding it himself. He's done quite a bit with JW Lua and I think he'll find contributing quite valuable (especially with the library).

rpatters1:
If we are following the other configurable scripts, it should be in its own config file, say `chord_accidental_adjust.config.txt`.

That said, as these little config files are starting to proliferate, it might make sense to consider consolidating them all in one master config file. Not for this pr, but as a consideration for future enhancement. I'm on the fence about which is the right way to go.

Nick-Mazuk:
Totally just forgot to replace the file name when I copy-pasted the config code. Fixed! I'll go ahead and merge.

> That said, as these little config files are starting to proliferate, it might make sense to consider consolidating them all in one master config file. Not for this pr, but as a consideration for future enhancement. I'm on the fence about which is the right way to go.

Definitely something to consider.

PR #99, opened 2021-08-20
Rgp/bug fixes 2

rpatters1:
- use correct text metrics call for calculating text width
- move all calls of DeleteNote to a library function, so that we can also delete associated notehead modifications
- add documentation comments for library/notehead.lua


Nick-Mazuk (approved):
Looks good to me. Good idea to move the "delete note" functionality to a library function to make sure all the related items are also deleted.

PR #100, opened 2021-08-22
add script to simplify accidentals

Nick-Mazuk:
Adds a script to simplify double sharps and flats as suggested by Jon Burr in this [Facebook post](https://www.facebook.com/groups/finalescript/permalink/1889108927934030/). Note: this is not a rehash of Utilities > Respell Notes as that also simplifies single sharps and flats if they don't comply with the key signature.

### DIFF
--- a/src/accidental_simplify.lua
+++ b/src/accidental_simplify.lua
@@ -17,9 +17,9 @@ local transposition = require("library.transposition")
 function accidentals_simplify()
     for entry in eachentrysaved(finenv.Region()) do
         local measure_number = entry.Measure
-        local measure = finale.FCMeasure()
-        measure:Load(measure_number)
-        local key_signature = measure:GetKeySignature()
+        local staff_number = entry.Staff
+        local cell = finale.FCCell(measure_number, staff_number)
+        local key_signature = cell:GetKeySignature()
 
         for note in each(entry) do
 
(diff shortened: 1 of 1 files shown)


rpatters1 (approved):
Very clever use of string manipulation and the "simplify spelling" option of chromatic_transpose. I like it.

Nick-Mazuk:
Not going to merge quite yet because I found a bug: if a staff is transposed (e.g., a Bb Clarinet), the `note_string` will still be in concert pitch. Therefore I need to find a new way to determine if a note has a double sharp/flat for transposed instruments.

rpatters1:
Use FCCell::GetKeySignature instead of FCMeasure::GetKeySignature. (You'll have to put it inside the loop.)

Nick-Mazuk:
Thanks @rpatters1! That works flawlessly.

rpatters1:
I just thought of a situation where simplify spelling doesn't get the job done. Consider Bbb in the key of Eb Major. It has a RaiseLower of -1, whereas A-natural has a RaiseLower of +1. Since there is no change in abs value, simplify spelling will not modify the note. You might want to call `transpose.enharmonic_transpose` explicitly. You could follow that with the current call to `chromatic_transpose` to clean up any triple sharp/flat stuff. For enharmonic_transpose, you can pass RaiseLower as the sign value.

Nick-Mazuk:
Good catch, just created a new PR to fix that #101

PR #101, opened 2021-08-23
fix slight bug in simplify accidentals

Nick-Mazuk:
Fixes the bug @rpatters1 noticed in #100

https://github.com/Nick-Mazuk/jw-lua-scripts/pull/100#issuecomment-903368928

Feel free to merge if approved

### DIFF
--- a/src/accidental_simplify.lua
+++ b/src/accidental_simplify.lua
@@ -35,6 +35,7 @@ function accidentals_simplify()
 
             -- checking for 'bb' and '##' will also match triple sharps and flats
             if string.match(note_string, "bb") or string.match(note_string, "##") then
+                transposition.enharmonic_transpose(note, note.RaiseLower)
                 transposition.chromatic_transpose(note, 0, 0, true)
             end
         end
--- a/src/accidental_simplify.lua
+++ b/src/accidental_simplify.lua
@@ -16,12 +16,18 @@ local transposition = require("library.transposition")
 
 function accidentals_simplify()
     for entry in eachentrysaved(finenv.Region()) do
+        if not entry:IsNote() then
+            goto continue_entry_loop
+        end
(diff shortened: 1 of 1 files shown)


rpatters1 (approved):
I'm wondering if we should also skip notes with RaiseLower of zero. I mean, the only situation that would arise is if the double-acci is in the key signature, which seems *really* unlikely. But in that extreme corner case, RaiseLower is not an appropriate value for `direction` to send to `enharmonic_transpose` (for a double flat). So the choices are a) let the chips fall where they may (leave as-is), b) skip any notes with RaiseLower of zero, or c) calculate the correct `direction` value when RaiseLower is zero. I vote b but I am happy with any of the choices. Still, choice c might be appropriate if the key signature contains double-accis *and* the staff is marked Hide Key Sig and Show Accis. It still seems like an extreme corner case to me though. Imma go ahead and approve and let you decide what to do.

rpatters1:
Maybe the "most correct" solution would be to ignore RaiseLower and calculate direction if "bb" then -1 else if "##" then +1.

rpatters1:
But crap if you do that, you'll have to skip calling simplify spelling when RaiseLower is zero. Because simplify will undo your enharmonic transpose. Dang.

Nick-Mazuk:
I went with option B.

There's still the edge case where the key signature may contain double sharps/flats and is hidden, but I think that's a) probably quite rare, b) fixing it would complicate the code quite a bit and make it less maintainable, and c) I'm unsure if it's even desirable to remove double sharps/flats in that case. You might have a ton of switching between sharps and naturals at that point. The better solution would probably be to fix the key signature which is out of scope for this script.

I'd say if someone complains and thinks it's broken, we could come back to it at that time.

PR #110, opened 2021-09-02
[WIP] Add scripts to create ensembles

Nick-Mazuk:
This is a port of my personal scripts to create ensembles based on the default document. That way, you can use a single "template" document for all your scores. It's kind of like the Setup Wizard but with the flexibility of JW Lua.

Currently, I've only added the script for creating trombone choir pieces (I'm a trombone player, after all). However, I also added a decently large library for manipulating items/staves in the score.

This PR is currently a work in progress, but feedback is more than welcome—especially since these commits already form the basis for every ensemble.

### DIFF
--- a/src/library/score.lua
+++ b/src/library/score.lua
@@ -13,7 +13,19 @@ configuration.get_parameters("score.config.txt", config)
 
 local score = {}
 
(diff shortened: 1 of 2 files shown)


rpatters1 (approved):
Looks good to me.

Nick-Mazuk:
Realized I never merged this. So merging.

PR #114, opened 2021-11-01
Rgp prep for lua54 etc

rpatters1:
- Get rid of UserValueInput in favor of FCCustomWindow
- Add comment to show how to do bitwise operation in lua 5.4 for Add Augmentation Dot


Nick-Mazuk (approved):
Looks good to me

PR #116, opened 2021-11-03
Add support for mixins and fluid interface for FC objects

ThistleSifter:
I wrote this about a year ago to simplify the process of writing plugins and to improve code maintainability. I recently stumbled upon this repo again so I thought I'd submit and see what happens. 

### Fluid Mixins

This PR does 2 things:
- Allows mixins to be added to any FC* objects
- Adds a fluid interface for any methods that don't return anything.

### Mixins
By default, it is not possible to override or extend any of the FC* objects in JW Lua because objects that are tied to their underlying C++ implementation are not tables. So through a little bit of magic with proxying, this PR enables new methods and properties to be added. If needed, the original functionality can always be accessed.

In short:
- Existing methods can be overridden
- New methods can be added
- New properties can be added
- Existing properties follow original behaviour (eg whether read-only or writable)
- Original method is always accessible by appending a trailing underscore to the method name eg SetText_(foo)

### Fluid Interface
This PR also adds a fluid interface in an effort to make writing plugins a little more DRY-friendly. The fluid interface works by returning the 'this' value from any method that would otherwise return nothing. Methods that return nil or false are not affected by this. Only methods that return zero values will have the fluid interface applied.

### Small Example
Here's an example of creating a window with some static text the standard way:
```
local window = finale.FCCustomLuaWindow()

local str = finale.FCString()
str.LuaString = 'this is some text'
window:CreateStatic(10, 10):SetText(str)

window:ExecuteModal(nil)
```

Now, if we include the Fluid Mixin library, immediately we have some improvements:
```
local window = finale.FCCustomLuaWindow()

window:CreateStatic(10, 10):SetText(finale.FCString():SetLuaString('this is some text'))

window:ExecuteModal(nil)
```

If we were to then override the `SetText` method so that it accepts plain strings like so:
```
mixin.register_default('FCCtrlStatic', 'SetText', function(this, str)
	if type(str) == 'string' then
		local temp = str
		str = finale.FCString():SetLuaString(temp) -- This works because of the fluid interface
	end

	-- Trailing underscore is used to refer to the original method
	this:SetText_(str)

	-- By not returning a value (copying the behaviour of the original method), the fluid interface is maintained
end)
```

The code could be reduced even further to this:
```
local window = finale.FCCustomLuaWindow()

window:CreateStatic(10, 10):SetText('this is some text')

window:ExecuteModal(nil)
```
In the example above, the same mixin could also be applied on mass by passing an array of classes and or method names.

There are other ways in which this could be used including:
- Fix methods that have bugs
- Where possible, allow methods to accept Lua types (eg string where `FCString`, array of strings where `FCStrings`, etc)
- Add additional high level functionality (I have written some of these for dialogs, which was especially useful in reducing boilerplating)
- In the absence of the ability to inherit from FC* objects, named mixins can be used to create subclasses

I believe all of these could offer significant DX improvements and make it easier to write JW Lua plugins.

There are other things I could add, depending on the direction that you want to go in. For example, I could automate the acceptance of Lua data types by introducing a secondary mixin layer, separate to user defined mixins. You could even use it as the basis for creating an whole higher-level framework for creating plugins, with greater abstraction and big reductions in boilerplate (which is something I'd personally like to see). There are plenty of options.

Let me know if I've missed something from the style guide. I've tried to follow it but I don't use Lua often so I may have overlooked something.

### DIFF
--- a/src/library/mixin.lua
+++ b/src/library/mixin.lua
@@ -126,7 +126,7 @@ Register a mixin for a Finale class that will be applied globally. Note that met
 
 @ class (string|array) The class (or an array of classes) to apply the mixin to.
 @ prop (string|table) Either the property name, or a table with pairs of (string) = (mixed)
-@ value (mixed) OPTIONAL: Method or property value. Will be ignored if prop is a table.
+@ value [mixed] OPTIONAL: Method or property value. Will be ignored if prop is a table.
 ]]
 function mixin.register_default(class, prop, value)
     class = type(class) ~= 'table' and {class} or class
@@ -134,7 +134,10 @@ function mixin.register_default(class, prop, value)
 
     for _, c in ipairs(class) do
         for p, v in pairs(prop) do
-            if type(p) == 'string' and p:sub(-1) ~= '_' then default_mixins[c][p] = copy_table(v) end
+            if type(p) == 'string' and p:sub(-1) ~= '_' then
+                default_mixins[c] = default_mixins[c] or {}
+                default_mixins[c][p] = copy_table(v)
+            end
         end
     end
 end
@@ -147,7 +150,7 @@ Register a named mixin which can then be applied by calling apply_named. If a na
 @ class (string|array) The class (or an array of classes) to apply the mixin to.
 @ mixin_name (string|array) Mixin name, or an array of names.
 @ prop (string|table) Either the property name, or a table with pairs of (string) = (mixed)
-@ value (mixed) OPTIONAL: Method or property value. Will be ignored if prop is a table.
+@ value [mixed] OPTIONAL: Method or property value. Will be ignored if prop is a table.
 ]]
 function mixin.register_named(class, mixin_name, method, func)
     mixin_name = type(mixin_name) ~= 'table' and {mixin_name} or mixin_name
--- a/src/library/mixin.lua
+++ b/src/library/mixin.lua
@@ -1,0 +1,266 @@
+--[[
+$module Fluid Mixins
+]]
+local mixin, default_mixins, named_mixins = {}, {}, {}
+
+-- Recursively copies a table, or just returns the value if not a table
+local function copy_table(t)
+    if type(t) == 'table' then
+        local new = {}
+        for k, v in pairs(t) do
+            new[copy_table(k)] = copy_table(v)
+        end
+        setmetatable(new, copy_table(getmetatable(t)))
+        return new
+    else
+        return t
+    end
+end
+
+-- Gets the real class name of a Finale object
+-- Some classes have incorrect class names, so this function attempts to resolve them with ducktyping
+local function get_class_name(object)
+    if not object or not object.ClassName then return end
+    if object:ClassName() == '__FCCollection' and object.ExecuteModal then
+        return object.RegisterHandleCommand and 'FCCustomLuaWindow' or 'FCCustomWindow'
+    end
+
+    return object:ClassName()
+end
+
+-- Attempts to determine if an object is a Finale object through ducktyping
+local function is_finale_object(object)
+    -- All finale objects implement __FCBase, so just check for the existence of __FCBase methods
+    return object and type(object) == 'userdata' and object.ClassName and object.GetClassID and true or false
+end
+
+-- Returns a function that handles the fluid interface
+function mixin.create_fluid_proxy(t, func)
+    local function proxy(...)
+        local n = select('#', ...)
+        -- If no return values, then apply the fluid interface
+        if n == 0 then
+            return t
+        end
+
+        -- Apply mixin foundation to all returned finale objects
+        for i = 1, n do
+            mixin.apply_mixin_foundation(select(i, ...))
+        end
+        return ...
+    end
+
+    return function(...)
+        return proxy(func(...))
+    end
+end
+
+-- Modifies an existing instance of an FC* object to allow adding mixins and adds primary mixins.
+function mixin.apply_mixin_foundation(object)
+    if not object or not is_finale_object(object) or object.is_mixin then return end
+    local class_name = get_class_name(object)
+    local mixin_store = {}
+    local meta = getmetatable(object)
+
+    -- We need to retain a reference to the originals for later
+    local original_index = meta.__index 
+    local original_newindex = meta.__newindex
+
+    meta.__index = function(t, k)
+        local prop
+
+        if k == 'is_mixin' then
+            return true
+        elseif mixin_store[k] then
+            prop = mixin_store[k]
+        elseif default_mixins[class_name] and default_mixins[class_name][k] then
+        	prop = default_mixins[class_name][k]
+        else
+            -- Strip trailing underscore if there is one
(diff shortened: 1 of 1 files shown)


rpatters1:
I quite like this, especially as I would like to see a move away from `finenv.UserValueInput` in favor of `FCCustomWindow`. (But the latter is currently much more verbose.) Would it also be possible to dispense with the chain call entirely and reimplement it?

Something like:

```lua
local window = finale.FCCustomLuaWindow()
window:CreateStatic(10, 10, 'this is some text')
```
I'm not a big fan of chain calls, but maybe that's because I personally read vertical code more easily than horizontal code. I would probably end up writing them vertically anyway, like:

```lua
window:CreateStatic(10, 10)
   :SetText('this is some text')
```

Nick and I have been discussing the future of Lua-in-Finale and changes we would like to see in the PDK Framework. If, just as a random hypothetical example, someday the `CreateStatic` function added an optional text parameter, how would that affect the mixin? (We aren't actually proposing this change, though it might not be a terrible idea.)

A lot of whether lua even has a future in Finale depends on whether Jari chooses to release his source code for JW Lua. I would like to talk to you privately about that, if you are interested. Some of this may directly impact the mixin library. You can contact me through my website:

[robertgpatterson.com](https://robertgpatterson.com)

ThistleSifter:
Methods can be implemented and over-ridden with whatever signature you want. The only technical limit is that in its current form only one method/property with the same name can be stored at a time. In other words, no more than 1 level of 'inheritance' (mixin + reference to original method via underscore). 

There's no obligation to take utilise chaining if you prefer not to, this just provides the option. I personally only started to use chaining relatively recently and although I like it now, it did take some time to acclimatize so I understand where you're coming from.

To address your last question: if the third parameter was optional and the existing `SetText` method was still retained with the same signature, then it wouldn't be an issue. If the third parameter was required then that would be a breaking change and any mixins would have to be updated, although the same would be true for any code that uses `CreateStatic`, so there's nothing unique in that sense. The area that is most vulnerable to breaking changes would be method chaining. If a method that previously returned nothing was updated to return a value, then any chains where that method was used could break (depending on where in the chain the call is and if the final result is utilised). But as I wrote earlier, there's no requirement to use them and I don't know what the likelihood is that a return type would be changed in such a fashion.

Thankyou for your feedback. I sent you message on your site.

rpatters1:
Well, rats my web server is having issues because of course it is.

You can just send me email directly: `robert` "at" (my domain name above).

Nick-Mazuk (approved):

Nick-Mazuk:
Approved.

My main two reservations relate to the maintainability of the codebase as a whole.

1. With the mixins and fluid interface, there are now two ways of doing things: chaining and not chaining. This could potentially lead to different parts of the codebase using each of the different code styles. As the codebase grows, this could become more of an issue as new contributors now need to learn two different ways of developing JW Lua to contribute. Since these scripts often serve as a model for those learning JW Lua, I could see potential confusion with beginners trying to copy an example using chaining and it not working.
2. I can see things becoming complicated if mixins become used inconsistently. Because then an FCNote in one script may not have the same methods and properties as an FCNote in another script. This could make scripts become less predictable for developers/code reviewers—especially if they are used in the library.

In general, though, I'm game to merge these changes. I definitely think they could improve DX a lot, and the chaining is something common in many other languages that I've missed in JW Lua.

We'll just need to be careful with future PRs in case things start to get out of hand and become unmaintainable.

@rpatters1, feel free to merge once all your concerns are met.

_P.S. @ThistleSifter, guessing you've coded professionally before? The code style, thoughtfulness, and high-level nature of this PR is really impressive for someone's first commit on GitHub._

rpatters1:
I am attempting to execute the sample code and got the following error on `mixin.register_default`:

```lua
mixin.lua:137: attempt to index field '?' (a nil value)"
```
My code to this point:

```lua

mixin.register_default('FCCtrlStatic', 'SetText', function(this, str)
	if type(str) == 'string' then
		local temp = str
		str = finale.FCString():SetLuaString(temp) -- This works because of the fluid interface
	end

	-- Trailing underscore is used to refer to the original method
	this:CreateStatic_(str)

	-- By not returning a value (copying the behaviour of the original method), the fluid interface is maintained
end)
```

and line 137 is the middle of the following 5 lines

```lua
    for _, c in ipairs(class) do
        for p, v in pairs(prop) do
            if type(p) == 'string' and p:sub(-1) ~= '_' then default_mixins[c][p] = copy_table(v) end
        end
    end
```

ThistleSifter:
@Nick-Mazuk
Funnily enough, I'm actually a classically trained trombone player, although I seem to play more jazz than anything else these days. But now I'm looking to move into the IT industry so I'm trying to getting out there more. I have done a fair amount of programming as a hobby but most of it is in a project with a private repo.

I think the solution lies in how the codebase is maintained and what policies you have in place.
- Firstly, you need control over how mixins are defined and what methods are added and overridden.
- I would add a mixins directory with two subdirectories, one for default mixins and one for named mixins.
- For default mixins, perhaps one file per class?
- If a plugin needs a mixin, it can `require` it. If a mixin needs another mixin, it can `require` it.
- Default mixins should not be allowed to require a named mixin (think of it in terms of OOP: parents can't know about children).
- All overridden methods in a default mixin should be 100% backwards compatible, since the consumer has no way of knowing whether a default mixin is applied or not (LSP).
- Named mixins should only be allowed to be defined once, and not be overridden (open-closed). [Actually I haven't enforced this, and I should. I'll make some more changes, don't merge this yet.]
- Plugins should only be allowed to use the default mixins defined in the default mixins directory. They should not be allowed to define their own. (if need, mixins can always be manually overridden on an individual object with the `apply_table` method)

This may limit some of the 'magic' that can be done (I don't like too much magic anyway) but it should keep things sane.

Since mixins are registered with function call, it should be easy to have a linter that checks for those function calls in PRs and whether they are allowed in that directory or not.

Another option is to have the mixin library applied by default (not individual mixins, just the library) which would level the playing field. If you were to do that, the library would essentially go from being a library to a framework and obviously there are a whole host of other things to think about if you do go in that direction.

I should mention that I haven't done any benchmarking with performance-intensive scripts, so I don't know what the performance implications would be for a script that has to perform tens-of-thousands of operations. It could be negligible but I don't know.

@rpatters1
That's what I get for making quick last-minute changes before uploading. Try it now, it should work.

I have also realised that I made a typo in the example I posted in the OP, which I have now fixed. So when you test it, use the updated example.

rpatters1:
This all looks good now. I'm going to merge it. I do have a question, though. Will this work in LuaJIT? (Which is more or less Lua 5.1.) I think one of the possible futures for lua-in-Finale is transitioning from Lua 5.2 to LuaJIT.

Finally, a lot of the work for creating class-specific mixins could be automated by creating a new script based on the `create_object_reflection.lua` script that I created in my fork of [Jari's test toolkit](https://github.com/rpatters1/jwluatesttoolkit). That script creates [refl-cpp](https://github.com/veselink1/refl-cpp) macros for all classes (or a selected class). It would just be a matter of changing the boilerplate in the script.

PR #117, opened 2021-11-05
Update Fluid Mixins.

ThistleSifter:
This is a quick follow up PR with some improvements to Fluid Mixins, mainly based around tighter enforcement of some OOP principles.

- Default mixins have been renamed to global mixins (that's what I had originally written in the comments and I think it is a more appropriate name).
- Applying named mixins is now done on the object itself via `object:apply_mixin('foo')`.
- Objects are now aware of which named mixins have been applied.
- It is now possible to check if an object has a named mixin applied by calling `object:has_mixin('foo')`. It is also possible to check if a global mixin is in place by calling `object:has_mixin('global')`.
- Named mixins can only be registered once and cannot be called 'global'.
- Properties cannot be overridden with methods and vice-versa.
- Errors are now caught and rethrown so that they display at the correct level with the correct line number and the correct method name.
- Aliased table.unpack for Lua 5.1 support.

Error handling proved somewhat tricky to get right but it's turned out better than I initially expected. The last remaining small issue with errors is that for methods called with a colon (eg `dialog:ExecuteModal()`), any 'bad argument' errors will have the argument number off by one (because under the hood the actual function call is not made with a colon). To fix this properly, there would need to be some way of differentiating whether a function was originally called with a dot or a colon but I don't think that's possible. The other option is that I can reduce all bad argument errors by 1, which would fix it for colon calls, but break it for dot calls. I guess it would depend on which type of call is more common. It's a very minor issue, but I thought I'd mention it.

### DIFF
--- a/src/library/mixin.lua
+++ b/src/library/mixin.lua
@@ -1,7 +1,57 @@
 --[[
 $module Fluid Mixins
 ]]
-local mixin, default_mixins, named_mixins = {}, {}, {}
+
+-- For compatibility with Lua <= 5.1
+local unpack = unpack or table.unpack
+
+local mixin, global_mixins, named_mixins = {}, {}, {}
+
+local base_mixins = {
+    is_mixin = true,
+
+--[[
+% has_mixin(name)
+
+Object Method: Checks if the object it is called on has a mixin applied.
+
+@ name (string) Mixin name.
+: (boolean)
+]]
+    has_mixin = function(t, mixin_list, _, name)
+        if name == 'global' then
+            return global_mixins[mixin.get_class_name(t)] and true or false
+        end
+
+        return mixin_list[name] and true or false
+    end,
+
+--[[
+% has_mixin(name)
+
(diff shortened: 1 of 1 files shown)


rpatters1:
I think the argument numbers being off by 1 for `:` calls is acceptable. Certainly better than making all the others be off by one.

ThistleSifter:
I've just thought of a further improvement I can make with error handling. Marking as draft for now.

ThistleSifter:
I have updated the error handling for mixins. It should be good to go now.

@rpatters1
Regarding LuaJIT compatibility: mixins should work but I haven't had the time to test it yet. I'll have the time to do that in a week or two. The only question mark in my mind is around how compilation affects error handling, in particular whether the error level information includes both a file and a line number.

rpatters1:
I'm cool with it. Just waiting for comment from @Nick-Mazuk.

Nick-Mazuk (approved):
Looks good to me

PR #118, opened 2021-12-06
Rgp changes for modeless dialogs

rpatters1:
This pr makes Transpose Chromatic and Transpose By Steps modeless dialogs when running on RGP Lua 0.56 or higher. They still work as a modal dialogs on JW Lua.


Nick-Mazuk (approved):
Thanks for updating the scripts!

PR #119, opened 2021-12-10
Class hierarchy mixins

ThistleSifter:
First off, apologies for the radio silence. Things have been quite hectic and I anticipate the next week will be even more so. Hopefully things will clear up afterwards.

I will get around to adding some actual mixins when time permits, but I had an idea that it would be good to be able to add mixins at any level in the class hierarchy to reduce unnecessary duplication. So this PR makes the following changes:
- Adds a class_hierarchy module to the library.
- Mixins can now be defined for parent classes, with the usual rules applying regarding global mixins for a child class overriding those for a parent.
- Also adds a utils module to the library for general (ie not Finale-specific) Lua utility functions (it's a bit sparse at the moment but I'd rather add to it as needed).
- Added a couple of missing checks for nil

### DIFF
--- a/src/library/mixin.lua
+++ b/src/library/mixin.lua
@@ -2,6 +2,8 @@
 $module Fluid Mixins
 ]]
 
+local utils = require("library.utils")
+
 -- For compatibility with Lua <= 5.1
 local unpack = unpack or table.unpack
 
@@ -19,10 +21,6 @@ Object Method: Checks if the object it is called on has a mixin applied.
 : (boolean)
 ]]
     has_mixin = function(t, mixin_list, _, name)
-        if name == 'global' then
-            return global_mixins[mixin.get_class_name(t)] and true or false
-        end
-
         return mixin_list[name] and true or false
     end,
 
@@ -53,24 +51,27 @@ Object Method: Applies a mixin to the object it is called on.
     end,
 }
 
--- Recursively copies a table, or just returns the value if not a table
-local function copy_table(t)
-    if type(t) == 'table' then
-        local new = {}
-        for k, v in pairs(t) do
-            new[copy_table(k)] = copy_table(v)
+-- Loads a global mixin, if it hasn't been loaded yet
+local function load_global_mixin(class_name)
+    if type(global_mixins[class_name]) ~= 'nil' then return end
+
+    success, result = pcall(function(c) return require(c) end, 'mixins.global.' .. class_name)
+
+    if not success then
+        -- If the reason it failed to load was anything other than module not found, display the error
+        if not result:match("module '[^']-' not found") then
+            error(result, 0)
         end
-        setmetatable(new, copy_table(getmetatable(t)))
-        return new
-    else
-        return t
+
+        -- Keep track of failed attempts to prevent calling require more than once per class
+        global_mixins[class_name] = false
     end
 end
 
 -- Catches an error and throws it at the specified level (relative to where this function was called)
 -- First argument is called tryfunczzz for uniqueness
 local function catch_and_rethrow(tryfunczzz, func_name, levels, ...)
-    -- If the line above moves from line #74, update this comment and the if statement below
+    -- IMPORTANT: If the line above moves from line #73, update this comment and the if statement below
     local success, result = pcall(function(...) return {tryfunczzz(...)} end, ...)
 
     if not success then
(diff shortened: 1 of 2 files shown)


rpatters1:
The class hierarchy can be determined programmatically, so I don't think it should be hard-coded here. You can also programmatically access all the method and property names, if that's needed. There is a slight difference between `JW Lua` and `RGP Lua` in how this works.

I've worked out how to do it as part of my work on `RGP Lua`. Here is sample code:

```lua
-- this function should only be called for classnames that start with "FC" or "__FC"
local get_parent_name = function(classname)
    local class = _G.finale[classname]
    if type(class) ~= "table" then return "" end
    if not finenv.IsRGPLua then -- old jw lua
        classt = class.__class
        if classt and classname ~= "__FCBase" then
            classtp = classt.__parent -- this line crashes Finale (in jw lua 0.54) if "__parent" doesn't exist, so we excluded "__FCBase" above, the only class without a parent
            if classtp and type(classtp) == "table" then            
                for k, v in pairs(_G.finale) do
                    if type(v) == "table" then
                        if v.__class and v.__class == classtp then
                            return tostring(k)
                        end
                    end                
                end
            end
        end
    else
        for k, _ in pairs(class.__parent) do
            return tostring(k)  -- in RGP Lua the v is just a dummy value, and the key is the classname of the parent
        end
    end
    return ""
end

rpatters1:
This is not a comment specific to this pull request, but I think mixins would get more traction if they were better documented. A lengthy explanation in the documentation pages (perhaps copied from the original commit of mixins) would be extremely helpful.

Also, I would like to see a fully implemented example, and I think `FCCustomWindow` (and/or `FCCustomLuaWindow`) would be a great choice to add to the library. I would convert both of mine to it for sure.

ThistleSifter:
Hmm, I agree that hardcoding is far from ideal but originally when I was testing it out, Finale would crash whenever I tried to access `__class`.  However, copy-pasting the function that you posted above seems to work fine so I'm not sure what the issue was there. In that case, is it alright if I just replace the hardcoded list with the function you posted?

And yes, I will be adding some proper documentation and examples.

Nick-Mazuk:
Nice work!

Don't have any other comments than what @rpatters1 said.

Just FYI when it comes to documentation, everything is just a markdown file in the `/docs` folder. However, everything in the `/docs/library` folder is automatically generated from the code, so updates you do there will be overwritten. You might have figured all that out already, but letting you know just in case you didn't.

rpatters1:
Sure, you are welcome to use my function.

It needs to be programmatic because there are 2 or 3 intentional discrepancies in the hierarchy between JW Lua and RGP Lua. Also, I expect that RGP Lua may be adding new classes in the future.

JW Lua crashes Finale if you try to index a finale class table with a non-existent key. The most likely culprit was `__parent` but maybe it was another. (There is a note in my routine about that.)

Both JW Lua and RGP Lua use `LuaBridge` to import classes into Lua, but `LuaBridge` underwent a lot of changes since JW Lua used it. One of them was removing the easy reflection of classes, methods, and properties. I recreated it with a home-grown version that is somewhat different but basically allows you to get the most of the same information.

rpatters1:
I was imagining the documentation would be embedded in the `mixin.lua` source code itself so that it would automatically be exported to the doc site. I don't know how much control over formatting there is, though.

ThistleSifter:
Just popping in to say I'm still alive. Life just got extraordinarily hectic. I'll be back for sure (with plenty of time to spare) in the 2nd week of February, maybe sooner if I'm lucky.

ThistleSifter:
I've included the function from rpatters1 and changed a couple of other things, including some performance improvements.

rpatters1:
This looks good to me. I hope at some point you will add a comment block to the `$module` section of `mixin.lua` that gives a detailed description of how to use the library and some use cases for why we might want to. I seem to recall that the original pull request that added it had a fairly good summary that could be used as starting point.

These comment blocks can use a subset of markdown to get formatting on the website. The `configuration.lua` library file as a fairly detailed comment block that gives an example of what I think would be helpful. The comment block is automatically added to the website when it is merged into `master`.

Finally, a really nice to have would be a mixin implementation of `FCCustomLuaWindow`. It is a mess now, and I would be very happy to use that as a starting point for working with mixins.

Nick-Mazuk:
I do think @rpatters1's thoughts are valid and it would be great to have more documentation explaining _why_ you want to use mix-ins.

However, I'm going to merge this since it looks good and we can add those comments later.

PR #120, opened 2021-12-24
Rgp changes for rgplua 0.58

rpatters1:
This pull request changes the `transposition.lua` library to get accurate custom key signature information from the document, using the new functions introduced in `RGP Lua v0.58`. It continues to work the same as before with earlier versions of `RGP Lua` and with `JW Lua`.


Nick-Mazuk (approved):
Looks good to me

PR #121, opened 2022-01-03
Add be_right_back

MichaelMcClennan:
A temporary floating bookmark.   The user 'Sets' the current position ... moves around the score(maybe to copy music) ... 'Goes Back' to the saved position.

### DIFF
--- a/src/be_right_back.lua
+++ b/src/be_right_back.lua
@@ -1,0 +1,89 @@
+function plugindef()
+    finaleplugin.Author = "Michael McClennan"
+    finaleplugin.Version = 1.0
+    finaleplugin.Copyright = "2022/01/03"
+    finaleplugin.HandlesUndo = false
+    finaleplugin.NoStore = true
+end
+
+local str = finale.FCString()
+str.LuaString = "BRB v1.0"
+BRB_dialog = finale.FCCustomLuaWindow()
+BRB_dialog:SetTitle(str)
+BRB_dialog:SetClientWidth(90)
+
+local button_go_back = BRB_dialog:CreateButton(0, 0)
+str.LuaString = "Go Back"
+button_go_back:SetWidth(60)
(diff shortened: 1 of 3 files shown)


rpatters1:
Great idea for a plugin, and I like your neat readable code. It's amazing what new possibilities a modeless window offers. I have a few suggestions.

- As a stylistic matter, I don't think boolean values should be compared to `true` or `false`. They should just be tested: `if BRB_table.page_view then`. This is especially true when testing for `false` conditions, though I did not see an example of that in this script.
- Because it has to be a modeless dialog, this plugin should have `finaleplugin.MinJWLuaVersion = 0.57` set in the `plugindef` function. That will prevent it from loading in `JW Lua` (ironically).
-  The `plugindef` function should return three strings. (See original `JW Lua` documentation for details.)
- I'm concerned that, at least on Windows, the calling of scroll view then page view menu options will cause ugly flickering. `FCMusicRegion` has a scroll-into-view method that works in either scroll or page view without changing views. (It probably works in Studio View as well, though I haven't tried it.) Look at `UI-switch-to-selected-part.lua` for details.
- But speaking of switch views, should the script remember which view you're in and restore that as well? If so, be careful not to call the menu option for, e.g., page view if you're already in page view because Finale may move the scroll bars.

Nick-Mazuk:
Hey @MichaelMcClennan, just wanted to circle back to see if you were still interested in completing this.

MichaelMcClennan:
Hi Nick

I am - 

I've been swamped with real work for a few months, 
but it mostly comes to an end at the end of August

I'll have time to devote to this and some other things then.

I think.

Nick-Mazuk:
No worries at all. Just wanted to check in.

rpatters1:
What is the status of this PR? Should we close it?

PR #126, opened 2022-01-31
add rgp lua docs to main repo

rpatters1:
Adds rgplua documentation to `docs`.


Nick-Mazuk (approved):
Looks good to me.

FYI, the images might not work right now. If they don't, I'll update the site so they work.

rpatters1:
The images aren't working, and the sidebar menu needs some work. Also, whatever translates text to the <code> tag is leaving html codes in the text. But it's a start.

Nick-Mazuk:
Docs site should be working now.

Couple things with how the docs site now parses and displays the docs:

- All images should go in a subfolder `assets`. An `assets` folder can be created anywhere inside the top-level `docs` folder in this repo. You can reference the images with relative paths as you'd expect (and did in your PR)
- `imgs` folder is renamed to `assets` as the site now supports static files of any type. For instance, you can also include files to download inside an `assets` folder in the docs.
- The name in the sidebar is based on the file name. So `docs/rgp-lua/getting-started` becomes "Getting Started". Words are split by `-` or `_`, and the beginnings of each word is capitalized (with two exceptions for "RGP Lua" and "JW Lua" so those are displayed correctly).
Screen Shot 2022-01-31 at 9 39 11 PM
- If a section does not have a "main" file (see next bullet), the top-level file for that section will just include the outline for that section. See [finalelua.com/docs/rgp-lua](https://www.finalelua.com/docs/rgp-lua) for an example.
- A section can include a "main" file. This is simply a markdown file in the `docs` folder with the same name as the section. For instance, `docs/getting-started/` has a "main" file of `docs/getting-started.md`.

Hopefully this makes sense. With #127, I've updated the docs from this PR to work with the docs site.

PR #129, opened 2022-02-01
Rgp/changes for rgplua 0.60

rpatters1:
This PR incorporates:

- all changes to scripts so that they can take advantage of new features in RGP Lua 0.60.
- updated RGP Lua docs that hopefully incorporate themselves gracfully into the finalelua.com site.
-


Nick-Mazuk (approved):
Looks good to me. I'll monitor the docs website but it looks like all the docs (including the `Notes`) should work.

PR #133, opened 2022-02-17
add comment from original mixins pull request

rpatters1:
This pull request adds a comment to `mixins.lua` that is basically a duplicate of the comment in the original pull request that added it. I have slightly edited it to get rid of headers (because I'm not sure if our auto-documentation engine supports them) and rewording here and there to make it more of a general comment. If this appears to be still accurate, we can at least use it as the starting point for documenting mixins.

@ThistleSifter

### DIFF
--- a/src/library/mixin.lua
+++ b/src/library/mixin.lua
@@ -1,5 +1,80 @@
 --[[
 $module Fluid Mixins
+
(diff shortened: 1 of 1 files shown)


Nick-Mazuk (approved):
Looks good to me

PR #138, opened 2022-02-20
Mixin bug fixes and documentation overhaul

### LINKED ISSUE #137: [BUG] Mixin library is not compatible with finenv.RetainLuaState
**Describe the bug**
The `mixin.lua` library is not compatible with the `finenv.RetainLuaState` property introduced in RGP Lua 0.60.

**To Reproduce**
1. Run the following script.
```lua
dialog = finale.FCCustomLuaWindow()
finenv.RetainLuaState = true
```
2. Run it a second time.

**Expected behavior**
The second time the script fails on
```lua
dialog = finale.FCCustomLuaWindow()
```
with
```
...mixin.lua:200: bad argument #1 to 'pairs' (table expected, got nil)"
```

**Desktop (please complete the following information):**
 - OS: macOS 10.12.1 (Monterey)
 - Finale 27.1
 - RGP Lua Version 0.61

**Additional context**
I tried changing mixin.lua:446 to
```lua
finale_ = finale_ or finale
```
but it makes no difference. I also tried this script:
```lua
if not finenv.RetainLuaState then
   mixin = require("library.mixin")
end
dialog = finale.FCCustomLuaWindow()
finenv.RetainLuaState = true
```
but it also makes no difference.

I'm still investigating it, but I though I would raise the issue so we know about it.

@ThistleSifter

ThistleSifter:
This should fix #137. I've tested it, but I would appreciate someone else double-checking.

I have rewritten the documentation so it should be more in line with the recent changes. Let me know if you'd like something changed or additional information included. I have also renamed a couple of methods for clarity.

Once this PR is merged, I'll start adding some mixins.

Also, quick reminder that what is currently line 223 and the if statement that it refers to both need to be updated if any lines are added or removed above it. Yes I know it's hacky, but it does guarantee correct propagation of errors and I haven't been able to find a better way of doing it.

### DIFF
--- a/src/library/mixin.lua
+++ b/src/library/mixin.lua
@@ -1,42 +1,137 @@
+--  Author: Edward Koltun
+--  Date: November 3, 2021
+
 --[[
 $module Fluid Mixins
 
-Fluid mixins simplify the process of writing plugins and improve code maintainability.
+The Fluid Mixins library simplifies the process of writing plugins and improves code maintainability.
 
 This library does 2 things:
 - Allows mixins to be added to any `FC*` objects
-- Adds a fluid interface for any methods that don't return anything.
+- Adds a fluid interface for any methods that return zero values.
 
-By default, it is not possible to override or extend any of the FC* objects in Finale Lua
-because objects that are tied to their underlying C++ implementation are not tables.
-So through a little bit of magic with proxying, this PR enables new methods and properties to be added.
-If needed, the original functionality can always be accessed.
 
-In short:
-- Existing methods can be overridden
-- New methods can be added
-- New properties can be added
-- Existing properties follow original behaviour (eg whether read-only or writable)
-- Original method is always accessible by appending a trailing underscore to the method name, e.g. `SetText_(foo)`
+By default, it is not possible to override or extend any of the `FC*` objects in Finale Lua because objects that are tied to their underlying C++ implementation are userdata, not tables.
+So through a little bit of magic with proxying, this library enables new methods and properties to be added, while maintaining access to the original methods and properties.
 
-The fluid interface strives to reduce the necessity of duplicate code when writing plugins.
-It works by returning the `this` value from any method that would otherwise return nothing.
-Methods that return nil or false are not affected by this. Only methods that return no values
-have the fluid interface applied.
 
-Here's an example of creating a window with some static text the standard way:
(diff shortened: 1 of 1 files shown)


rpatters1:
This mixin code is now working for my test cases (including `finenv.RetainLuaState`), with one change. That is, in the function `get_parent_class`, the references to `_G.finale` must be changed to `finale_`. (I found two of them.)

I have some concerns (possibly of no consequence) about the updated documention:
- I am not certain if the documentation generator handles all the markdown features you've added. Specifically, I seem to recall that splat-italics and double-splat headers may not have been implemented. (I know that both single- and triple-tick code blocks *are* implemented.)
- I am not certain if the document generator knows to skip single-line comments before the `$module` comment.

(These are questions for @Nick-Mazuk.)

Finally, about the hard-coded line number (that is already again not the correct line number per the comment), is it possible to avoid this hack using the following?

```lua
local line = debug.getinfo(1, "l").currentline -- 1 is top-of-stack; "l" limits info retrieval to current line (for efficiency)
```
That's a version I found on stackoverflow. If this is inefficient to call every time, it could be stored once in a global.

Nick-Mazuk:
> * I am not certain if the documentation generator handles all the markdown features you've added. Specifically, I seem to recall that splat-italics and double-splat headers may not have been implemented. (I know that both single- and triple-tick code blocks _are_ implemented.)

This docs generator supports these and all markdown features you can use in the other `/docs` files.

> * I am not certain if the documentation generator knows to skip single-line comments before the `$module` comment.

Good eye, but this will work fine. There's actually a test case explicitly for this in the parser (https://github.com/finale-lua/lua-docs-generator/blob/main/test-files/inputs/unit-test-1.lua)

Nick-Mazuk:
Since @rpatters1 is more familiar with the underlying bug, I'll let him approve and merge this PR.

rpatters1:
Outstanding issues appear to be:
- replace `_G.finale` with `finale_` in `get_parent_class` (2 instances, I believe)
- investigate if `debug.getinfo` can be used to get the line number instead of relying on a fundamentally unreliable comment.

ThistleSifter:
Thanks for that solution @rpatters1, it works perfectly on my end.
I have fixed both of the issues that you mentioned.

rpatters1 (approved):
Looks good

PR #139, opened 2022-02-22
Add control and window mixins

ThistleSifter:
This PR includes some global mixins for controls.

It's not ready to go yet, but I wanted to mention or discuss some things before going too much further.

- What's the best way of handling documentation for mixins? Should it just be the same as for everything else? I'm assuming the new directory would be picked up by the parser.
- Do you have any comments about the general direction? I have included two additional convenience methods in the global mixin for FCCtrlPopup, but they're generic and I think worth including at the global level.
- Since mixins can't have truly private methods or properties, I have adopted the JS convention of using a leading underscore.


rpatters1:
The code looks good to me. I'll defer to @Nick-Mazuk for questions about the documentation parser.

rpatters1:
Is it possible to modify the calling sequence of the constructor? I sometimes wish I could do this:

```lua
local dialog = finale.FCCustomWindow("My Title")
```

Or at least modify a parameter list like:

```lua
local myctrl = dialog:CreateButton("Button Name")
```

My primary reservation about these ideas and really all of the mixin concepts is the confusion they will cause in someone new to the codebase. I am wondering if our mixin classes should have a different prefix such as `FCM`, as in `FCMCustomLuaWindow`. That would reduce confusing in someone trying to understand the code from reading the [doc pages](https//:pdk.finalelua.com).

Another option would be, instead of overwriting the `finale` namespace, to create a new one. `mixin` would do nicely, I imagine.

ThistleSifter:
I admit that I've had similar thoughts about the need to avoid potential confusion.

I like your idea of having a different prefix. I did a little brainstorming and came up with the following possibility (although I'd still need to think about how inheritance would work with `FCM*` classes.):
- `FC*` - Unaltered classes
- `FCM*` - Modified classes that retain full backwards compatibility with the `FC*` classes (like the ones in this PR). If an `FCM*` class is not defined, will default to an `FC*` class.
- `FCX*` - Customised classes with no requirement for backwards compatibility.


Modifying a parameter list is easy to do. Modifying a constructor isn't currently supported and would require some re-working. If we go with the prefixing idea, it should be easier to implement that in a workable fashion and I'd be happy to look into it

rpatters1:
That sounds like a good plan to me. I would also encourage that original `FC` classes not receive the fluid interface, if that's feasible.

Nick-Mazuk:
+1 on disambiguating things through the naming

> * What's the best way of handling documentation for mixins? Should it just be the same as for everything else?

The way you've documented things so far looks good with me. Were you wanting to document another aspect of mixins? Currently, there isn't a way to document internal/private functions.

> * I'm assuming the new directory would be picked up by the parser.

For clarity, there are 3 separate parsers:

- Documentation comments (used for generating the documentation on the websites)
- Metadata (used to generate the script details on the website)
- Bundler (bundles each script and its imports into a single file)

Currently all of these assume library files are not nested into subfolders (i.e., they're all in the top-level `src/library` folder). I could probably rewrite them to also work with the `src/mixin/global` folder, though I'm unsure when (hopefully within a week or two?).

With that said, I like this organization. As long as these mixins are not imported by any of the scripts (directly or indirectly), merging this PR should not cause any errors for users (e.g., people should still be able to download scripts from the site). So one option is to let @ThistleSifter continue working on the mixins as is. Then when I'm able to update parsers, I'll go ahead and do that. At that point, anyone should be able to use the mixins in any script.

ThistleSifter:
Okay, I'll shelve this PR for now and work on making the changes to support the new naming convention. It will probably entail some major changes.

@Nick-Mazuk Thanks for the explanation. Maybe hold off making changes to the parser until after I've sorted out prefixes, because it will change some things and I'm still in the process of planning how it's going to look.

ThistleSifter:
A quick update. I have rewritten the mixin library and I just need to finish testing and writing updated documentation. I'm going to be away for a couple of days but I'm still hoping to have it done by the weekend.

ThistleSifter:
I have updated this PR to be compatible with the latest version of the mixin library. At the moment I have only included one mixin because there are some things I wanted to discuss before going too much further.

1) `Set*` methods that accept a single `FCString` parameter
I think it would be a good idea to set a standard for this scenario and follow it across the board since there will be many mixin methods of this type (eg `SetTitle`, `AddString`, etc). These are the options that I think are viable:
- Accept `FCString` and string
- Accept `FCString`, string and number
- Accept anything and cast it to a string

The last one is the simplest and alleviates the need to assert argument types, although perhaps that is too much leeway. Out of the other two, I prefer the second one for its added flexibility (eg populating an `FCCtrlPopup` with numbers).


2) Asserting argument types
Since mixins are new and there is no pre-existing familiarity, throwing good errors is important as it makes debugging a lot easier. So I think arguments should be asserted where possible as a bad argument error at the call site is much nicer than an obscure error from somewhere in the body of a function.

To alleviate the performance cost of doing this, I have added an extra assert function. `assert_argument` will now only assert if `finenv.DebugEnabled == true` while `force_assert_argument` will assert all of the time. Under the hood, `assert_argument` is replaced with a dummy function, so `finenv.DebugEnabled` is only checked once per script.


3) Documentation (@Nick-Mazuk)
- Is it possible to have tags on mixin methods? The ones I had in mind were:
Fluid - can be used in a method chain (can this be automatically applied to methods whose doc block has no return info?)
Override - Overrides an inehrited method
Internal - For internal use only.
Disabled - Inherited methods that have been overridden with an empty function. Needed for some FCX mixins, like the example in the mixin library documentation. Perhaps these might be better listed separately rather than tagged with the rest of the usable methods?

At the moment I haven't marked the fluid methods and the others I have just written in capitals at the start of the description.

- What would be the best way of documenting inheritance? Can inheritance be generated automatically somehow?

- Should all mixin methods have the first `self` parameter documented? 

- I almost wonder if having a separate solution for documenting mixins would be a good idea, since they're a bit different to library methods. I'm not sure how much work that would be but I could put some work into it if you want to go down that road.

rpatters1:
I like option #2. If you pass a function, boolean, user data, or table to the method it isn't likely to be anything you would want to present to a user, whereas strings, FCStrings, and numbers make sense.

I also like the idea of tying asserts to `finenv.DebugEnabled`. Just keep in mind that if the script is running in JW Lua that value will be `nil`.

ThistleSifter:
I should have clarified that it actually checks for `finenv.IsRGPLua and not finenv.DebugEnabled` when replacing `assert_argument` with a dummy function, so in JWLua asserts will run every time.


I've just added the rest of the mixins that I've written so far. I'll add documentation after we work out the best way to proceed.

ThistleSifter:
I've marked this as ready to go. I still need to go through the rest of the controls but I wanted to get some feedback on these first since the rest will be in a similar vein.

Summary of changes:
- `library/mixin.lua`:
   - Style fixes and some housekeeping
   - Fixed an issue with `nil` not being respected as a return value from mixin functions.
   - Two new assert functions: `assert` and `force_assert`. Since the native assert throws an error in place, these are needed to throw the error a level higher.
- `library/utils.lua`:
   - Added `table_remove_first` function
- Mixins:
   - Most methods that accept `FCString` now also accept `string` and `number` (there are a couple that won't accept `number`, like the folder loading methods in `FCMStrings`)
   - Methods that accept `FCStrings` now also accept a table of `string`s and `number`s
   - Added some additional convenience methods for controls
   - Parent window can now be accessed from control
   - Controls can be optionally named when they are created. Named controls can be retrieved with `GetControl`, which means local storage is no longer necessary and no more scope juggling
   - Event handlers are now passed the original object, thereby preserving mixin data for the life of the window
   - Multiple event handlers can be added to windows
   - Event handlers can be attached to controls (I'd recommend doing it this way). The above changes in combination with the ability to store absolutely anything you want on the dialog object means that you should be able to completely decouple event handlers from their surrounding scope (see example below).
   - Custom control events. Pretty much anything is possible here, but for now I've added two that I thought would be quite useful: `HandleChange` on `FCMCtrlEdit` and `HandleSelectionChange` on `FCMCtrlPopup`


With the changes in this PR, here is an example of the kind of magic you can weave:
```lua

local dialog = mixin.FCMCustomLuaWindow():SetTitle("Test Dialog")

dialog:CreateStatic(10, 10, "last_item") -- This displays the previously selected item in the popup

dialog:CreatePopup(80, 10, "my_popup"):SetStrings({"a", "b", "c"}):AddHandleSelectionChange(function(control, last_item, last_item_text, is_deleted)
    control:GetParent():GetControl("last_item"):SetText(tostring(last_item) .. ": " .. last_item_text .. (is_deleted and " (deleted)" or ""))
end)

dialog:CreateButton(10, 50):SetText("Clear Popup"):AddHandleCommand(function(control)
    control:GetParent():GetControl("my_popup"):Clear()
end)

dialog:CreateButton(10, 100):SetText("Repopulate Popup"):AddHandleCommand(function(control)
    control:GetParent():GetControl("my_popup"):SetStrings({"d", "e", "f"})
end)

dialog:ExecuteModal(nil)
```

rpatters1:
This looks good to me. I'll let Nick review and merge.

Nick-Mazuk (approved):
This looks good, though I may have to update the bundler and doc parser to handle the new `src/library/mixin` folder.

Unless there's a need to merge this ASAP, I'll wait until those two parsers are updated to merge this PR. That way we don't accidentally break the website. I'll work on updating those as soon as possible, hopefully within a week.

ThistleSifter:
No worries, there's no hurry.

I added some methods (overriding is not an option) to `FCMCtrlSwitcher` so that you can use the title of the page instead of the index but I'm not sure what to call them. For now I have just added 'String' to the end of each of them but it doesn't come across very well. Any ideas?

rpatters1:
How about `—ByTitle` rather than `—String`? So, `SetSelectedPageByTitle` and `AttachControlStringByTitle`.

But I am curious why it couldn't be an override. (I'm not saying it should be, just curious why it couldn't be.) You could detect the type of the initial parameter to take the appropriate action. I actually don't advocate this, because finding by title and by index should probably be two different functions anyway, but I am curious if there is another reason.

Should `—ByTitle` functions accept numbers for the title?

ThistleSifter:
One more little dilemma which just came to mind...

In custom control events, when the handlers are passed a `string`, it's a regular Lua `string`.  I'm wondering if it should be an `FCString` (or perhaps `FCMString`) for consistency with the PDK (I mean in general, I know that no event handlers in the PDK receive strings). Although I did notice that the `Alert*` methods in `FCUI` accept Lua strings.

There's not that much difference between writing `last_item_text` and `last_item_text.LuaString` so I'm not that concerned about convenience, I'd just like to pick something that makes sense. Once it's locked in, we can't change it in the future without breaking things.

rpatters1:
The `FCUI::Alert...` functions are the way they are because `LuaBridge` can't handle overridden `c++` functions. In the underlying framework there are versions that take `FCString`, c strings encoded ANSI (or Mac), and c strings encoded UTF8. So I don't think there is any harm in accepting `FC(M)String` as well as `string`. (Or, perhaps, any `FC` class that has a `LuaString` property.)

But if there is any cost to it, then it may not be worth it.

ThistleSifter:
Ah whoops, I don't think I explained myself very well.

Let's say I have the following code snippet:
```lua
dialog:CreatePopup(10, 10):AddHandleSelectionChange(function(control, last_item, last_item_text, is_deleted)
    -- last_item_text is the text value of the previously selected item in the popup
    print("selection change...")
end)
```

What type should `last_item_text` be when passed to the callback? `string` or `FCMString`?

rpatters1:
My preference is a lua string.

ThistleSifter:
Cool, I'll leave it as is then.

A couple of changes:
- New `is_instance_of` method in `library.mixin`
- Added iterator `Each` method to `FCMCustomWindow` to allow iterating over controls by class
- Added a few more convenience methods and `PageChange` event to `FCMCtrlSwitcher`
- Methods that accepted `FCStrings` now take multiple arguments instead of a table of strings. It makes for much better documentation and allows for proper argument validation and reporting. If someone wants to pass a table of values, they can use `table.unpack`

ThistleSifter:
More changes:

- Added `mixin_helper` library to eliminate vast amounts of copy-pasting in mixin definitions. I wasn't sure what to call the existing `FCCustomLuaWindow` events so I just called them 'native', but I'm happy to change it if a more suitable name comes up.
- Refactored control events to use `mixin_helper` library. No change in event functionality.

rpatters1:
Instead of "native" I would prefer "standard".

ThistleSifter:
More changes:
- Renamed `native` to `standard`
- To aid in demystifying mixins, I added a short summary of the modifications in the module docblock for each one.
- Added `FCMCtrlButton` (this just disables the inherited checkbox events),
- Added `FCMCtrlCheckbox`
- Added `FCMCtrlListBox`. I noticed that the PDK says that the inheritance is different on Mac and Windows. I don't have a Mac to test it on, so I hope it works okay. Should I disable the inherited standard `DataListCheck` and `DataListSelect` control events from `FCMCtrlDataList` on Mac seeing as it already has the custom `SelectionChange` event?
- Added `FCMCtrlSlider`. Getting the custom `ThumbPositionChange` event to work was difficult as `Command` event doesn't seem to fire on sliders. For earlier versions, the event dispatcher can only be triggered by a `Command` event on any other control. In newer versions, I set a timer function for 1 second, but for some reason the timer never fires for me. So I made it that the `Command` fallback remains until the timer actually fires at least once. But in theory, it should work okay. And after all that, I don't even know if anyone has used the slider in a Lua plugin!

rpatters1:
I would like to merge this PR, but it is no longer compatible with the master branch. Can @ThistleSifter take a look?

ThistleSifter:
@rpatters1 It should be okay now.

PR #144, opened 2022-03-12
Rewrite Fluid Mixins

ThistleSifter:
Here is the rewrite of the Fluid Mixin library. Most of the important information is in the documentation which I rewrote with more examples. Here's a short summary of the changes
- Mixins are now in their own namespace and the `finale` namespace has been left *almost* untouched*
- Creating an object in the original `finale` namespace does not result in mixins or the fluid interface becoming available**
- New mixin class structure with `FCM` and `FCX` mixin classes.
- Added helper functions to add more control when defining mixins. Using them can add some complexity, but I think it's worth it because it makes using the mixins more user-friendly.
- Rewrote documentation with improved examples, including how to have private properties without a memory leak.

*I added a single property to keep track of which `FC` classes had their metatables modified.
** The only scenario in which an `FC` object could be unintentionally converted to an `FCM` object is if the `FC` object is passed to an `FCM` method which then returns that same object, as mixins are activated on the return of an `FC` object from an `FCM` method. But this is quite a small catch and I doubt it will cause any problems.

Also note that when accessing a method statically (eg `finalemix.FCMControl.SetText`), the proxied method will be returned, not the original (proxying takes care of automatically promoting all returned `FC` objects to `FCM` and handles the fluid interface). This was mainly to ensure consistency of return behaviour across all mixin methods, regardless of context.

### DIFF
--- a/src/library/mixin.lua
+++ b/src/library/mixin.lua
@@ -4,217 +4,284 @@
 --[[
 $module Fluid Mixins
 
-The Fluid Mixins library simplifies the process of writing plugins and improves code maintainability.
-
-This library does 2 things:
-- Allows mixins to be added to any `FC*` objects
-- Adds a fluid interface for any methods that return zero values.
-
-
-By default, it is not possible to override or extend any of the `FC*` objects in Finale Lua because objects that are tied to their underlying C++ implementation are userdata, not tables.
-So through a little bit of magic with proxying, this library enables new methods and properties to be added, while maintaining access to the original methods and properties.
+The Fluid Mixins library does the following:
+- Modifies Finale objects to allow methods to be overridden and new methods or properties to be added. In other words, the modified Finale objects function more like regular Lua tables.
+- Mixins can be used to address bugs, to introduce time-savers, or to provide custom functionality.
+- Introduces a new namespace for accessing the mixin-enabled Finale objects.
+- Also introduces two types of formally defined mixin: `FCM` and `FCX` classes
+- As an added convenience, all methods that return zero values have a fluid interface enabled (aka method chaining)
 
 
-**Adding New Methods and Properties**
-Including the mixin library instantly enables the addition of new methods and properties. The example below demonstrates adding a new property and a new method which accesses that property.
-```
-local mixin = require('library.mixin')
-local dialog = finale.FCCustomLuaWindow()
-
-dialog.my_custom_property = 'foo'
-
-dialog.my_custom_method = function(t) print(t.my_custom_property) end
-```
-*Note: methods and properties cannot end in an underscore. For more info, see 'Accessing Original Methods'*
-
-
-**Overriding Existing Method and Properties**
-In the same way, it is also possible to override existing methods of `FC*` objects.
-```
-local mixin = require('library.mixin')
-local dialog = finale.FCCustomLuaWindow()
-
-dialog.ExecuteModal = function(t, ...)
-    print('Showing modal dialog window')
-    return t.ExecuteModal_(...)
-end
-```
-*Note: Only one copy of a mixin method is stored at a time. This means that overriding either an already overridden method or a new method will result in the first method being replaced, rendering it inaccessible.*
-
-
-To minimise conflicts, existing properties continue to follow their original behaviour and this behaviour cannot be modified. This essentially means that existing methods cannot be overridden.
-In other words:
-- Existing read-only properties will remain read-only
-- Existing writable properties remain writable
-
-```
-local mixin = require('mixin')
-
-local cell = finale.FCCell(m, s)
-cell.Measure = 2 -- Fails with an error as it is still a read-only property
+## finalemix Namespace
+To utilise the new namespace, simply include the library, which also gives access to he helper functions:
+```lua
+local finalemix = require('library.mixin')
 ```
 
+All defined mixins can be accessed through the `finalemix` namespace in the same way as the `finale` namespace. All constructors have the same signature as their `FC` originals.
 
-**Accessing Original Methods**
-For various reasons, it may be desirable to access the original method once overridden. This can be done by appending a trailing underscore to the method name.
+```lua
+local fcstr = finale.FCString()
 
+-- Base mixin-enabled FCString object
+local fcmstr = finalemix.FCMString()
+
+-- Customised mixin that extends FCMString
+local fcxstr = finalemix.FCXString()
+
+-- Customised mixin that extends FCXString. Still has the same constructor signature as FCString
+local fcxcstr = finalemix.FCXMyCustomString()
 ```
-local mixin = require('library.mixin')
-local dialog = finale.FCCustomLuaWindow()
+For more information about naming conventions and the different types of mixins, see the 'FCM Mixins' and 'FCX Mixins' sections.
 
-dialog.ExecuteModal = function(t, ...)
-    print('Showing modal dialog window')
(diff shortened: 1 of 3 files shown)


src/library/mixin.lua
```diff
@@ -4,227 +4,368 @@
...
+-- Base mixin-enabled FCString object
+local fcmstr = finalemix.FCMString()
 
+-- Customised mixin that extends FCMString
+local fcxstr = finalemix.FCXString()
 
-**Adding New Methods and Properties**
-Including the mixin library instantly enables the addition of new methods and properties. The example below demonstrates adding a new property and a new method which accesses that property.
+-- Customised mixin that extends FCXString. Still has the same constructor signature as FCString
+local fcxstr2 = finale.FCXMyCustomString()
```
rpatters1, line 33: Is `finale` correct here? Did you mean `finalemix` or `fcsstr` instead?

rpatters1 (commented):
This is looking very good to me. I will find some time to test it in the coming week.

src/library/mixin.lua
ThistleSifter: Good catch, fixed.

Nick-Mazuk:
This is looking awesome to me. I really like the clear separation between the `finale` and `finalemix` namespaces.

Nothing stands out to me that needs fixing. Though I'm not sure if I'm going to get time this week to test it. @rpatters1 , if you're able to test before me, feel free to go ahead and merge as well.

I'm not sure if the docs parser has been tested with using the ``` notation to create codeblocks, but it should work. If it doesn't, I'll update that parser to make sure it works.

ThistleSifter:
I was working on updating the control mixins PR and I realised that some functions would be better placed in the other libraries so I moved `is_finale_object` to `general_library` and I added an `unpack` function to `utils`.

ThistleSifter:
I was doing some thinking about state management, or more specifically state persistence, and I wanted to discuss whether the mixin library should do anything about it.

Here's an example:
```lua
local edit = dialog:CreateEdit(10, 10)

    print(ctrl:GetControlID() == edit:GetControlID()) --true
    print(ctrl == edit) --false
end)

dialog:ExecuteModal(nil)
```

If the first two lines were replaced with the following, it starts to get a little more troublesome as all mixin data would be lost in the handler:
```lua
local dialog = finalemix.FCXMyDialog()
local edit = dialog:CreateCustomEdit()
```

For controls and dialogs, it's quite an easy situation to fix, since I think it would be expected that the control passed to the handler would be the same as the one created (including if it's an `FCX` mixin). In this case, I'd just write some mixin method intercepts to store the original controls, compare IDs, and make sure the original is always returned when needed.

But what about something like this (which is a little contrived, but in a more complex plugin, it would be possible to run into a similar scenario while traversing function calls or iterators):
```lua
local m = finalemix.FCMMeasure()
local mm = finalemix.FCMMeasure()

-- Load the same measure
m:Load(1)
mm:Load(1)

m.Foo = 1
mm.Foo = 2 -- m.Foo is still 1
```

The PDK is quite large and I'm not familiar with a lot of it, so I don't know if anyone would expect any mixin state to be preserved between the two. My thinking is probably not, since it's not expected that state would persist between `FC` objects either, even if they do share some sort of identifier.

What do you think? I'm basically trying to work out if this needs to be handled by the library, or whether the dialog controls scenario is an outlier which can be treated separately and handled by itself.

rpatters1:
If I understand your question, you are asking if there could be a situation where all instances of a particular class need to share a property value. (Or all instances loaded from a particular Finale data structure.) I can't see that being needed in general, but perhaps specific classes might use it. Whether to implement it in the general case would depend a great deal on how complex the code is (remembering that every line of code is a liability) and if it carries any appreciable performance cost.

I also encourage you to start thinking in terms of (and testing with) the expanded capabilities of RGP Lua (if you haven't already) since that is the future. RGP Lua allows a script to maintain state between invocations from Finale. So state will become increasingly important.

One of the primary classes I want to implement (or see implemented) as an `FCM` or `FCX` class is a wrapper class for `FCCustomLuaWindow`. One of the primary uses of the wrapper class will be to maintain state of the window so that every time the user opens it, it returns to the same position it was in when the user closed it and is in the same state as when the user closed it. (I see this as primarily useful for modeless windows, but it could be useful for modal windows as well.) This means, at a minimum, storing the x,y position of the window plus allowing the user to provide additional context-specific state.

You can see *ad hoc* examples of this in [`transpose_by_step.lua`](https://github.com/finale-lua/lua-scripts/blob/master/src/transpose_by_step.lua) and [`transpose_chromatic.lua`](https://github.com/finale-lua/lua-scripts/blob/master/src/transpose_chromatic.lua). (A much more complex example is the [class browser](https://github.com/finale-lua/rgpluaclassbrowser/blob/master/rgpluaclassbrowser.lua).

I would like to formalize this, if not in the mixin code per se, in a custom mixin wrapper for `FCCustomLuaWindow`. I'm not sure how directly this comment relates to your question about state, but it seems like it does because of maintaining the controls in the state. This might allow the window to be automatically rehydrated from the state without the need of user code, a definite nice-to-have.

rpatters1:
Should this code work, or is an additional `FCXCustomLuaWindow.lua` module required?

```lua
local dialog1 = mixin.FCXCustomLuaWindow()
print("ClassName of dialog1 is "..dialog1:ClassName())
local dialog2 = mixin.FCXCustomLuaWindow()
print("ClassName of dialog2 is "..dialog2:ClassName())
```

Right now this code produces an error message:

```lua
"ERROR: ...Projects (XCode, Lazarus)/Lua Scripts/src/mixin_test.lua:27: attempt to call field 'FCXCustomLuaWindow' (a nil value)"
```
If an additional module is required, maybe that needs to be spelled out in the documentation. Also line `34` seems like it should be a triple-backtick rather than a single-backtick.

Otherwise, I'm cool with merging this, unless @ThistleSifter  wants to work more magic on it.

ThistleSifter:
Thanks for the detailed explanation. I think I'll leave the library as is for now and handle any specific cases as needed. There's more than enough magic as is!

The example you posted above is meant to fail in the way that it did. Basically, creating an `FCM` object gives you what the library previously did: an editable object with a fluid interface. But then it also loads any defined `FCM` mixins, and these are optional. When an `FCX` object is created, it first creates the underlying `FCM` and then applies modifications in layers. If the requested `FCX` mixin doesn't exist, it will return `nil`. If any of its parents don't exist, an error will be thrown during loading.

For example, take the following hierarchy:
`FCXMixC` > `FCXMixB` > `FCXMix` > `FCMString`

When calling `finalemix.FCXMixC()`, it would first create the `FCMString` object, then it would apply the `FCXMixA` mixin, then apply `FCXMixB` on top of that, and then apply `FCXMixC` on top of that.

Essentially, anything that is defined in an `FCX` mixin can just be applied manually, like so (new properties, `FCX` mixins, and manual modifications all share the same storage):
```lua
local str = finalemix.FCMString()

str.SetLuaString = function(self, str) end
str.Foo = 'foo'
str.NewMethod = function(self) end
```

Enabling them to be formally defined as `FCX` mixins (and being able to access the current 'class' in `MixinClassName`) just adds some structure to what could otherwise be the wild west.


I have updated the documentation to hopefully clear some of these things up.


Also, following on from what you were saying about state management, it's entirely conceivable that a plugin could be completely encapsulated in a customised dialog object. This could also open up interesting possibilities with interactions between plugins.

PR #148, opened 2022-03-23
Moved functions to library

jwink75:
- moved hide_stem() function to library.note_entry

- created library.layer file with layer.copy(), layer.clear(), and layer.swap() functions

### DIFF
--- a/src/library/layer.lua
+++ b/src/library/layer.lua
@@ -1,0 +1,87 @@
+--[[
+$module Layer
+]]
+
+layer = {}
+
+--[[
+% copy(source_layer, destination_layer)
+
+Duplicates the notes from the source layer to the destination. The source layer remains untouched.
+
+@ source_layer number the number (1-4) of the layer to duplicate
+@ destination_layer number the number (1-4) of the layer to be copied to
+]]
+function layer.copy(source_layer, destination_layer)
+    local region = finenv.Region()
+    local start=region.StartMeasure
+    local stop=region.EndMeasure
+    local sysstaves = finale.FCSystemStaves()
+    sysstaves:LoadAllForRegion(region)
+    source_layer = source_layer - 1
+    destination_layer = destination_layer - 1
+    for sysstaff in each(sysstaves) do
+        staffNum = sysstaff.Staff
+        local noteentry_source_layer = finale.FCNoteEntryLayer(source_layer,staffNum,start,stop)
+        noteentry_source_layer:Load()     
+        local noteentry_destination_layer = noteentry_source_layer:CreateCloneEntries(destination_layer,staffNum,start)
+        noteentry_destination_layer:Save()
+        noteentry_destination_layer:CloneTuplets(noteentry_source_layer)
+        noteentry_destination_layer:Save()
+    end
+end -- function layer_copy
+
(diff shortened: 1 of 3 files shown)


jwink75:
Oh, I also updated note_bariolage() to use those new locations...

src/library/layer.lua
```diff
@@ -0,0 +1,87 @@
...
+--[[
+% copy(source_layer, destination_layer)
+
+Duplicates the notes from the source layer to the destination. The source layer remains untouched.
+
+@ source_layer number the number (1-4) of the layer to duplicate
+@ destination_layer number the number (1-4) of the layer to be copied to
+]]
+function layer.copy(source_layer, destination_layer)
+    local region = finenv.Region()
```
Nick-Mazuk, line 16: Can we move `region` to be an argument to the function? For instance:

```lua
layer.copy(region, source_layer, destination_layer)
```

In general when writing functions—especially when it's part of a standard library—it's best to make sure the function's actions depend solely on the given arguments.

Why? There are many reasons but the most applicable one is predictability.

For instance, if you see this in the codebase `layer.copy(1, 2)`, you might be confused what _exactly_. Is this on the first staff? The first measure? The entire piece? By explicitly passing in a region `layer.copy(region, 1, 2)` you instantly know that you're copying layer 1 into layer 2 for the specified region.

If you want to read up more on this, learn about coupling and cohesion, as well as pure functions:

- [Coupling and cohesion](https://youtu.be/eiDyK_ofPPM)
- [Pure functions](https://youtu.be/4IIWib5MZKg)

_Note that pure functions aren't always strictly what we want, there are real reasons why you wouldn't want to use them (e.g., if you want to change state, i.e., actually copying music). But the idea of only relying on the inputs is generally a good idea since it lowers coupling a lot._

src/library/layer.lua
```diff
@@ -0,0 +1,87 @@
...
+--[[
+% clear(layer_to_clear)
+
+Clears all entries from a given layer.
+
+@ layer_to_clear number the number (1-4) of the layer to clear
+]]
+function layer.clear(layer_to_clear)
+    layer_to_clear = layer_to_clear - 1 -- Turn 1 based layer to 0 based layer
+    local region = finenv.Region()
```
Nick-Mazuk, line 43: Ditto, let's move this to an argument

src/library/layer.lua
```diff
@@ -0,0 +1,87 @@
...
+--[[
+% swap(swap_a, swap_b)
+
+Swaps the entries from two different layers (e.g. 1-->2 and 2-->1).
+
+@ swap_a number the number (1-4) of the first layer to be swapped
+@ swap_b number the number (1-4) of the second layer to be swapped
+]]
+function layer.swap(swap_a, swap_b)
+    local region = finenv.Region()
```
Nick-Mazuk, line 65: Same here

Nick-Mazuk (commented):
Looks good for the most part. You got the syntax of creating library functions and documenting them perfect!

Just adding a couple of nits to improve code maintainability.

Nick-Mazuk (approved):
Looks good to me. Thanks for the changes!

Issue #5, opened 2020-06-09
[BUG] Lua Linter setup


rpatters1:
**Describe the bug**
This is not exactly a bug, but I don't see a better choice for where to put this.

The instructions for installing luac on Mac should be: `brew install lua`

Also, I have installed Visual Studio Code, lua (including luac), and lualinter. It says it is running in Visual Studio Code, with the default options. And my Code explorer shows the Lua Scripts folder that contains luaconfig.config. But I'm getting no de-linting joy. Any ideas?

Nick-Mazuk:
Ok, I'll update the markdown file.

If you go to preferences (⌘ comma), is "Format on Save" turned on? If not, that's likely the issue. I'm also adding that to the markdown file.

rpatters1:
All three of Format on Save, Paste, and Type were unchecked. But I'm still
not getting the love.

rpatters1:
I mean, I'm still not getting the love after checking them.

rpatters1:
I'm getting some formatting, but it seems to be generic. Like it
automatically maintains the indent levels for if end, for end, etc. But it
does not change the formatting if you put an if statement on a single line.

Nick-Mazuk:
Interesting. Honestly, I don't know what's causing that. I had no trouble with the setup myself. I've tried to replicate the issue on my end, but I haven't been able to.

Nick-Mazuk:
I may have written about the wrong plugin…

In short, I was wondering why my code formatting wasn't working anymore for Lua files and had to debug it. The `lua-linter` extension did not seem to be formatting the code. Instead, I had another extension called `vscode-lua-format` which formats the code.

This extension was already installed for me, but I had to add this to VS Code's `settings.json`

```json
"[lua]": {
    "editor.defaultFormatter": "Koihik.vscode-lua-format"
},
```

So my guess is that to get the code formatting, you need to:

1. Install [`vscode-lua-format`](https://marketplace.visualstudio.com/items?itemName=Koihik.vscode-lua-format)
2. Add the above snippet to your `settings.json`

If this works for you, I'll update the docs.

rpatters1:
Remind me how to tell if it's working. Also, the settings.json I found was
in

~/Library/Application Support/Code

Adding it to that didn't seem to make any difference. Is there another
settings.json?

Nick-Mazuk:
If everything is working, this…

```lua
if variable<2 then print('hello') end
```

…should be reformatted to this when you save:

```lua
[...]

PR #156, opened 2022-03-27
Add recommended VS Code settings and extensions

Nick-Mazuk:
Because of the ongoing efforts to resolve https://github.com/finale-lua/lua-scripts/issues/5, I decided to add my current VS Code settings and extensions to the `.vscode` folder.

In theory, this does two things:

1. New users will get notified to install the recommended extensions
2. All users will be working with the same VS Code settings (e.g., code formatting, linting, etc.)

That means we can help make everyone's development environment consistent. And aside from pressing the button to "download recommended extensions", developers won't need to do anything to setup VS Code.

To be honest, I've never done this before so it may not work as expected. But it's potentially worth a shot. At the very least, it might resolve the existing [VS Code code formatting issue](https://github.com/finale-lua/lua-scripts/issues/5) since we'll share the same settings and extensions.

The may drawback is that if anyone has a `.vscode` folder in their local version of the repo, their settings may also get unintentionally added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@ jwlua.fxt
 jwluatagfile.xml
 SciLexer.dll
 jwlua64.fxtco
-.vscode
 
 # ignore Finale files (used for testing)
 *.mus
@@ -22,3 +21,6 @@ coverage
 mobdebug.lua
 
 src/personal*
+
+.vscode/settings.json
+*/**/.vscode
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,3 @@
+{
+    "recommendations": ["Koihik.vscode-lua-format", "sumneko.lua"]
+}
--- a/.vscode/recommended_settings.json
+++ b/.vscode/recommended_settings.json
@@ -1,0 +1,21 @@
+{
+    "Lua.diagnostics.globals": [
+        "finale",
+        "finenv",
+        "finaleplugin",
+        "each",
+        "eachentrysaved",
+        "eachbackwards",
+        "eachentry",
+        "pairsbykeys",
+        "loadall",
+        "bit32"
+    ],
+    "editor.formatOnSave": true,
+    "Lua.diagnostics.disable": ["lowercase-global", "luadoc-miss-cate-name"],
+    "vscode-lua-format.configPath": "./luaconfig.config",
+    "[lua]": {
+        "editor.defaultFormatter": "Koihik.vscode-lua-format"
+    },
+    "Lua.runtime.version": "Lua 5.2"
+}


Nick-Mazuk:
> This brings up another issue, though. Is it wise to force everyone to have the same `.vscode` settings?

These settings will only affect this current repository. You can still have your own settings in the "User" scope. However, any settings included in this `.vscode/settings.json` will override user settings. So if this PR is merged, I'd recommend keeping the settings as minimal as possible.

https://code.visualstudio.com/docs/getstarted/settings

Nick-Mazuk:
> Your `.vscode` settings are considerably different than mine, so yes I am hoping it can (finally) resolve #5. My settings currently include these that aren't in yours:
> 
> ```
>     "editor.formatOnType": true,
>     "editor.formatOnSave": true,
>     "lualinter.warnOnSave": true
> ```

I do have `"editor.formatOnSave": true,`, but the other two I don't. I don't think they affect whether or not code formatting works?

Unless they do, I'd recommend not adding them to this `settings.json` for the same reasons in my previous comment.

rpatters1:
I guess my discomfort with this PR stems from the fact that it will be very easy to modify the `.vscode` directory without intending to. Also, just as an example, I currently actually point VS Code directly at the `src` subdirectory, so that's where my `.vscode` directory is. With this PR, that directory would be added into the repo as well.

Of course, I can easily change to not do that, but every user of this repo is likely to have slightly different work habits and preferences. It could become difficult to manage them all. I am wondering if instead of including `.vscode` in the repo we should instead put the recommended json files in a `recommended_vscode` directory. Then users would be free to user them or not. I realize this adds room for error in updating them. Maybe there is some other way I haven't thought of.

Nick-Mazuk:
Hmm, that does make sense. The `recommended_vscode` might be the best option, then.



> Also, just as an example, I currently actually point VS Code directly at the `src` subdirectory, so that's where my `.vscode` directory is. With this PR, that directory would be added into the repo as well.

I can update the `.gitignore` to only track the `.vscode` folder in the root directory, but then you wouldn't get the automatic updates.

> I guess my discomfort with this PR stems from the fact that it will be very easy to modify the `.vscode` directory without intending to.

This is true, but I feel like we can easily catch this in code review.

rpatters1:
Should we add this line to `settings.json`?

```json
    "lua.targetVersion": "5.2"
```

Nick-Mazuk:
> Should we add this line to `settings.json`?
> 
> ```json
>     "lua.targetVersion": "5.2",
> ```

That would be a good idea.

Nick-Mazuk:
@rpatters1 

I'm not seeing a setting named `lua.targetVersion`. I do see:

```json
"Lua.runtime.version": "Lua 5.2"
```

Nick-Mazuk:
I've updated the settings to hopefully resolve your hesitations.

- All `.vscode` folders not in the root will be git ignored.
- The settings file is not called `recommended_settings.json`
- I've kept the `extensions.json` as is since it only recommends extensions, and a user can easily ignore this message (and tell VS Code to not bother them again about it).

rpatters1:
`lua.targetVersion`

Some extension somewhere added it to my settings.json. Maybe it was that debugger. BTW I have a whole `launch.json` file that may need to be added at some point, if I can ever get debugging to work.

At any rate, I suppose `"Lua.runtime.version": "Lua 5.2"` is fine for now.

Also, feel free to go ahead and merge this one.

PR #165, opened 2022-04-29
Scope staff_number to local context

Nick-Mazuk:
One variable was just missing the `local` keyword. I added it.


rpatters1 (approved):
Sorry I missed this somehow and just noticed it tonight.

PR #167, opened 2022-05-01
Measurement Unit Mixins

ThistleSifter:
This PR mainly contains mixins that deal with measurements and being able to easily control the measurement unit that they are displayed in. It also contains some bug fixes and some other small additions. There are a lot of changes/additions so I'll just summarise the ones that are important or that might need discussing. I wrote all of these a few weeks ago so I hope I've remembered everything.

Library Changes:
- `library/mixin.lua`
  - In keeping with the changes in #150, personal mixins can now be added to a `personal_mixin` folder. Personal mixins are checked first, so they can override a public mixin.
  - Added `UI` function which is the mixin equivalent of `finenv.UI()`
- `library/measurement.lua`
  - Added methods that return an abbreviation for the measurement unit. I haven't come across an official abbreviation for EVPU or Pica, so let me know if you would like something different.
- `mixin_helper`
  - Mainly just a large refactor to enable code reuse between custom control and window change events.

FCM Mixin Changes:
- `FCMCtrlEdit`
  - Added some missing change event triggers
- `FCMCtrlPopup`, `FCMUI`
  - Additional convenience methods.
- `FCMCustomWindow`
  - Parent window is now stored and is accessible via `GetParent` while window is showing as modal.
- `FCMCustomLuaWindow`
  - `Command` events occasionally fire twice (eg at Init, or when a child modal is opened) for the same control, the second while the first is still running. So I added a check to avoid concurrency issues and race conditions.

FCX Mixins:
- `FCXCustomLuaWindow`
  - Includes the ability to set the measurement unit at the window level. This is then adhered to by the control mixins below.
  - Any control that has a method called `UpdateMeasurementUnit` will be instantly updated when the measurement unit is changed.
- `FCXCtrlMeasurementUnitPopup`
  - This is a popup that allows the user to set the measurement unit for the window in the same manner as Finale's in-built dialogs (eg Document Options, Page Format, etc) 
- `FCXCtrlMeasurementEdit`
  - Displays measurements in the parent window's measurement unit and updates automatically when it changes.
- `FCXCtrlStatic`
  - Added methods for setting a measurement value
  - Measurement values ares displayed in the parent window's measurement unit with the option to set a display suffix.
- `FCXCtrlUpDown`
  - `ConnectMeasurementEdit` will only accept a `FCXCtrlMeasurementEdit`
  - Added methods to set a custom step size for different measurement units. I included a set of default step sizes which are more closely in harmony with the measurement unit they relate to, rather than a blanket step size of 1 EVPU. Let me know what you think or if any of the defaults need to be changed. Play around with the test script below to see it in action. There are different defaults depending on whether the MeasurementEdit's type is EVPU or EFIX.

Here is a script for testing out measurement edits, the new up/down mixin and the measurement unit popup:
```lua

local edit = dialog:CreateMeasurementEdit(10, 10)

-- To test default EFIX step sizes, uncomment the line below
--edit:SetTypeMeasurementEfix()

dialog:CreateUpDown(80, 10):ConnectMeasurementEdit(edit, 0, 100)

dialog:CreateMeasurementUnitPopup(10, 40)

dialog:ExecuteModal(nil)
```

### DIFF
--- a/src/library/measurement.lua
+++ b/src/library/measurement.lua
@@ -3,6 +3,33 @@ $module measurement
 ]] --
 local measurement = {}
 
+local unit_names = {
+    [finale.MEASUREMENTUNIT_EVPUS] = "EVPUs",
+    [finale.MEASUREMENTUNIT_INCHES] = "Inches",
+    [finale.MEASUREMENTUNIT_CENTIMETERS] = "Centimeters",
+    [finale.MEASUREMENTUNIT_POINTS] = "Points",
+    [finale.MEASUREMENTUNIT_PICAS] = "Picas",
+    [finale.MEASUREMENTUNIT_SPACES] = "Spaces",
+}
+
+local unit_suffixes = {
+    [finale.MEASUREMENTUNIT_EVPUS] = "e",
+    [finale.MEASUREMENTUNIT_INCHES] = "i",
+    [finale.MEASUREMENTUNIT_CENTIMETERS] = "c",
+    [finale.MEASUREMENTUNIT_POINTS] = "pt",
+    [finale.MEASUREMENTUNIT_PICAS] = "p",
+    [finale.MEASUREMENTUNIT_SPACES] = "s",
+}
+
+local unit_abbreviations = {
+    [finale.MEASUREMENTUNIT_EVPUS] = "ev",
+    [finale.MEASUREMENTUNIT_INCHES] = "in",
+    [finale.MEASUREMENTUNIT_CENTIMETERS] = "cm",
+    [finale.MEASUREMENTUNIT_POINTS] = "pt",
+    [finale.MEASUREMENTUNIT_PICAS] = "",
+    [finale.MEASUREMENTUNIT_SPACES] = "sp",
+}
+
 --[[
 % convert_to_EVPUs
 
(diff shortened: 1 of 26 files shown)


rpatters1:
I would encourage you to get setup with Visual Studio Code using the recommended settings recently committed to the master branch. this will allow you to delint your code. The lastest mixin merge (#139) introduced a number of lint issues that look like they are probably bugs.

Nick-Mazuk:
In keeping inline with Robert's suggestion, all you need to do is copy the contents of `.vscode/recommended_settings.json` into a new `.vscode/settings.json`, then install the recommended extensions (either VS Code will prompt you, or you can manually install the ones listed in `.vscode/extensions.json`.

This will give you several benefits:

- Automatic code formatting
- Linting to remove bugs and code likely to contain bugs
- Better syntax highlighting and code completion

ThistleSifter:
Okay, I have VS Code up and running and I have fixed all the linting issues that showed up.

Nick-Mazuk:
@ThistleSifter unrelated to the PR, did you find getting VS Code setup easy? Any issues you may have run into?

I ask because Robert and I spent a long time trying to get things setup for him (finally just figured it out a week or two ago), and it would be great to get some feedback to see if we're on the right direction of helping others get setup as well.

The PR looks good to me.

Nick-Mazuk (approved):

ThistleSifter:
I didn't run into any issues setting it up, but I have noticed that occasionally the style fixer does odd things.

Eg The comment in line 242 in `mixin/FCMCtrlListBox.lua` applies to the `else` below, so it should only be indented once, not twice.

I'm not a fan of some of the line splitting, although I understand that these boil down to personal preference so I'm not going to quibble about it:
- Line 447 in `mixin/FCMCtrlListBox.lua`, which it won't let me split into multiple lines so it visually matches the following arguments.
- I'd prefer to have the `for...do` in line 216 in `mixin/FCMCusomWindow.lua` be all on the same line
- The splitting on line 32 and below in `mixin/FCXCustomLuaWindow.lua` isn't great.

rpatters1:
I tend to agree with @ThistleSifter that I don't like the forced line splitting. In fact, I continue to do most of my editing in ZeroBrane because of the debugging. But it also allows me to fix lint issues found by VSCode without having to take the line splitting on board.

Nick-Mazuk:
Since the comments about the formatter itself are valid but off-topic to this PR, I moved that conversation to a separate issue (#168). That way, we can also easily keep track of it in the future.

TLDR, I'm open to improving the automated formatting.

@rpatters1 , feel free to merge this PR if it looks good to you.

PR #175, opened 2022-05-17
Store window state

ThistleSifter:
This PR adds some mixin methods that allow window state (position & size) to persist across script executions and window showings.

This is primarily controlled with two methods: `SetAutoRestorePosition` and `SetAutoRestoreSize`. The default behaviour in `FCMCustomLuaWindow` is disabled (to keep it backwards compatible) and the default behaviour in `FCXCustomLuaWindow` is enabled for position only.

While testing I noticed that restoring position for a window that can't be user-resized can have strange results:
```lua

local function create_window()
    local window = mixin.FCXCustomLuaWindow()
    window:SetAutoRestoreSize(true)
    window:CreateEdit(10, 10):SetText("")

    finenv.RegisterModelessDialog(window)
    return window
end

if not dialog then
    dialog = create_window()
    finenv.RetainLuaState = true
end

dialog:ShowModeless()
```

On second running, the window balloons out to a much larger size.

### DIFF
--- a/src/library/mixin.lua
+++ b/src/library/mixin.lua
@@ -613,13 +613,7 @@ function mixin.subclass(object, class_name)
         error("Object is not a finale object.", 2)
     end
 
-    local success, result = pcall(mixin.subclass_helper, object, class_name)
-
-    if not success then
-        error(result, 2)
-    end
-
-    if not result then
+    if not catch_and_rethrow(mixin.subclass_helper, 2, object, class_name) then
         error(class_name .. " is not a subclass of " .. object.MixinClass, 2)
     end
 
@@ -628,13 +622,21 @@ end
 
 -- Returns true on success, false if class_name is not a subclass of the object, and throws errors for everything else
 -- Returns false because we only want the originally requested class name for the error message, which is then handled by mixin.subclass
-function mixin.subclass_helper(object, class_name)
+function mixin.subclass_helper(object, class_name, suppress_errors)
     if not object.MixinClass then
-        error("Object is not mixin-enabled.", 0)
+        if suppress_errors then
+            return false
+        end
+
+        error("Object is not mixin-enabled.", 2)
     end
 
     if not is_fcx_class_name(class_name) then
-        error("Mixins can only be subclassed with an FCX class.", 0)
+        if suppress_errors then
+            return false
+        end
+
+        error("Mixins can only be subclassed with an FCX class.", 2)
     end
 
     if object.MixinClass == class_name then return true end
@@ -642,7 +644,11 @@ function mixin.subclass_helper(object, class_name)
     mixin.load_mixin_class(class_name)
 
     if not mixin_classes[class_name] then
-        error("Mixin '" .. class_name .. "' not found.", 0)
+        if suppress_errors then
+            return false
+        end
+
+        error("Mixin '" .. class_name .. "' not found.", 2)
     end
 
     -- If we've reached the top of the FCX inheritance tree and the class names don't match, then class_name is not a subclass
@@ -651,8 +657,10 @@ function mixin.subclass_helper(object, class_name)
     end
 
     -- If loading the parent of class_name fails, then it's not a subclass of the object
-    if mixin_classes[class_name].props.MixinParent ~= object.MixinClass and not mixin.subclass_helper(object, mixin_classes[class_name].props.MixinParent) then
-        return false
+    if mixin_classes[class_name].props.MixinParent ~= object.MixinClass then
+        if not catch_and_rethrow(mixin.subclass_helper, 2, object, mixin_classes[class_name].props.MixinParent) then
+            return false
+        end
     end
 
     -- Copy the methods and properties over
@@ -663,7 +671,7 @@ function mixin.subclass_helper(object, class_name)
 
     -- Run initialiser, if there is one
     if mixin_classes[class_name].props.Init then
-        props.Init(object)
+        catch_and_rethrow(object.Init, 2, object)
     end
 
     return true
@@ -686,9 +694,9 @@ function mixin.create_fcx(class_name, ...)
 
     if not object then return nil end
 
-    local success, result = pcall(mixin.subclass_helper, object, class_name)
-
-    if not success or not result then return nil end
+    if not catch_and_rethrow(mixin.subclass_helper, 2, object, class_name, false) then
+        return nil
+    end
 
     return object
 end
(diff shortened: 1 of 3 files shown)


rpatters1:
One of the disadvantages I've discovered about calling C++ from Lua is that a C++ function can have a required `bool` parameter, but there is no way to force the Lua caller to provide it. It silently defaults to `false`.

The bug here is cause by the fact that the `StorePosition` method has a required `bool` parameter (in C++) that determines if the window size data is to be stored as well as the position. To fix this bug, the `restore_position` function should be:

```lua
local function restore_position(window)
    if private[window].HasBeenShown and (finenv.MajorVersion > 0 or finenv.MinorVersion >= 60) then
        if private[window].AutoRestoreSize and private[window].StoredWidth ~= -1 then
            window:StorePosition(true) -- <<===
            window:SetRestorePositionData_(
                private[window].StoredX, private[window].StoredY, private[window].StoredWidth,
                private[window].StoredHeight)
            window:RestorePosition()
        elseif private[window].AutoRestorePosition then
            window:StorePosition(false) -- <<===
            window:SetRestorePositionOnlyData_(private[window].StoredX, private[window].StoredY)
            window:RestorePosition()
        end
    end
end
```

Note the arrows I've added by each call to `StorePosition`.

ThistleSifter:
I've changed the lines you marked but I'm still getting the same behaviour when restoring size.

In the test script, it's setting these values:
`width: 108`
`height: 114`

But after calling `RestorePosition()` it has these values:
`width: 480`
`height: 320`

rpatters1:
I'm trying to get my `transpose_by_step.lua` script to work with mixins. It's very close but I'm having some challenges. Please consider the following sample code:

```lua

-- require("mobdebug").start()

function on_ok()
    finenv.UI():AlertInfo(editctrl_val, "On Okay")
end

function on_edit_change(editctrl, strtext)
    editctrl_val = strtext
end

local function create_window()
    local window = mixin.FCXCustomLuaWindow()
    window:CreateEdit(10, 10):AddHandleChange(on_edit_change):SetText("1234")
    window:CreateOkButton()
    window:RegisterHandleOkButtonPressed(on_ok)
    window.OkButtonCanClose = false

    finenv.RegisterModelessDialog(window)
    return window
end

if not dialog then
    dialog = create_window()
    finenv.RetainLuaState = true
end

dialog:ShowModeless()
```

I'm having to issues with this code:

1. When `on_edit_change` is called, the value in `strtext` is truncated by one character.
2. `on_edit_change` is not called when `ShowModeless` happens, so my global `editctrl_val` is stale until the user actually edits the value in the dialog box.

I would welcome suggestions for how to address item 2. My goal is not to have to keep a reference to the edit control, but I suppose that would work. (Item 1 is easy to work around by calling the edit control directly.)

rpatters1:
I'm working on this right now. Something still not right, but in the mean time, I'm wondering how you would feel about adding this to the default HandleCloseWindow function:

```lua
    if finenv.DebugEnabled then
        if global_dialog:QueryLastCommandModifierKeys(finale.CMDMODKEY_ALT) or
            global_dialog:QueryLastCommandModifierKeys(finale.CMDMODKEY_SHIFT) then
            finenv.RetainLuaState = false
        end
    end
```

The issue is that without some way to end the Lua State, it is impossible to stop debugging a script and start editing it. (At least with ZeroBraneStudio.) Without this code, one has to restart Finale in order to edit the script.

rpatters1:
In response to your comments about destroying the underlying `FCCustomLuaWindow` instance, I will just throw out that there is no provision in the PDK Framework for reusing the same instance to open a different window. The Framework maintains a lot of state about the window at a lot of different levels, and I don't think it's ever been tested to try to run a new window on state that hasn't been initialized from scratch.

I guess we're going to find out.

rpatters1:
Re `clone`, it's ugly to force all implementations to do that. What happens if they don't? My feeling is that either reuse is allowed or it isn't. We shouldn't leave it up to consumers of the library.

rpatters1:
Got it.

If an edit text is not initialized when it is created, it does not retain its value across invocations of the script. Here is a simple example:

```lua

if not finenv.RetainLuaState or not dialog then
    dialog = mixin.FCXCustomLuaWindow()
    dialog:CreateEdit(10, 10) -- no initialization
    finenv.RetainLuaState = true
end

finenv.RegisterModelessDialog(dialog)
dialog:ShowModeless()
```

ThistleSifter:
Yeah, you're right, it's not pretty. I still think reuse should be allowed. Regardless of whether I was using this library or not, I would expect that storing a dialog in a global variable would persist with `finenv.RetainLuaState`. If it doesn't (and in a way that's already the case with controls), I'd consider that a PDK bug, but I don't know if that's fixable. If we do get to a worst case scenario and we find that something doesn't work, I can always write some proxying so that the `FCCustomLuaWindow` mixins have another `FCCustomLuaWindow` mixin in the background that everything is forwarded to, which can then be silently cloned and swapped out in the background on subsequent openings.

Yes it has to be before the window is first shown. I'll work on a proper solution for persisting control values across script executions anyway, so we don't need to rely on hacks.

rpatters1:
Followup: getting an error in this line in JW Lua

```lua
local dialog = mixin.FCXCustomLuaWindow():SetTitle("Transpose By Steps")
```
"Attempt to index nil value".

The require seems to have worked but `mixin.FCXCustomLuaWindow()` is returning nil.

Not sure how much effort should go into JW Lua support, but the nice thing about supporting it is that it allows the script to work in 32-bit versions of Finale.

rpatters1:
More info. This line seems to be the (or at least a) culprit:

```lua
FCXCustomLuaWindow.lua:40: attempt to call method 'SetAutoRestorePosition' (a nil value)
```

rpatters1:
As a test I modified `FCXCustomLuaWindow.lua` line 40 to be this:

```lua

    if finenv.IsRGPLua then
        self:SetAutoRestorePosition(true)
    end
```

My script now works correctly in JW Lua. I'm not sure if this is the best fix, though.

ThistleSifter:
I've fixed the issue in `FCXCustomLuaWindow`. Part of the problem is that errors are currently suppressed from `Init` functions, so I'll look at fixing that too.

ThistleSifter:
I've sorted out the issue with errors being suppressed from `Init` methods and also improved the displaying of errors from `mixin.subclass`.

rpatters1:
Okay, I'm going ahead with merging this PR. We can deal with how to deploy in a future one.

PR #176, opened 2022-05-17
Rgp changes for rgplua 0.62

### LINKED ISSUE #169: Allow for users to download "suites" of scripts
Currently on the website, downloading scripts is simple—one download per script, and one script per download. 

However, many of the scripts could be bundled together to form a "suite" of scripts. For instance, there are many scripts for moving dynamics/chords/staves/expressions/measure-numbers up/down. These all fall into the "move things vertically" category. Wouldn't it be nice for users to be able to download all of these with one click?

Or perhaps another example with a stronger case. Most of my unpublished scripts now are just slight variations on one idea. I have 14 scripts to change key signatures. I have another 12 or so for time signatures. I have 8 to change barlines. Etc.

I've resisted publishing these since that could flood our download page with a lot of noise. Want to search for a script related to key signatures? Too bad, you have 14 more scripts to wade through that are almost identical.

Now I have these scripts broken out since Keyboard Maestro makes it very quick and easy to select each of them. And if I were to combine these into a single script with a menu to select the specific key (for instance), now it's no better than Finale's key signature menu—and you lose out on the massive speed gains from the KM/Lua combo.

So if there was some way I could also publish these scripts but be able to group them into a suite of scripts with a single download button, that would allow others to get use out of my scripts while still making it easy to find scripts on the website.

No idea how to do this yet, but creating an issue in case anyone else has ideas or wants to help.

### LINKED ISSUE #171: Tie Library
This "issue" is to discuss the support for `FCTieMod` that is coming in RGP Lua 0.62. I have full support now working for `FCTieMod` on my development machine. However, I want to alert this community to a big issue around editing ties in Finale. To initialize an `FCTieMod` so that you can modify a tie's placement, you have to know:

- the direction (up or down) of the tie
- both the tied-from and tied-to notes
- whether it has Outer or Inner placement
- whether it starts/ends on note side or stem side
- whether it crosses a system break
- the specific combination of preferences selected in `FCTiePrefs`
- the specific combination of preferences for ties selected in `FCLayerPrefs`

Just looking down the list, you begin to realize the number of dependencies to know these (especially the tie direction) is quite high. Furthermore, it depends entirely on reverse-engineering Finale's current behavior, which is sometimes quite idiosyncratic in corner cases such as ties to nothing. Finally, even if you figure all this out, the code is extremely brittle if Finale's behavior changes.

I know of no way to get Finale to tell us this information. There are no "tie metrics" in the PDK, more's the pity.

The good news is that I figured out all this code over twenty years ago for my Tie Mover plugin. And Finale's behavior for ties does not appear meaningfully to have changed since then. The question then comes down to where to put this code. Originally my idea was to lard it into the PDK Framework. This would have the advantage of making it available to any Lua programmer. But it turns out that whether you can find both notes in the tie depends a great deal on how the `FCNote` instance was created. And if I'm not going to handle that for Lua scripts, the whole plan starts to unravel. You get into issues around how to keep references to variables alive, and it turns into a rabbit hole.

So the solution I elected instead was to force the Lua programmer to do all the work. I provide all the methods on `FCTieMod` to initialize the Tie Placement endpoints, but the Lua script has to figure out what to pass in. Obviously, this has the drawback that basically no one will likely ever take the trouble.

Enter `tie.lua`, a tie library that I would like to add to this repo. It basically is a dump of all the work I did in ~2000 to figure out Finale's behavior. Now initializing `FCTieMod` for editing tie placement is as simple as:

```lua
-- This code nudges selected ties up by 12 evpu.
-- It doesn't handle the case where a note has both a tie end and a tie start.
-- That's a refinement for a real implementation.

local tie = require("library.tie")

tie_prefs = finale.FCTiePrefs()
tie_prefs:Load(0)

for entry in eachentrysaved(finenv.Region()) do
    for note in each(entry) do
        local tie_mod
        if note.Tie then
            tie_mod = finale.FCTieMod(finale.TIEMODTYPE_TIESTART)
        elseif note.TieBackwards then
            tie_mod = finale.FCTieMod(finale.TIEMODTYPE_TIEEND)
        end
        if tie_mod then
            local loaded = tie_mod:LoadAt(note)
            tie.activate_endpoints(note, tie_mod, finenv.UI():IsPageView(), tie_prefs) -- activate any endpoint that needs it
            tie_mod.StartVerticalPos = tie_mod.StartVerticalPos + 12
            tie_mod.EndVerticalPos = tie_mod.EndVerticalPos + 12
            tie_mod:SaveAt(note)
         end
      end
   end
end
```

All the magic happens in `tie.activate_endpoints`.

The more I think about this approach, the more I think it is the correct one. Now the Lua script owns all the variables. But more importantly, if Finale ever changes its behavior around ties, we don't have to wait for a new version of RGP Lua to incorporate those changes. Same goes if I have introduced bugs in extracting my Tie Mover code or if my original code had errors. They will be easy to fix if the code is in Lua.

I know that @Nick-Mazuk doesn't like incorporating library code that isn't used anywhere. But this seems like a chicken and egg situation. No one is likely to use ties without the library.

rpatters1:
This pull request contains:

- Documentation updates for RGP Lua 0.62.
- `tie.lua` library with code for initializing fields of `FCTieMod` that require activation. (Resolves #171.)
- various updated and consolidated scripts that use the new `finaleplugin.AdditionalMenuOptions` field to consolidate multiple scripts into one. (Resolves #169.)
- Change Lua formatter line length from 120 to 180 (which suits my machine, but I'm not fixated on that number.)


Nick-Mazuk (approved):
LGTM

PR #178, opened 2022-05-20
bug fixes May 2022

rpatters1:
- Fixes issue where transposition scripts can't run in JW Lua.
- Fixes issue with staff exploders not preserving user selected region


Nick-Mazuk (approved):
LGTM

PR #187, opened 2022-05-23
Update cross_staff_offset.lua

cv-on-hub:
Tiny cosmetic tweak and added line of instructive information.

### DIFF
--- a/src/cross_staff_offset.lua
+++ b/src/cross_staff_offset.lua
@@ -38,10 +38,14 @@ function user_selects_offset()
         current_vert = current_vert + vertical_step
     end
 
-    local static = dialog:CreateStatic(0, current_vert + 5)
-    str.LuaString = "units are in EVPUs: 144 evpu = 0.5 inch / 1.27 cm"
+    local static = dialog:CreateStatic(0, current_vert + 8)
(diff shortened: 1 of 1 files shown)


Nick-Mazuk:
Looks good to me, so I'm approving it.

For future reference, we do have a library function that can convert any valid measurement in Finale to EVPUs. So for instance, you can use `measurement.convert_to_EVPUs("2s")` and get `48`.

https://www.finalelua.com/docs/library/measurement#convert_to_evpus

It might be worth adding this to make that dialog more flexible for end users. Just a thought.

Nick-Mazuk (approved):

cv-on-hub:
I hadn't noticed measurement.lua before but it looks very useful. I think I'll leave it set to EVPU in this case because the "24" unit is a good order of magnitude for the task (and is at least familiar to Finale old-timers like me). The (minimal) instructions become very messy if converting to every other possible unit!

Nick-Mazuk:
Ok, makes sense.

If you haven't gotten a chance yet, I would recommend exploring the library functions. They're all documented here: https://www.finalelua.com/docs/library

We've started to collect enough shared functions in there that sometimes you can even use those functions for like 95% of a script.

PR #191, opened 2022-05-25
change cross_staff_offset.lua to remember settings across calls

rpatters1:
Change cross_staff_offset.lua to remember settings across calls and skip dialog altogether if modifier keys are pressed on invocation. I would like to see *all* dialogs do this, but this one was an easy example to change. Eventually, once our stand-alone script generator support mixins, I would like this dialog (and others) to be modeless as well. Modeless is so much more user-friendly.

Tagging @cv-on-hub for info and comment. Also not that you can put line-breaks in the `Notes` field so that they are readable in a text editor and the line breaks are removed when the comments are transferred to the webpage. Like Markdown language.

### DIFF
--- a/src/cross_staff_offset.lua
+++ b/src/cross_staff_offset.lua
@@ -6,86 +6,103 @@ function plugindef()
     finaleplugin.Version = "v1.21"
     finaleplugin.Date = "2022/05/23"   
     finaleplugin.Notes = [[
-     When creating cross-staff notes using the option-downarrow shortcut, the stems of 'crossed' notes are reversed (on the wrong side of the notehead) and so appear too far to the right (if shifting downwards) by the width of one notehead (typically 24EVPU). This script lets you set a horizontal offset for all cross-staff notes in the currently selected region, with a different offset for non-crossed notes. For crossing to the staff below use (-24,0) or (-12,12). For crossing to the staff above use (24,0) or (12,-12). Also specify which layer number (1-4) or "all layers" (0). (This also offers a simple way to reset the horizontal offset of all notes in the selection to zero).
- ]]
+        When creating cross-staff notes using the option-downarrow shortcut, the stems of 'crossed' notes
+        are reversed (on the wrong side of the notehead) and so appear too far to the right (if shifting downwards)
+        by the width of one notehead (typically 24EVPU). This script lets you set a horizontal offset for all cross-staff
+        notes in the currently selected region, with a different offset for non-crossed notes. For crossing to the
+        staff below use (-24,0) or (-12,12). For crossing to the staff above use (24,0) or (12,-12). Also specify
+        which layer number (1-4) or "all layers" (0). (This also offers a simple way to reset the horizontal offset
+        of all notes in the selection to zero).
+
+        If you invoke the plugin with Option key (macOS) or Shift key pressed, the script uses the last settings
+        from the dialog.
+    ]]
     return "CrossStaff Offset", "CrossStaff Offset", "Offset horizontal position of cross-staff note entries"
- end
+end
  
- function user_selects_offset()
-     local current_vert, vertical_step = 10, 25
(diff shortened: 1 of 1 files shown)


cv-on-hub:
RetainLuaState is very cool - I hadn't absorbed this before. My personal problem is that I can't get my KM macro palette actions (which I use for all RGP script calls) to register a held modifier while selecting a menu item, so the modifier-bypassed script is only possible with a physical menu selection (and modifier). I can't fathom by what magic the "local" variables cross_offset, non_cross_offset and layer_number are pushed back into cross_staff_offset() between invocations, but I'll take your word for it!

In your revision, cross_staff_offset() tests for dialog_result on the 4th line, before user_selects_offset() has been invoked. Is that an error?

Also personally, modeless dialogs confuse my workflow (as in your "Transposition" scripts) - but that is probably just a question of familiarity. I'm happy if you just want to commit your revision or make more revisions before commitment. Let me know if you want me to take it "back" and make modeless etc.

rpatters1:
Lua uses a "Lua State" that holds all functions and variables. The way `RetainLuaState` works is that it never closes the Lua state, so when your script starts the next time it is in the same state it had when it exited previously.

Think of it like this. Suppose you have a simple script called `my_script.lua`:

```lua
if not global_variable then
   global_variable = "Test"
end

finenv.UI():AlertInfo(global_variable, "")
finenv.RetainLuaState = true
```

If you think of the above code not as a script but rather as a function, the net effect of `RetainLuaState = true` is that it is like calling the function over and over within a single script. Any undefined variable has the value `nil`, so the first time the "function" is called, the `global_variable` is initialized (because `nil` tests false in an if statement) and thereafter it is not.

This gets to your second question about testing `dialog_result`. The first time through the script it will be `nil`, so the script is guaranteed to execute `user_selects_offset()` the first time through. Thereafter, `dialog_result` holds the result from the previous run of the script.

It's the same for `cross_offset`, `non_cross_offset`, and `layer_number`. The first time through the script they are `nil` and thereafter they hold the values from the previous execution of the dialog (until changed by the new execution of the dialog). I changed them to be **global** variables because Lua garbage collection eventually destroys `local` variables.

This pull request does not change the dialog to modeless. That's a topic for a future discussion. I happen to believe that all dialogs should be modeless, but there are some legitimate discussions about the pros and cons and how to mitigate them.

cv-on-hub:
Thanks for such a thorough explanation - much appreciated. All makes sense, though still "feels" like wicked voodoo! Retentive settings like this are terrific, but presumably reset on each Finale reboot (unlike prefs files). Please commit your revision with my blessing and modeless discussion later on. I'll also ask KM if they can implement modifier key holds through a menu simulation.

cv-on-hub (approved):
I'm not sure if I need to perform any other action to progress this update!

cv-on-hub:
PS: having checked with the KM User Forum it seems that RGPLua doesn't get passed System Event modifier keys, only what Finale tells it on a menu call - with actual physical modifier keys. So for my KM purposes I need to add to the script an extra test for pre-existing global modifier_keys_pressed (from RGP preset) to get optional dialog-free operation. No biggie.

rpatters1:
For macOS, I believe RGP Lua *is* looking at the event modifier keys. (It gets them from `[NSApp currentEvent]`.) On Windows it is using `GetKeyState` which I believe gets the physical state of the keyboard at the time the function is called.

cv-on-hub:
That's interesting! I tried 100 different AppleScripts to "hold" shift and option wrapped around the KM menu invocation, and frequently managed to freeze those keys in place and freeze the whole computer, while "CrossStaff Offset" resolutely ignored the modifier condition.

cv-on-hub:
But it's not a vast issue as it's easy to have RGP include an alternate version that sets modifier_keys_pressed.

rpatters1:
It doesn't seem like a bridge to cross until we want to make the dialog modeless.

PR #193, opened 2022-05-27
Add Run to FCXCustomLuaWindow

ThistleSifter:
I figured the easiest way of working this out is to have something to look at, so I quickly wrote this up. Marking as draft for now.

In general, these are the things that I see that need to be agreed upon before this can be merged:
- [ ] Clearly defining ideal behaviour with modifier keys and the 'two states'
- [ ] Decide how best to handle the modal result.
- [ ] How best to handle whether region selection is required or not.

### DIFF
--- a/src/mixin/FCXCustomLuaWindow.lua
+++ b/src/mixin/FCXCustomLuaWindow.lua
@@ -36,10 +36,10 @@ local each_last_measurement_unit_change
 ]]
 function props:Init()
     private[self] = private[self] or {
-            MeasurementEdits = {},
             MeasurementUnit = measurement.get_real_default_unit(),
             UseParentMeasurementUnit = true,
             HandleTimer = {},
+            RunModelessDefaultAction = nil,
         }
 
     if self.SetAutoRestorePosition then
@@ -371,6 +371,22 @@ Removes a handler added with `AddHandleTimer`.
 end
 
 --[[
+% RegisterHandleOkButtonPressed
+
+**[Fluid] [Override]**
+Stores callback as default action for `RunModeless`.
+
+@ self (FCXCustomLuaWindow)
+@ callback (function) See documentation for `FCMCustomLuaWindow.OkButtonPressed` for callback signature.
+]]
+function props:RegisterHandleOkButtonPressed(callback)
+    mixin.assert_argument(callback, "function", 2)
+
+    private[self].RunModelessDefaultAction = callback
+    mixin.FCMCustomLuaWindow.RegisterHandleOkButtonPressed(self, callback)
+end
+
+--[[
 % ExecuteModal
 
 **[Override]**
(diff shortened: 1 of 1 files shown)


ThistleSifter:
Some thoughts and questions.

**Keyboard Behaviour**
Does the `OkButtonCanClose = modifier_keys_on_invoke` need to apply to modals as well?


**Modal Result**
Returning the modal result from `Run` was always going to end up being messy. So the best option I saw was to include the ability to add a handler for the modal result, which would be called from `ExecuteModal`, regardless of whether it was called via `Run`. Eg:
```lua
local function create_dialog()
    ...
    dialog:RegisterHandleModalResult(function(self, result)
        -- insert result handling here...
    end)
end
```
This goes further in the direction of turning the window object into a completely self-contained plugin class (which I like).


**Music Region Selection**
To my mind this one is a little trickier, because if the requiring/not requiring region selection is going to be baked in at one level, should it be baked in at others? As it stands, if `Run` displays the window as a modal, it takes care of ensuring that a region is selected for you. But if it's shown as a modeless window, the author still needs to handle it themselves. Thoughts?

rpatters1:
With a modal dialog, the Ok button always closes the window, doesn't it? I'm not sure that `OkButtonCanClose` even works with modal dialogs. But even if it does, it is a Cadillac feature in this context. (See next paragraph.)

I think a handler function for modal result is overkill. We already have handler functions for Ok and Cancel. In my mind the `Run` function is about providing consistent functionality around modeless dialogs, and `ExecuteModal` is only involved as a backup for JW Lua support. In fact, perhaps the function should be called `RunModeless`. At most it could be a read-only property that is `nil` unless and until `ExecuteModal` has been executed.

There is fundamentally a difference between a modeless and a modal about how non region selection should be handled. For a modal, we don't want to allow them to go through the whole process of filling out the window and hitting okay before they discover they failed to select anything. By contrast, for a modeless it's no big deal. They simply select something and hit okay again. (This is a key reason I think all dialogs should be modeless.) My personal preference would be to make it a property on the window that defaults to true but can be overridden by the caller by setting it to false.

As for the two 'states', I lean towards selection with modifiers only causes OK to close the window the first time. Thereafter, to open the window you have to not use the modifier key, and OK does not close the window. This seems like the most consistent behavior to me. But I could be persuaded differently.

ThistleSifter:
Ah, I think I misunderstood what you were after and I was going for some sort of master method. In that case, I've revised it and hopefully that's what you're looking for.

rpatters1:
This is looking very good. Is there any way to have `RegisterHandleOkButtonPressed` automatically set up a default action as well? Or vice versa? It seems a little cumbersome to have to register both.

I try to avoid boolean variables as parameters to functions because there is no way to know what they are doing just by looking at the function call. Could this be a property instead? (That defaults to `true`?)

ThistleSifter:
> Is there any way to have `RegisterHandleOkButtonPressed` automatically set up a default action as well? Or vice versa?
> 
I don't think so, not without it getting somewhat messy. What if someone didn't want the `OkButtonPressed` handler to be the default action? What if a script has multiple `OkButtonPressed` handlers that were `Add`ed, not `Register`ed? etc, etc

> Could this be a property instead?
> 
The issue I see here is that it has no meaning or use outside that one method and there's no reason for anything else to need to access or modify it. (DefaultAction has a similar problem really...) There's also the following option to have named parameters with default values:

```lua
-- in FCXCustomLuaWindow.lua
function props:RunModeless(params)
    local realParams = {requiresSelection = true, defaultAction = nil}
    for k, v in pairs(params) do
        realParams[k] = v
    end
    -- etc, etc
end

-- Then in script you have the following
global_dialog:RunModeless{requiresSelection = false, defaultAction = global_dialog.MyAction}
```

> Also, if `RunModeless` is Fluid, should it not be returning `self`?
> 
The mixin library takes care of that. All void mixin methods (including user-defined ones) return `self`.

rpatters1:
>What if someone didn't want the `OkButtonPressed` handler to be the default action?

The 80% case is that they *do* want it to be the default action. Isn't the whole point of this exercise to standardize the user experience for modeless dialogs? In that case, selecting the menu option with modifiers should have the same effect as hitting ok on the (now-closed) dialog. If the vast majority of time that's what we want, it should be baked in.

Another disadvantage is that now the `on_ok` function must be explicit, because it has to be registered twice. I think the call to register the default action should add it to the Ok handlers as well. Maybe there could be an optional parameter to override this.

ThistleSifter:
I've revised it again. I've gone down the route of having the registered `OkButtonPressed` handler be the default `RunModeless` action, which can be overridden or cancelled by an optional parameter.  Also changed the region selection parameter so that region selection is now checked by default.

rpatters1:
This looks good to me. I'm ready to merge and give it a whirl.

PR #194, opened 2022-05-30
Create barline_set.lua

cv-on-hub:
A nice simple script making sensible use of RGP 0.62 AdditionalMenus feature.

### DIFF
--- a/src/barline_set.lua
+++ b/src/barline_set.lua
@@ -1,0 +1,50 @@
+function plugindef()
+	finaleplugin.RequireSelection = true
+	finaleplugin.Version = "v1.1"
+	finaleplugin.Date = "2022/05/30"
+    finaleplugin.AdditionalMenuOptions = [[
+		Barline Set Double
+		Barline Set Final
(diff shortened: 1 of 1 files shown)


rpatters1:
I have no objection to adding these, however you realize there is already a function the does this more conveniently than a plugin/script if you right click the bar?

cv-on-hub:
Of course, but I really, really don't like using the mouse that much, always preferring a simple key click (or two) to invoke menu items. (Especially popups). If you think it seems a silly option then don't bother, but the speed of my workflow owes much to the simplification of regular actions.

Nick-Mazuk:
@cv-on-hub You beat me to it! This was one of the scripts I was going to add, but happy to see that you've gone ahead and wrote it up!

On my end, I have several of these barline scripts broken out into separate scripts (before RGP v0.62 launched with this feature), and haven't gotten around to combining them into a single script.

Approving and merging.

Nick-Mazuk (approved):

PR #195, opened 2022-05-31
2 new short scripts

cv-on-hub:
Simple layer-based menu-driven scripts to Mute/Unmute and Hide/Unhide selected region.


Nick-Mazuk (approved):
LGTM

cv-on-hub:
Thanks Nick. No problem. I had started to wonder if there was a problem so profound that it couldn't be mentioned!

PR #196, opened 2022-06-02
Create notes_tie.lua

cv-on-hub:
New small script. You can tell by now that I'm moving, step by step, to eliminate dependency on JW plugins.

### DIFF
--- a/notes_tie.lua
+++ b/notes_tie.lua
@@ -1,0 +1,72 @@
+function plugindef()
+    finaleplugin.RequireSelection = true
+    finaleplugin.Author = "Carl Vine"
+    finaleplugin.AuthorURL = "http://carlvine.com"
+    finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
+    finaleplugin.Version = "v0.58"
+    finaleplugin.Date = "2022/06/01"
+    finaleplugin.AdditionalMenuOptions = [[ Untie Notes ]]
+    finaleplugin.AdditionalUndoText = [[    Untie Notes ]]
+    finaleplugin.AdditionalPrefixes = [[    untie_notes = true ]]
+    finaleplugin.AdditionalDescriptions = [[ Untie all notes in the selected region ]]
+    finaleplugin.MinJWLuaVersion = 0.62
+    finaleplugin.Notes = [[ 
+    Ties notes in adjacent entries if matching pitches are available. 
+    RGPLua (0.62 and above) creates a companion menu item, UNTIE Notes.
+    ]]
+    return "Tie Notes", "Tie Notes", "Tie suitable notes in the selected region, with matching Untie option"
+end
+
+-- default to "tie" notes for normal operation
+untie_notes = untie_notes or false
+
+local function note_pitch(note)
+    -- thanks for this bitwise math magic, Robert G Patterson!
+    return bit32.bor(note.RaiseLower, bit32.lshift(note.Displacement, 16))
+end
+
+local function tie_notes_in_selection()
+    local rgn = finenv.Region()
+    local chords = {} -- collate pitches in each chord
+    local slot = 1 -- count each entry slot
+
+    for entry in eachentry(rgn) do -- just collating pitches, not SAVING entries
+        chords[slot] = {} -- start with empty chord
+        if entry:IsNote() then
+            for note in each(entry) do
+                chords[slot][note_pitch(note)] = true -- flag this pitch in the chord
+            end
+        end
+        slot = slot + 1
+    end
+
+    slot = 1 -- restart from slot 1
+    for entry in eachentrysaved(rgn) do
+        if entry:IsNote() then
+            for note in each(entry) do
+                if nil == chords[slot+1] then -- last chord
+                    break
+                end
+                if chords[slot+1][note_pitch(note)] then
+                    note.Tie = true -- tie the note forward
+                end
+            end
+        end
+        slot = slot + 1
+    end
+end
+
+function untie_notes_in_selection()
+    for entry in eachentrysaved(finenv.Region()) do
+        for note in each(entry) do
+            note.Tie = false
+            note.TieBackwards = false
+        end
+    end
+end
+
+if untie_notes then
+    untie_notes_in_selection()
+else
+    tie_notes_in_selection()
+end


rpatters1:
This script will fail to correctly create a tie if

- There is a v2 launch between a tied-from and tied-to note.
- The same note occurs in two different layers in certain configurations.
- There is a key change between the tied-from and tied-to notes.

Also, it is dependent on the order that `eachentry` serves up entries. This mainly could get you in trouble if there are multiple layers, because the measure position resets to zero. You could end up trying to tie to a note that is before the tied-from note.

At the very least, I would like to see V1/V2 support added and only tie together two notes from the same layer. (V1/V2 support will complicate this script a good bit, but it needs to be there.)

I would consider changing this to use `FCNoteEntryLayer` instead of `eachentrysaved`. That will allow you to run one layer at a time easily, and by using the `FCNoteEntry.Next` method, you can guarantee the entry sequence.

Support for key changes in the middle of a tie is considerably more complicated and perhaps not important enough to attempt.

There is a private function `tie_span` in the `tie.lua` library. It wrestles with these issues. Unfortunately you can't use it as-is because it is finding the span of an already-existing tie so it is counting on the `Tie` and `TieBackwards` properties already being set.

rpatters1:
No need to remove it. Just commit more changes until it's ready to merge.

rpatters1:
Instead of `FCNoteEntryLayer` you could pass the 1-based layer number as the second parameter to `eachentry` or `eachentrysaved`. That also gives you control over layers. (This is an apparently little known feature of these built in functions. I really need to document them better than perhaps Jari did. There are some that I've never seen used that are really useful.)

However, the two built-in functions serve up entries one cell at a time, whereas `FCNoteEntryLayer` gets them all at once. This allows the `Next` method to cross barlines the way you need it to for ties.

Also, I'm not sure if you saw my comment about grace notes. Grace notes do not tie to other grace notes. They tie to the first non-grace-note entry.

cv-on-hub:
I used layer filtering in eachentry on my layer_mute and layer_hide scripts. (Pull request #195 - pending). Will examine the options!

cv-on-hub:
I've been racking my brain about this, and need you to confirm that I was trying to achieve the impossible.

I think I have most of this in place now. But I have a V2 entry that extends to the end of a measure, which I was trying to tie to a matching V2 entry at the beginning of the next measure. But if you set the TIE bit on the first note, it seems as though Finale wants to tie it to the "subsequent" V1 note, occurring earlier in the same measure. Is that true? And so I should just stop trying to get V2 to tie across barlines?

(And yes - FCNoteEntryLayer handles gaps between measures more usefully than eachentry() )

rpatters1:
If the edu length of the V2 sequence is not exactly equal to the V1 entry that launched it, Finale does weird things. It could be that it might try to tie a V2 note to a V1 entry before it. But I think you should focus only on the case where the length of the V2 sequence matches (or is no more than) the length of the the V1 entry that launched it. In that case, a tie from the last V2 entry can be tied to either the following V1 entry or to a V2 entry launched from that next V1 entry. And the next V1 entry can be in the next bar.

I strongly encourage you to use the logic in `tied_to` and`equal_note` from the `tie.lua` library in this repo. The only change you need to make is to compare the actual `Displacement` and `RaiseLower` rather than the staff position and current state of the tie flags. That logic is the result of tens of hours of experimentation with Finale back when I wrote the Tie Mover plugins (some 20-odd years ago).

Then the pseudocode inside your loop can be like:

```lua
next_entry = next_entry:Next()
if not next_entry then
   break
end
for each note in (next_entry) do
   local tied_to_note = tied_to(note)
   if tied_to_note then
      note.Tie = true
      tied_to_note.TieBackwards = true
   end
end
```
I'm tempted to rewrite the code in `tie.lua` so that it can be reused by both, but there are some challenges to trying to make the functions generally applicable to all callers.

That pseudocode should get it working in all cases except when a key change happens in the middle of a tie. That's a bridge you can cross separately if you decide to. (Perhaps using the note string function instead of disp/raiselower.)

cv-on-hub:
I thought I'd show you where I'm up to on the whole script, which has become massively simpler when no longer trying to tie V2 across barlines. In fact I've specifically exclude chasing entries "leaving" V2 from tying to anything, as that seems to lead inexorably towards tying (backwards) to previous notes. 

I think I'm using the same logic as tied_to(note), though it's distributed differently. I'm using entry:FindPitch(note) as a convenience - will that suffice? I haven't been using note.TieBackwards since I thought that was handled by eachentrysaved(), but is it advisable under FCNoteEntryLayer? Easy enough to implement in my tie_between_entries().

I've attached the sample file I've been using to test V1/V2 behaviour. Fascinating that copying whole measures (with option click) corrupts the distribution between V1 & V2, favouring shorter stretches of V2 interrupted by intervening V1 entries. Easy to see why V2 was so quickly deprecated.

Which leaves key changes across ties, and I figure anyone who wants to do that shouldn't expect a tie automation script to compensate for their cavalier behaviour.

[tie_test_v2.musx.zip](https://github.com/finale-lua/lua-scripts/files/8839070/tie_test_v2.musx.zip)

rpatters1:
`FindPitch` is an excellent find. It will kind of work with key signatures as well. (It may try to tie F# to Gb across a key sig change, but we can live with that.)

I did not mean to imply that V2 doesn't tie across barlines. V2 absolutely *can* tie across a barline if it is the last entry in the bar.

You should also set `TieBackwards` on the tied-to notes.

I find this code overly complicated and suggest you do the way I suggested. My code is the result ofm **extensive** experimentation with ties. Also, I didn't see anywhere that you are skipping grace notes as candidates for tied-to. Perhaps I missed it.

cv-on-hub:
Tried simplifying back to the precepts of tie.tied_to() and tie.equal_note() but without tracking the transitions between V1 & V2 this inevitably failed to tie two V1 notes where the first is a Voice2Launch, and always tied an "ending" V2 entry to the "next" V1 entry that follows it in the Edit Frame note sequence but precedes it in the measure.

Creating ties has different needs to tracking existing ties (as in tied_to() and equal_note()) but I can't see another way to do this without the complexity of tracking V1/2 transitions, one way or another. If you run this script over the Finale file I attached before it works at everything except tying V2 over barlines, and I still can't get that to happen in Finale using any available tool whatsoever!

rpatters1:
Your sample file is too perverse. You will not be able to tie the V2s across the barline in that file even using Finale's UI. (I tried it with Speedy.) Don't worry about cases where the V2 sequence is longer than the V1 entry that launched it. Finale is not designed to do that.

That said, you did uncover a bug in my `tied_to` function that I will have to correct, even in the Tie Mover plugin. That is that a V1 can never tie to a V2 note in an entry that it launches. Here is a version of your code I wrote that handles (normal) V2 across barline no problem. I also fixed it to skip entries not contained in the selected region (on both ends).

```lua

local note_entry = require('library.note_entry')

-- default to "tie" notes for normal operation
untie_notes = untie_notes or false

local function tied_to(note)
    if not note then
        return nil
    end
    local next_entry = note.Entry
    if next_entry then
        if next_entry.Voice2Launch then
            next_entry = note_entry.get_next_same_v(next_entry)
        else
            next_entry = next_entry:Next()
        end
        if next_entry and not next_entry:IsRest() and not next_entry.GraceNote then
            local tied_to_note = next_entry:FindPitch(note)
            if tied_to_note then
                return tied_to_note
            end
            if next_entry.Voice2Launch then
                local next_v2_entry = next_entry:Next()
                tied_to_note = next_v2_entry:FindPitch(note)
                if tied_to_note then
                    return tied_to_note
                end
            end
        end
    end
    return nil
end

function tie_notes_in_selection()

    for staff_number = region.StartStaff, region.EndStaff do
        for layer_number = 0, 3 do  -- run through layers [0-based]
            local saved_v1_entry = nil -- v1 entry held while v2 running
            local v2_is_active = false
            local entry_layer = finale.FCNoteEntryLayer(layer_number, staff_number, region.StartMeasure, region.EndMeasure)
            entry_layer:Load()
            for entry in each(entry_layer) do
                if not entry:IsRest() and region:IsEntryPosWithin(entry) then
                    for note in each(entry) do
                        local tied_to_note = tied_to(note)
                        if tied_to_note and region:IsEntryPosWithin(tied_to_note.Entry) then
                            note.Tie = true
                            tied_to_note.TieBackwards = true
                        end
                    end
                end
            end
            entry_layer:Save()
        end
    end
end
```

cv-on-hub:
That works nicely, and very neatly, and V2 ties over barlines as you promised if the bar-end-V2 is a fresh V2 entry at that point. note_entry.get_next_same_v() is very nice. The script is now mostly yours, so should I list you as author or make it authorless?

My "perverse" use of V2 is how I always used V2 before Layers were introduced (1992?) .. as if V2 were Layer 2, and I don't recall any restrictions from Coda at the time about the extent of V2 spans. Mind you I probably didn't read that bit of the manual. The current manual is also mute on that point, in the only place I can find that it mentions V2 at all!

rpatters1:
Attribute it however you like. At some point I would like to combine the `tied_to` function back into `tie.lua`. You certainly deserve credit for thinking of it and for discovering the `FindPitch` method.

cv-on-hub:
That's kind of you, in which case I'll leave it with the dual accreditation if that's OK with you.

rpatters1:
I have one other thought about the tie-removal routine. I don't think you should be removing tie-ends if you are not also removing tie-starts. You can actually delete entirely the `untie_notes_in_selection` routine and change the two "payload" line from

```lua
                            note.Tie = true
                            tied_to_note.TieBackwards = true
```

to

```lua
                            note.Tie = not untie_notes
                            tied_to_note.TieBackwards = not untie_notes
```

This will force both ends of the tie to be in the selected region.

cv-on-hub:
A beautifully elegant solution Robert.

rpatters1:
I merged and then realized that we may want the untie side to remove ties to nothing. It would be easy enough to do, but it will potentially leave orphaned tie-ends without a lot of fussy extra code. I mean to revisit this when I enhance `tie.lua` to make a single `tied_to` function. I can think about it then, if you like.

cv-on-hub:
I also thought of unmatched tie ends and artefacts from other actions that one could reasonably expect "untie" to handle indiscriminately. For my own use I'll re-enable the indiscriminate untie routine. Also Nick just pointed out that this script is not in /src. I'll move it back there now.

PR #197, opened 2022-06-02
Create note_ends.lua

cv-on-hub:
another JW plugin substitution

### DIFF
--- a/note_ends.lua
+++ b/note_ends.lua
@@ -22,12 +22,12 @@ end
 eighth_notes = eighth_notes or false
 
 function expand_note_ends()
-	local delete_next = false
+	local should_delete_next = false
     local note_value = eighth_notes and finale.NOTE_8TH or finale.NOTE_QUARTER -- [ 512 / 1024 ]
     local beat_duration = note_value
 	local measure_number = 0
 	local measure = finale.FCMeasure()
-    local compound_meter = false
+    local is_compound_meter = false
 	
 	for entry in eachentrysaved(finenv.Region()) do
 	    if measure_number ~= entry.Measure then -- calculate beat duration for each new measure
@@ -35,12 +35,12 @@ function expand_note_ends()
 	        measure:Load(measure_number)
 	        local time_sig = measure:GetTimeSignature()
 	        beat_duration = time_sig:CalcLargestBeatDuration()
-            compound_meter = (beat_duration % 3 == 0) 
+            is_compound_meter = (beat_duration % 3 == 0)
 	    end
 	    
-		if delete_next then -- last note was expanded
+		if should_delete_next then -- last note was expanded
 			entry.Duration = 0 -- so delete this rest
-			delete_next = false -- and start over
+			should_delete_next = false -- and start over
             -- OTHERWISE
 		elseif entry:IsNote() -- this is a note
                 and entry:Next() -- with a following entry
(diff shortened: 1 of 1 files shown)


note_ends.lua
```diff
@@ -0,0 +1,79 @@
...
+        It has 2 modes: Eighth Notes and Quarter Notes.
+]]
+    return "Note Ends Quarters", "Note Ends Quarters", "Change smaller notes followed by rests into quarter notes"
+end
+
+-- default to quarter notes for normal operation
+eighth_notes = eighth_notes or false
+
+function expand_note_ends()
+	local delete_next = false
```
Nick-Mazuk, line 25: Perhaps `should_delete_next` instead of `delete_next`?

note_ends.lua
```diff
@@ -0,0 +1,79 @@
...
+	local measure = finale.FCMeasure()
+    local compound_meter = false
+	
+	for entry in eachentrysaved(finenv.Region()) do
+	    if measure_number ~= entry.Measure then -- calculate beat duration for each new measure
+	        measure_number = entry.Measure
+	        measure:Load(measure_number)
+	        local time_sig = measure:GetTimeSignature()
+	        beat_duration = time_sig:CalcLargestBeatDuration()
+            compound_meter = (beat_duration % 3 == 0)
```
Nick-Mazuk, line 38: Perhaps `is_compound_meter`?

note_ends.lua
```diff
@@ -0,0 +1,79 @@
...
+			entry.Duration = 0 -- so delete this rest
+			delete_next = false -- and start over
+            -- OTHERWISE
+		elseif entry:IsNote() -- this is a note
+                and entry:Next() -- with a following entry
+                and entry:Next():IsRest() -- that is a rest
+                and entry.Duration < note_value -- this note is too short
+            then
+
+            local valid = true -- refine the validity criteria
```
Nick-Mazuk, line 51: What is valid? Perhaps `is_valid_note` or  `is_valid_entry`?

note_ends.lua
```diff
@@ -0,0 +1,79 @@
...
+            if beat_duration < (position_in_beat + note_value) -- not enough room in beat for expanded note
+                or note_boundary ~= 0  -- not falling on an allowed duration boundary
+                then
+                valid = false -- don't expand
+                if not compound_meter and not eighth_notes and note_boundary == 0 and start_beat % 2 == 0 then
+                    valid = true  -- special case for quarters on beats 1 & 3
+                end
+            end
+
+            if valid then
```
Nick-Mazuk, line 65: Just from the code itself, I'm unsure why you'd want to modify a valid entry? Given the context of knowing what the script should do it makes sense. But perhaps the code itself could be a bit clearer.

Nick-Mazuk (commented):
Overall looks good, and I'm always a fan of replacing JW Plugins with Lua (especially since JW plugins don't work with a native Apple Silicon build of Finale).

Just a couple nit-picks with code style.

One general rule of thumb with booleans is that the variable should signify that it's a boolean. Often, you'll want to prefix them with `is_` or `should_`, or something else that describes the purpose of the variable. That way, you can read `should_count_to_ten` and instantly know that variable is a boolean.

note_ends.lua
cv-on-hub: Many thanks for the advice Nick. Corrected script follows.

Nick-Mazuk (approved):
LGTM

Nick-Mazuk:
@cv-on-hub Don't know how I just noticed this, but can you create a follow-up PR to move the `note_ends.lua` and `note_tie.lua` files into the `/src` folder?

PR #199, opened 2022-06-04
Add library to get client details and capabilities

Nick-Mazuk:
For the purposes of Finale Lua, the client is the Finale application that's running on someones machine. Therefore, the client has details about the user's setup, such as their Finale version, plugin version, and operating system.

One of the main uses of using client details is to check its capabilities. As such, the bulk of this library is helper functions to determine what the client supports.

There are a few reasons to add a library to figure out what a client supports rather than doing individual checks:

1. Improved readability. It's a lot easier to understand the purpose of `if client.supports_retained_state() then` than `if finenv.IsRGPLua then`
2. Improved stability. As we add more capabilities, it makes it easier for us to check if a client supports something. Therefore, it's easier for us to write scripts that continue to work on older plugin/Finale versions, or fail gracefully.
3. If a new version of RGP Lua or Finale comes out that actually removes a capability (unlikely, but possible), we can just update the single helper function and all scripts will continue to work. The alternative that we'd need to update every individual script (and since there typically isn't a good find-and-replace mechanism for these if checks, it would be a very manual process).

In this PR, I've also gone through and replaced the obvious existing capability checks with the functions I added to this library.

CC @ThistleSifter, @cv-on-hub, and @jwink75 since if merged, this could impact how you'd want to write your scripts.

### DIFF
--- a/src/library/clef.lua
+++ b/src/library/clef.lua
@@ -5,6 +5,8 @@ A library of general clef utility functions.
 ]] --
 local clef = {}
 
+local client = require("library.client")
+
 --[[
 % get_cell_clef
 
@@ -53,18 +55,6 @@ function clef.get_default_clef(first_measure, last_measure, staff_number)
 end
 
 --[[
-% can_change_clef
-
-Determine if the current version of the plugin can change clefs.
-
-: (boolean) Whether or not the plugin can change clefs
-]]
-function clef.can_change_clef()
-    -- RGPLua 0.60 or later needed for clef changing
-    return finenv.IsRGPLua or finenv.StringVersion >= "0.60"
-end
-
---[[
 % restore_default_clef
 
 Restores the default clef for any staff for a specific region.
@@ -74,7 +64,7 @@ Restores the default clef for any staff for a specific region.
 @ staff_number (number) The staff number for the cell
 ]]
 function clef.restore_default_clef(first_measure, last_measure, staff_number)
-    if not clef.can_change_clef() then
+    if not client.supports_clef_changes() then
         return
     end
 
--- a/src/library/client.lua
+++ b/src/library/client.lua
@@ -1,0 +1,113 @@
+--[[
+$module Client
+
+Get information about the current client. For the purposes of Finale Lua, the client is
+the Finale application that's running on someones machine. Therefore, the client has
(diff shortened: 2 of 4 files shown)


rpatters1 (commented):
I'm unhappy that you have allowed the code formatter to destroy many of the carefully planned line breaks throughout the code. FWIW, I have removed that extension in my environment because I found the damage it caused to be far worse than the benefit gained. (I kept the delinter, which is great.)

Did you mean to include double-underscores in some of the function names? It looks weird to me.

I'm generally happy with the idea of doing this.

rpatters1:
Also, transpose_chromatic and transpose_by_step are drastically changed in a pending PR, so please don't modify them.

Nick-Mazuk:
> I'm unhappy that you have allowed the code formatter to destroy many of the carefully planned line breaks throughout the code. FWIW, I have removed that extension in my environment because I found the damage it caused to be far worse than the benefit gained. (I kept the delinter, which is great.)

Ok. I'll fix it so this PR doesn't overwrite your formatting. FYI I did find another code formatter I'm going to play with (https://github.com/prettier/plugin-lua). In general, I've had very, very good results with prettier in other languages like JavaScript and recently found out theirs a Lua plugin. In JavaScript at least, it can understand intentional line breaks (like making every table item being on a new line) and will preserve them while simultaneously fixing all the little details that are annoying to do manually. So that might end up being the solution.



> Did you mean to include double-underscores in some of the function names? It looks weird to me.

Will fix.

> Also, transpose_chromatic and transpose_by_step are drastically changed in a pending PR, so please don't modify them.

I'll remove them from this PR.

ThistleSifter:
Can you leave out changes to the mixins? I'm currently rewriting large chunks of them for #212. I'll go back and add the client capability tests afterwards.

Nick-Mazuk:
> Can you leave out changes to the mixins? I'm currently rewriting large chunks of them for #212. I'll go back and add the client capability tests afterwards.

When I get time to refactor this PR, I think the best solution is likely to only add the library. Then in subsequent PRs we can use the library. Since the mixins issue you bring up isn't the only place this is affected.

rpatters1 (approved):
Looks good to me. Please feel free to merge.

PR #200, opened 2022-06-04
Add script tp change meters

Nick-Mazuk:
Nothing more to add. Title says it all.

### DIFF
--- a/src/meter_change.lua
+++ b/src/meter_change.lua
@@ -1,0 +1,111 @@
+function plugindef()
+    finaleplugin.Author = "Nick Mazuk"
+    finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
+    finaleplugin.Version = "1.0.0"
+    finaleplugin.Date = "June 4, 2022"
+    finaleplugin.CategoryTags = "Meter"
+    finaleplugin.MinJWLuaVersion = 0.62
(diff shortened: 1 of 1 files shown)


rpatters1 (commented):
I just recently discovered you can replace 

```lua
        local allMeasures = finale.FCMeasures()
        allMeasures:LoadAll()
        for m in each(allMeasures) do
```
with

```lua
for m in loadall(finale.Measures()) do
```

It's been there all along in JW Lua. (There's also a similar `loadallforregion` built-in func.)

Also, don't the compound meters need a display-as as well?

rpatters1 (commented):
Also, if the `AdditionalUndoText` is identical to `AdditionalMenuOptions` then you can omit the undo text block.

Nick-Mazuk:
That's an awesome JW Lua feature! Will remove a ton of code.

I don't believe the compound meters need a display-as. I've been using this for years and haven't had any issues with compound meters as is.



> Also, if the `AdditionalUndoText` is identical to `AdditionalMenuOptions` then you can omit the undo text block.

Good to know. Will remove it.

rpatters1:
Yes, on further reflection you don't need the display time sig. That would be for compound meters.

`loadallinregion` only works for collections that have a `LoadAllInRegion` method. (And `loadall` only works for collections with `LoadAll`.) I don't think `loadallinregion` will work in the script here, more's the pity, but `loadall` will.

PR #201, opened 2022-06-06
Create note_ends.lua

cv-on-hub:
Re-submitting this script in the correct location! (/src).


Nick-Mazuk:
Could you try rebasing the branch? I think it may be out of date. I'm forgetting how to do this in GitHub desktop, but here's how you do it via the terminal.

```
git pull --rebase upstream master
```

Essentially, rebasing makes your commit be based on the most recent commit in the main repository rather than an older one.

https://git-scm.com/book/en/v2/Git-Branching-Rebasing

PR #202, opened 2022-06-06
re-create notes_tie.lua

cv-on-hub:
Moving the script to the correct location. (/src).


cv-on-hub:
PS: as soon as this lot is cleared I'll rebuild my GitHub connection. Confusion has developed between Git and the script editor.

Nick-Mazuk:
Same comment here about rebasing as from the comment in the other PR about rebasing.

rpatters1:
I also noticed an issue, which is that you can't iterate from `StartStaff` to `EndStaff` on a region. The staff numbers can be in any order. Instead, iterate from `StartSlot` to `EndSlot` and then use `CalcStaffNumber` to get the staff number inside the loop.

cv-on-hub:
Indeed - which is the workaround I used in all the Staff_Explode scripts. Will fix in the next iteration of notes_tie once my Desktop GitHub stops spewing redundant copies of old scripts.

Nick-Mazuk (approved):
LGTM

PR #205, opened 2022-06-08
Update notes_tie.lua

cv-on-hub:
Converting Staff loop into Slot loop as suggested by RGP. Also testing the stability of my local repo.


Nick-Mazuk:
LGTM, but I'll let @rpatters1 review and merge it since it was his suggestion.

PR #206, opened 2022-06-09
Create secondary_beam_breaks.lua

cv-on-hub:
New script to break secondary beams, including consideration for compound meters.

### DIFF
--- a/src/secondary_beam_breaks.lua
+++ b/src/secondary_beam_breaks.lua
@@ -1,0 +1,76 @@
+function plugindef()
+    finaleplugin.RequireSelection = true
+    finaleplugin.Author = "Carl Vine after Jari Williamsson"
+    finaleplugin.AuthorURL = "http://carlvine.com"
+    finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdobreak_beams/zero/1.0/"
+    finaleplugin.Version = "v1.2"
+    finaleplugin.Date = "2022/06/06"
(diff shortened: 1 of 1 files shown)


Nick-Mazuk (approved):
LGTM. This script looks awesome! I'm definitely going to use it myself.

rpatters1:
This is duplicating a function in TGTools, but there's nothing wrong with that. While the script as-is works in the most common situations, a future refinement might be to only break the "correct" number of beams. Right now you are basically assuming that every subgroup will add up to an 8th (or dotted 8th in compound time). But technically, you should only break the number of beams to make the subgroup equal its grouping. So 64th notes in 4/8 should break through 2 beams and leave two beams solid, because each subgroup of 4 64ths adds up to a 16th.

Admittedly, we are pretty far into the weeds here. As-is, this will definitely cover the vast majority of situations.

rpatters1:
P.S. I'm pretty sure the TGTools function deals with this by letting the user specify how many beams to break.

cv-on-hub:
You speak the truth Robert. I used TGT beam-breaker for years but was always frustrated by its inability to deal with compound meters ... which is precisely why I made this variant. I'm also being entirely selfish - it is streamlined for the rhythmic structures I use all the time in my own music. The other thing to add, which I pondered at length before discarding as too mind-blowingly complex, was to divide beams appropriately within tuplets. Fine for triplets but a big mess for anything more complex.

PR #207, opened 2022-06-10
Rgp/refactor tie to tie from

rpatters1:
This pull request refactors the `tied_to` and `tied_from` functions in `tie.lua` so that external callers can use them. It then changes `notes_tie.lua` to call the new `calc_tied_to` function instead of its own copy. In addition, it refactors the "Untie Notes" option in `notes_tie.lua` as follows:

- Any tie with both endpoints selected is removed.
- Any tie to/from nothing on a selected note is removed.
- Any `FCTieMods` are removed.

That last point is worthy of discussion before checking it in. Should `FCTieMods` be removed? Finale, fwiw, does *not* remove them when you delete a tie with Speedy. The choice is between

- a completely clean start if you add back a tie
- a cleaner Finale file if you don't

vs.

- any manual edits are preserved if you add back a tie
- accumulating cruft in your document if you don't

Thoughts? (I'm on the fence on this one, though I obviously lean towards removal.)

@cv-on-hub

### DIFF
--- a/src/library/tie.lua
+++ b/src/library/tie.lua
@@ -9,60 +9,38 @@ local tie = {}
 local note_entry = require('library.note_entry')
 
 -- returns the equal note in the next closest entry or nil if none
-local equal_note = function(entry, target_note, for_tieend)
-    if entry:IsRest() then
-        return nil
+local equal_note = function(entry, target_note, for_tieend, tie_must_exist)
+    local found_note = entry:FindPitch(target_note)
+    if not found_note or not tie_must_exist then
+        return found_note
     end
-    -- By using CalcStaffPosition we can support a key change in the middle of a tie. But it is at the cost
-    -- of not supporting a clef change in the middle of the tie. A calculation comparing normalized concert
-    -- pitch is *much* more complicated code, and clef changes in the middle of ties seem like a very limited
-    -- use case.
-    local target_staffline = target_note:CalcStaffPosition()
-    for note in each(entry) do
-        local this_staffline = note:CalcStaffPosition()
-        if this_staffline == target_staffline then
-            if for_tieend then
-                if note.TieBackwards then
-                    return note
-                end
-            else
-                if note.Tie then
-                    return note
-                end
-            end
+    if for_tieend then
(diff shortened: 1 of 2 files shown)


cv-on-hub:
I can't garner the full impact of `tie.lua` to comment more helpfully on those changes, but I endorse eliminating file cruft wherever possible. (Finale leaves enough around on its own). As long as `notes_tie.lua` is becoming more robust I'm in favour.

ThistleSifter:
Out of curiosity, do you have any inkling as to why Finale would leave the `FCTieMods`?

rpatters1:
I'm fairly certain Finale doesn't remove any entry details until an entry is actually deleted. Whether this was a considered choice could probably only be answered by Phil Farrand. If so, perhaps it is to preserve manual edits if one changes one's mind. But I lean towards thinking it was an oversight, pure and simple.

Nick-Mazuk:
My preference is to remove cruft whenever possible. Though I have an extreme bias towards simplicity.

I'm in favor of removing the FC Tie mods.

src/library/tie.lua
```diff
@@ -6,51 +6,57 @@ as well as providing other useful information about ties.
 ]] --
 local tie = {}
 
+local note_entry = require('library.note_entry')
+
 -- returns the equal note in the next closest entry or nil if none
-local equal_note = function(entry, target_note, for_tieend)
-    if entry:IsRest() then
-        return nil
+local equal_note = function(entry, target_note, for_tieend, tie_must_exist)
```
Nick-Mazuk, line 12: Nit pick, `for_tie_end` seems more readable than `for_tieend`.

Nick-Mazuk (approved):
One small nit. Feel free to merge when fixed.

rpatters1:
To be clear, a "Tie End" in Finale is one of the little stubs at the beginning of a system when a tie crosses from the previous system. They have separately editable `FCTieMod` records. But now that `equal_note` and `calc_tie_span` are public functions, the boolean has wider applicability, so I changed the name.

PR #208, opened 2022-06-13
Create midi_duration.lua

cv-on-hub:
New script to change playback start/stop times

### DIFF
--- a/src/midi_duration.lua
+++ b/src/midi_duration.lua
@@ -3,8 +3,8 @@ function plugindef()
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "http://carlvine.com"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "v1.2"
-    finaleplugin.Date = "2022/06/12"
+    finaleplugin.Version = "v1.3"
+    finaleplugin.Date = "2022/06/13"
     finaleplugin.CategoryTags = "MIDI"
     finaleplugin.Notes = [[
     Change the playback START and STOP times for every note in the selected area on one or all layers. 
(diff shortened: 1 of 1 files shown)


src/midi_duration.lua
```diff
@@ -0,0 +1,97 @@
...
+    
+    if start_offset < -9999 or start_offset > 9999 or stop_offset < -9999 or stop_offset > 9999 then
+        show_error("bad_offset", start_offset .. " / " .. stop_offset)
+        return
+    end
+    if layer_number < 0 or layer_number > 4 then
+        show_error("bad_layer_number", layer_number)
+        return
+    end
+    if nil ~= finenv.RetainLuaState then
```
Nick-Mazuk, line 75: Ok, I guess this code review is just going to be me being picky about style.

This line should be:

```lua
finenv.RetainLuaState ~= nil
```

The way you originally wrote it is called a [yoda condition](https://en.wikipedia.org/wiki/Yoda_conditions). While they are valid syntax and there are a few minor advantages of them, they can in general make it harder to read code.

Nick-Mazuk (commented):
This review ended up with me just being picky about syntax, but the overall script looks good.

cv-on-hub:
No problem with picky. The remedy is easy and I'm intrigued by your thought processes. 

I recall you saying that an `is_` test should return true/false, so I've made `is_ok` do that. I only used `RetainLuaState ~= nil` because I've seen that formulation in so many FinLua scripts and thought it was a Lua thing I didn't know about. (Compilation efficiency?) It jars my historically ingrained Perl sensibility something awful.

Nick-Mazuk:
> I only used `RetainLuaState ~= nil` because I've seen that formulation in so many FinLua scripts and thought it was a Lua thing I didn't know about. (Compilation efficiency?)

Yeah, using `nil ~= finenv.RetainLuaState` is nothing special in Lua.

Nick-Mazuk (approved):
LGTM.

PR #209, opened 2022-06-13
secondary beam breaks handle Ind Time Sig and correct number of beams

rpatters1:
This is a couple of refinements on the new secondary beam break script

- Only break the correct number of beams
- Handle Independent Time Signatures

@cv-on-hub

### DIFF
--- a/src/secondary_beam_breaks.lua
+++ b/src/secondary_beam_breaks.lua
@@ -23,14 +23,15 @@ clear_breaks = clear_breaks or false
 
 function break_secondary_beams()
     local measure_number = 0
+    local staff_number = 0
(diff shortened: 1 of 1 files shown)


rpatters1:
Ideas for future possible enhancements:

- Support uneven meters like 7/8.
- Don't break beams if it will only end up with 2 entries in the subgroup. (optionally, perhaps).

Nick-Mazuk (approved):
LGTM, but I'll let @cv-on-hub view it before merging.

cv-on-hub (approved):
This is all very solid and welcomed, especially the entry:Previous() test and allowance for independent meters.

Nick-Mazuk:
Sounds good. Merging.

PR #211, opened 2022-06-13
Create midi_velocity.lua

cv-on-hub:
New script to set key velocity on individual layers. Uses RetainLuaState.


Nick-Mazuk (approved):
LGTM. (in case you haven't figured this out, it LGTM stands for "looks good to me").

PR #212, opened 2022-06-14
Restore control state

ThistleSifter:
This PR allows control state to persist across executions without the script author needing to perform any additional handling.

At the moment I've just done the base control properties from `FCControl`, but before I continue with the control-specific properties, I was wondering if someone could test it on Mac, since I only have a Windows machine (I'm assuming it should be fine but this proved to be more of a headache than I initially anticipated, so I would like to make sure).

Here is a test script (also uncomment the line near the bottom for additional testing):
```lua

function create()
    local window = mixin.FCXCustomLuaWindow()
    window:CreateEdit(0, 0, "edit")

    window:CreateButton(0, 40):SetWidth(75):SetText("Toggle Enable"):AddHandleCommand(function(self)
        local edit = self:GetParent():GetControl("edit")
        edit:SetEnable(not edit:GetEnable())
    end)
    
    window:CreateButton(80, 40):SetWidth(75):SetText("Toggle Visible"):AddHandleCommand(function(self)
        local edit = self:GetParent():GetControl("edit")
        edit:SetVisible(not edit:GetVisible())
    end)

    window:CreateButton(0, 70):SetText("Move Left"):AddHandleCommand(function(self)
        local edit = self:GetParent():GetControl("edit")
        edit:SetLeft(edit:GetLeft() - 10)
    end)

    window:CreateButton(80, 70):SetText("Move Right"):AddHandleCommand(function(self)
        local edit = self:GetParent():GetControl("edit")
        edit:SetLeft(edit:GetLeft() + 10)
    end)

    window:CreateButton(0, 100):SetText("Narrower"):AddHandleCommand(function(self)
        local edit = self:GetParent():GetControl("edit")
        edit:SetWidth(edit:GetWidth() - 10)
    end)

    window:CreateButton(80, 100):SetText("Wider"):AddHandleCommand(function(self)
        local edit = self:GetParent():GetControl("edit")
        edit:SetWidth(edit:GetWidth() + 10)
    end)

    window:CreateButton(0, 130):SetText("Shorter"):AddHandleCommand(function(self)
        local edit = self:GetParent():GetControl("edit")
        edit:SetHeight(edit:GetHeight() - 10)
    end)

    window:CreateButton(80, 130):SetText("Taller"):AddHandleCommand(function(self)
        local edit = self:GetParent():GetControl("edit")
        edit:SetHeight(edit:GetHeight() + 10)
    end)

    return window
end

dialog = dialog or create()

-- This should make the edit control one notch shorter on each opening (and invisible on the very first opening, since GetHeight returns 0)
-- (tests the ability to edit control values outside of window execution and still preserve correct state)
--dialog:GetControl("edit"):SetHeight(dialog:GetControl("edit"):GetHeight() - 10)

finenv.RetainLuaState = true

dialog:ShowModeless()
```

### DIFF
--- a/src/mixin/FCMControl.lua
+++ b/src/mixin/FCMControl.lua
@@ -8,15 +8,28 @@ Summary of modifications:
 - In getters with an `FCString` parameter, the parameter is now optional and a Lua `string` is returned. 
 - Ported `GetParent` from PDK to allow the parent window to be accessed from a control.
 - Handlers for the `Command` event can now be set on a control.
+- Added methods for storing and restoring control state, which allows controls to correctly maintain their values across multiple script executions
 ]] --
 local mixin = require("library.mixin")
 local mixin_helper = require("library.mixin_helper")
 
 -- So as not to prevent the window (and by extension the controls) from being garbage collected in the normal way, use weak keys and values for storing the parent window
 local parent = setmetatable({}, {__mode = "kv"})
+local private = setmetatable({}, {__mode = "k"})
 local props = {}
 
 local temp_str = finale.FCString()
+
+--[[
+% Init
+
+**[Internal]**
+
+@ self (FCMControl)
+]]
+function props:Init()
+    private[self] = private[self] or {}
+end
 
 --[[
 % GetParent
@@ -53,10 +66,164 @@ function props:RegisterParent(window)
 end
 
 --[[
+% GetEnable
+
+**[Override]**
+Hooks into control state restoration.
+
+@ self (FCMControl)
+: (boolean)
+]]
+
+--[[
+% SetEnable
+
+**[Fluid] [Override]**
+Hooks into control state restoration.
+
+@ self (FCMControl)
+@ enable (boolean)
+]]
+
+--[[
+% GetVisible
+
+**[Override]**
+Hooks into control state restoration.
+
+@ self (FCMControl)
+: (boolean)
+]]
+
+--[[
+% SetVisible
+
+**[Fluid] [Override]**
+Hooks into control state restoration.
+
+@ self (FCMControl)
+@ visible (boolean)
+]]
+
+--[[
+% GetLeft
+
+**[Override]**
+Hooks into control state restoration.
+
+@ self (FCMControl)
+: (number)
+]]
+
+--[[
+% SetLeft
+
+**[Fluid] [Override]**
+Hooks into control state restoration.
+
+@ self (FCMControl)
+@ left (number)
+]]
+
+--[[
+% GetTop
+
+**[Override]**
+Hooks into control state restoration.
+
+@ self (FCMControl)
+: (number)
+]]
+
+--[[
+% SetTop
+
+**[Fluid] [Override]**
+Hooks into control state restoration.
+
+@ self (FCMControl)
+@ top (number)
+]]
+
+--[[
+% GetHeight
+
+**[Override]**
+Hooks into control state restoration.
+
+@ self (FCMControl)
+: (number)
+]]
+
+--[[
+% SetHeight
+
+**[Fluid] [Override]**
+Hooks into control state restoration.
+
+@ self (FCMControl)
+@ height (number)
+]]
+
+--[[
+% GetWidth
+
+**[Override]**
+Hooks into control state restoration.
+
+@ self (FCMControl)
+: (number)
+]]
+
+--[[
+% SetWidth
+
+**[Fluid] [Override]**
+Hooks into control state restoration.
+
+@ self (FCMControl)
+@ width (number)
+]]
+for method, valid_types in pairs({
+    Enable = {"boolean", "nil"},
+    Visible = {"boolean", "nil"},
+    Left = "number",
+    Top = "number",
+    Height = "number",
+    Width = "number",
+}) do
+    props["Get" .. method] = function(self)
+        if mixin.FCMControl.UseStoredState(self) then
+            return private[self][method]
+        end
+
+        return self["Get" .. method .. "_"](self)
+    end
+
+    props["Set" .. method] = function(self, value)
+        mixin.assert_argument(value, valid_types, 2)
+
+        if mixin.FCMControl.UseStoredState(self) then
+            private[self][method] = value
+        else
+            -- Fix bug with text box content being cleared on Mac when Enabled or Visible state is changed
+            if (method == "Enable" or method == "Visible") and finenv.UI():IsOnMac() and finenv.MajorVersion == 0 and finenv.MinorVersion < 63 then
+                self:GetText_(temp_str)
+                self:SetText_(temp_str)
+            end
+
+            self["Set" .. method .. "_"](self, value)
+        end
+    end
+end
+
+
+--[[
 % GetText
 
 **[Override]**
 Returns a Lua `string` and makes passing an `FCString` optional.
+Also hooks into control state restoration.
 
 @ self (FCMControl)
 @ [str] (FCString)
@@ -69,7 +236,11 @@ function props:GetText(str)
         str = temp_str
     end
 
-    self:GetText_(str)
+    if mixin.FCMControl.UseStoredState(self) then
+        str.LuaString = private[self].Text
+    else
+        self:GetText_(str)
+    end
 
     return str.LuaString
 end
@@ -79,6 +250,7 @@ end
 
 **[Fluid] [Override]**
 Accepts Lua `string` and `number` in addition to `FCString`.
+Also hooks into control state restoration.
 
 @ self (FCMControl)
 @ str (FCString|string|number)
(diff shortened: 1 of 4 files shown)


ThistleSifter:
Is that macOS's native behaviour? I don't have that issue on Windows and I haven't done any magic with the `Set*` methods.

Edit: when you say it loses the contents, do you mean on restore, or immediately while the dialog is running?

ThistleSifter:
Hmm, what happens if you try the following vanilla script?

```lua
local edit = dialog:CreateEdit(0, 0)
local btn = dialog:CreateButton(0, 40)

    if ctrl:GetControlID() == btn:GetControlID() then
        edit:SetEnable(not edit:GetEnable())
    end
end)

dialog:ExecuteModal(nil)
```

rpatters1:
So it turn out this *is* macOS behavior. The contents of an edit text box do not update internally until the control loses focus. On these test dialogs, by default there is no way to make the edit text lose focus because there is only one edit text box. Thus when enabled or visible state changes, the OS updates with the last known contents which, at least at first, is nothing. (You can test with your original test script by entering some data, closing the window, then reopening. At that point when you click the buttons, it retains the original contents from the last time it opened.)

Another way is to enable the tab-to-all-controls keyboard mode in macOS that allows you to tab to other controls besides edit text boxes, like Windows does. Then you can tab out of the edit text and it remembers what you typed.

There appears to be an option on edit text to update continuously. I am going to look into what it would take to make that available in RGP Lua.

More details at [this Stackoverflow article](https://stackoverflow.com/questions/2271908/nstextfield-setenabled).

rpatters1:
It looks like this solves the problem, even on existing Lua plugins:

```lua
    window:CreateButton(0, 40):SetWidth(75):SetText("Toggle Enable"):AddHandleCommand(function(self)
        local edit = self:GetParent():GetControl("edit")
        local str = finale.FCString()
        edit:GetText(str)
        edit:SetText(str)
        edit:SetEnable(not edit:GetEnable())
    end)
```

ThistleSifter:
Awesome, glad it got sorted out!

Is it worth adding some backwards-compatibility for `SetEnable` and `SetVisible` to `FCMControl`? Eg
```lua
if finenv.UI():IsOnMac() and finenv.MajorVersion == 0 and finenv.MinorVersion < 63 then
    self:GetText_(temp_str)
    self:GetText_(temp_str)
end
```

rpatters1:
I'm not sure how much it matters. But surely you meant:

```lua
if finenv.UI():IsOnMac() and finenv.MajorVersion == 0 and finenv.MinorVersion < 63 then
    self:GetText_(temp_str)
    self:SetText_(temp_str) -- <== *SetText
end
```

ThistleSifter:
Ah whoops, thanks for picking that up.

ThistleSifter:
Apologies for the hold up, life got in the way for a little while there. I'll resume working on it this week.

ThistleSifter:
@rpatters1 
A quick question about something that came up in testing: If an error is thrown in an attached `CloseWindow` handler, then storing control state won't work, since control state is stored after all `CloseWindow` handlers are executed. Two options come to mind: 1) Do nothing, leave it as is. 2) Execute handlers in a `pcall`, then store control state, and if there's an error, throw it afterwards.

I'm leaning towards number 2, but I guess the underlying question is, if there is an error, do we want control state stored? (I can't think of any reason why not, I just thought I'd ask).

rpatters1:
Option 2 seems like a good option to me as well. I guess it would ultimately depend on the nature of the error, but I can't think of one where I wouldn't prefer to store the state.

ThistleSifter:
@rpatters1 A couple more questions:

- Should I add an `FCMCtrlStatic` mixin to store/restore `SetTextColor`? In my testing, setting the color while the window is showing produced no visible change anyway (although it updates to the new color the next time the window is shown). It's only a small job, I'm just wondering whether it's useful.
- Timers only seem to work if they are `Set` after the window has been shown. This had me stumped for some time, so would it be possible to add this to the PDK Framework documentation?
- Command/Control events don't fire for `FCCtrlSlider` controls when clicking and dragging it. Is this able to be fixed in `RGPLua`?


Other than these I only have `FCMTree` left, and I'm still working out how best to tackle it. Would it be safe to make `FCTreeNode` objects persist across multiple executions? Or am I better off just storing the tree nodes as Lua strings in a table and recreating the whole thing every time the window is shown? If I do that, then any `FCTreeNode` objects that might be held by the script will be voided (perhaps they shouldn't be held anyway?).

rpatters1:
`SetTimer` has to be called while the Window is running. The earliest option is during `InitWindow`. I'll add that to the docs.

I think it would be nice if `SetTextColor` is retained the next time the window opens. If this requires `FCMCtrlStatic` to be created, then it probably should be.

If you have a script for `FCCtrlSlider` that isn't working, or is lacking features you'd like added, please send it to me and I'll look at it. I have not touched the code for `FCCtrlSlider`.

Same goes for `FCTree`. I don't know anything about it. My instinct would always be to retain the context in Lua strings/tables and recreate the underlying controls from scratch. So if there is any doubt, that would be the direction I'd choose.

ThistleSifter:
Okay, here are a couple of test scripts:


First one is testing slider (the text should update as the slider is moved):
```lua
local str = finale.FCString()
local static = dialog:CreateStatic(0, 0)

local slider = dialog:CreateSlider(0, 30)
slider:SetMinValue(0)
slider:SetMaxValue(10)
slider:SetThumbPosition(2)

str.LuaString = "Thumb position: " .. slider:GetThumbPosition()
static:SetText(str)

    str.LuaString = "Thumb position: " .. slider:GetThumbPosition()
    static:SetText(str)
end)

dialog:ExecuteModal(nil)
```
The text doesn't change at all for me on Windows.


Next one is static control text color:
```lua
local function create_dialog()
    local str = finale.FCString()
    str.LuaString = "Here is some text"
    local static = dialog:CreateStatic(0, 0)
    static:SetText(str)

    str.LuaString = "Red"
    local redbtn = dialog:CreateButton(0, 30)
    redbtn:SetText(str)
    redbtn:SetWidth(30)

    str.LuaString = "Green"
    local greenbtn = dialog:CreateButton(40, 30)
    greenbtn:SetText(str)
    greenbtn:SetWidth(30)

    dialog:RegisterHandleControlEvent(redbtn, function(ctrl)
        static:SetTextColor(255, 0, 0)
    end)

    dialog:RegisterHandleControlEvent(greenbtn, function(ctrl)
        static:SetTextColor(0, 255, 0)
    end)

    return dialog
end

window = window or create_dialog()
finenv.RetainLuaState = true
window:ExecuteModal(nil)
```
When I press either button, nothing happens. But if I close the window and run the script again, the text shows the color of the last button pressed.

rpatters1:
About `SetTextColor`, you may need to rewrite the text into the control for it to change. Here is a script I've been working on to monitor the current layer. I haven't completed it because it only works when the document is set to "show only current layer". But then it can change color without any issues.

```lua
-- These colors are the Finale default program settings for layer colors.
-- The PDK does not provide access to these values, but
-- you can override them with a configuration file.
config = {
    color_layer1 = {0, 0, 0},
    color_layer2 = {170, 31, 55},
    color_layer3 = {52, 112, 92},
    color_layer4 = {46, 85, 138}
}

function calc_current_layer_string()
    local misc_prefs = finale.FCMiscDocPrefs()
    local curr_doc = finale.FCDocument(-1) -- current document
    if curr_doc.ID > 0 and misc_prefs:Load(1) then
        if misc_prefs.ShowActiveLayerOnly then
            for layer_number = 1, finenv.UI():GetMaxLayers() do
                if finenv.UI():IsLayerVisible(layer_number) then
                    return tostring(layer_number)
                end
            end
        else
            return "All"
        end
    else
        return "N/A"
    end
    return "Err"
end

function calc_layer_color(layer_string)
    local retval = config["color_layer"..layer_string]
    if not retval then
        retval = config.color_layer1
    end
    return retval
end

global_timer_id = 1         -- per docs, we supply the timer id, starting at 1

function create_dialog_box()
    local dialog = mixin.FCXCustomLuaWindow():SetTitle("Layer")
    local font = finale.FCFontInfo()
    font.Name = "Helvetica"
    font.Size = 36
    font.Bold = true
    dialog:CreateStatic(0, 10, "layer_string")
                :SetText(calc_current_layer_string())
                :SetFont(font)
                :SetWidth(85)
                :SetHeight(45)
    dialog:CreateOkButton():SetText("Close")
    dialog:RegisterHandleTimer(function(self, timer_id)
            local static = self:GetControl("layer_string")
            local layer_text = calc_current_layer_string()
            local layer_color = calc_layer_color(layer_text)
            static:SetTextColor(layer_color[1], layer_color[2], layer_color[3]) -- table.unpack generates a warning, so don't use it
            static:SetText(layer_text)
        end
    )
    dialog:RegisterInitWindow(function(self)
            self.OkButtonCanClose = true -- override default behavior of RunModeless()
            self:SetTimer(global_timer_id, 100) -- timer can't be set until window is created
        end
    )
    dialog:RegisterCloseWindow(function(self)
            self:StopTimer(global_timer_id)
        end
    )
    return dialog
end

function layer_monitor()
    global_dialog = global_dialog or create_dialog_box()
    global_dialog:RunModeless(true) -- true: no selection required
    --global_dialog:StartTimer(global_timer_id, 100) -- timer can't be set until window is created
end

layer_monitor()
```

I'll look into the example for `FCCtrlSlider`.

Nick-Mazuk:
I'm noticing this is starting to become a very large PR—currently at almost 2,000 lines of code changed. That's going to make it very difficult to review, and potentially cause several merge conflicts.

Would it be possible to go ahead and review/merge this PR as is, then have a follow-up PR add the remaining items in the todo list?

In general, it's a really good idea to keep PRs decently small in both size and scope. Many small PRs are preferable to one large PR.

https://google.github.io/eng-practices/review/developer/small-cls.html

ThistleSifter:
Okay, I'll make the mixin version of `SetTextColor` set the text again under the hood.

@Nick-Mazuk 
Sure thing. I can split up this PR as well and resubmit parts of it later if you'd prefer?

Nick-Mazuk:
> @Nick-Mazuk Sure thing. I can split up this PR as well and resubmit parts of it later if you'd prefer?

That would be great. Thanks!

ThistleSifter:
I've marked this as ready for review. It just contains the base changes required for control state restoration to work, as well as a couple of indentation things.

rpatters1:
This message is just to confirm that `FCCtrlSlider` does not send change notification on either macOS or Windows. But they will be added in the next release of RGP Lua.

rpatters1:
What's the status of this pull request? Is it ready to be merged? Is it still compatible with later PRs?

ThistleSifter:
Yes it is ready to be merged.

The only possible issue that I can think of is the change in line 417 of `mixin/FCXCustomLuaWindow.lua` which I think was already fixed in one of your PRs. I'm not sure how git handles these things.

rpatters1:
Btw, @ThistleSifter, there is an enhancement in RGP Lua 0.64 that may affect this PR down the line. Until 0.64, the control instance passed back to the `HandleControlEvent` and `HandleCommand` handlers only had access to the base methods and properties of `FCControl`, even though `GetClassName` returned the proper subclass name. As of 0.64, the controls passed into those handlers have *all* the methods and properties pertaining to the subclass.

ThistleSifter:
@rpatters1 The mixin dialogs disregard the instance that's passed to the handler and instead always use the original object that was returned from the `Create*` method.

PR #214, opened 2022-06-14
Create layer_clear_selective.lua

cv-on-hub:
new script - clear chosen layer of current selection

### DIFF
--- a/src/layer_clear_selective.lua
+++ b/src/layer_clear_selective.lua
@@ -1,0 +1,70 @@
+function plugindef()
+    finaleplugin.RequireSelection = true
+    finaleplugin.Author = "Carl Vine"
+    finaleplugin.AuthorURL = "http://carlvine.com/?cv=lua"
+    finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
+    finaleplugin.Version = "v1.03"
+    finaleplugin.Date = "2022/06/09"
+    finaleplugin.CategoryTags = "Note"
+    finaleplugin.Notes = [[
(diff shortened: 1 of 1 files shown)


src/layer_clear_selective.lua
```diff
@@ -0,0 +1,70 @@
...
+    finaleplugin.Notes = [[
+        Clear all music from the chosen layer in the surrently selected region. 
+        (Note that all of a measure's layer will be cleared even if it is partially selected).
+    ]]
+    return "Clear layer selective", "Clear layer selective", "Clear the chosen layer"
+end
+
+-- RetainLuaState will return global variable: clear_layer_number
+
+function clear_chosen_layer(layer_number)
```
Nick-Mazuk, line 18: There's a library function `layers.clear`. Would that work here?

https://www.finalelua.com/docs/library/layer#clear

cv-on-hub: Yes of course. Silly me. Amended version follows. (This is a really old script that I've dragged up to fin-lua expectations).

cv-on-hub:
Is that last `Merge branch` the right thing to do? I just wanted my cv-on-hub fork to match the finale-lua "dist" contents.

Nick-Mazuk (approved):
LGTM

Nick-Mazuk:
> Is that last `Merge branch` the right thing to do? I just wanted my cv-on-hub fork to match the finale-lua "dist" contents.

It depends. That's what's called a merge commit (in case you want to Google it).

Some people say to never do it since it makes a messy commit history. Personally, I don't care since almost no one ever looks at the commit history, so I'm going to go ahead and approve and merge.

PR #215, opened 2022-06-16
amending 2 staff explode scripts

cv-on-hub:
this is a minor fix to both scripts which have been capable of splitting "pairs" with a minimum polyphony of 3 notes, but were previously restricted to work on minimum 4-voice polyphony.


Nick-Mazuk (approved):
LGTM

PR #222, opened 2022-06-24
Create delete_selective.lua

cv-on-hub:
This script replaces 11 independent "deletion" scripts that I've used for years. Not sure how useful it will be to others, depending on their workflow preferences.

### DIFF
--- a/src/delete_selective.lua
+++ b/src/delete_selective.lua
@@ -1,0 +1,121 @@
+function plugindef()
+    finaleplugin.RequireSelection = true
+    finaleplugin.Author = "Carl Vine"
+    finaleplugin.AuthorURL = "http://carlvine.com/lua"
+    finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
+    finaleplugin.Version = "0.51"
+    finaleplugin.Date = "2022/06/24"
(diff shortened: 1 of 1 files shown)


Nick-Mazuk (approved):
Yes please! I currently have a KM macro to do this through the Finale UI. But a script is so much faster. Thanks!

cv-on-hub:
I started making these scripts in direct response to your SuperUser KM macro-palette system. Not sure if I should include MIDI data (note and/or continuous).

rpatters1:
You should be using `eachbackwards` with articulations as well. Right now you are failing to delete one or more if there are multiple articulations. However, this is being masked by the fact that you are resetting the articulation flag.

rpatters1:
Some other issues:

- I was going to tell you you should be calling `DeleteEverything` (with loaded center shapes) instead of `DeleteData`, but Finale itself seems to be cleaning up for you (better than `DeleteEverything` would). This is weird magic to me, so I have asked my contact at MakeMusic to confirm it.
- You expression deletion code is leaving the `FCMeasure.ExpressionFlag` set, even if you delete all the expressions. My recommendation is to do a fresh call of `CreateExpressions` and reset it if the returned collection is empty.
- Your `measure_expression` code should use `eachbackwards`.

PR #223, opened 2022-06-25
Update configuration.lua

cv-on-hub:
Draft for adding a save_parameters() function to the configuration library file. No frills, but it works for me.

### DIFF
--- a/src/library/configuration.lua
+++ b/src/library/configuration.lua
@@ -127,4 +127,37 @@ function configuration.get_parameters(file_name, parameter_list)
     end
 end
 
+--[[
+% save_parameters
+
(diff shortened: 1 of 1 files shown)


Nick-Mazuk:
@cv-on-hub in #226 you asked me to review this one, but it looks like the PR is still draft. Were you wanting me to look it over?

cv-on-hub:
I thought Robert should review this since it adds to his script. Thought "Draft" was the appropriate setting, but it is ready for review so have upgraded to "Ready".

rpatters1 (approved):
It seems like this would be more flexible with simply:

```lua
        if type(v) == "string" then
            v = "\"" .. v .."\""
        else
            v = tostring(v)
 ```

Then it would easily support numbers as well. But I'm fine with it as-is.

PR #229, opened 2022-07-04
Update staff_explode.lua

cv-on-hub:
automatically create a new config file so the user doesn't have to

### DIFF
--- a/src/staff_explode.lua
+++ b/src/staff_explode.lua
@@ -26,6 +26,7 @@ local clef = require("library.clef")
 local config = {fix_note_spacing = true}
 
 configuration.get_parameters("staff_explode.config.txt", config)
+configuration.save_parameters("staff_explode.config.txt", config) -- create new file so the user doesn't have to
 
 function show_error(error_code)
     local errors = {


rpatters1:
This should either be put in the configuration library as part of `get_parameters` or else this should not be there at all. I think scripts should behave consistently with regard to configuration.

The advantage of this change is:

- The file gets created automatically making it easier for the user to find and edit it.

Disadvantage:

- Scripts will litter the user's hard drive with config files without the user's knowledge.

Personally, I am in the "teach 'em to fish" camp. If a user is picky enough to want other than default behavior, they should learn how to manipulate config files for themselves. However, I am willing to be persuaded, provided there is some effort required from users to learn better how to control the monster on their desk (or lap).

cv-on-hub:
Fair points, if a little strict, and this suggestion was only in response to a single complaint on the Facebook group. My counterargument is that it's sensible to save a new config file without the user's knowledge in a script wanting to retain user preferences, so is this materially different? 

My impression is that the vast majority of Finale users already find installing RGPLua sufficiently daunting to prevent them going to the trouble in the first place, and creating little text files in tricky locations is now such an unfamiliar task that it's an even greater discouragement.

rpatters1:
I'm not saying no to this. I just want more of our community to weigh in. Should all config files be created automatically?

I mean, as hard as I've worked on all this, I just don't have much patience for people who refuse to put in a little learning curve effort to save themselves hundreds of hours of miserable slogging.

cv-on-hub:
Horse. Water. Drink.

Nick-Mazuk:
If we can find a way to let users edit their configuration _without_ opening up the config files (e.g., create a new script to do that for them), I'm all for saving the config files.

However, if we're just writing the config files to then directly read from, I think that's less than ideal. The only advantage I can see with that is that then when a user finds the config file, they know what the configuration parameters are.

If they don't go looking for it there are a few downsides:

- We're creating files without any purpose
- Users might be confused why we're writing files to their computer
  - Some people may even mistake this for something malicious and start to distrust Lua
- I presume we're all using reasonable defaults for our configs. However, if we write all the config values down, we no longer are able to update those reasonable defaults as the script is modified since we have no idea if a _user_ updated the config file or if a previous version _of our script_ wrote the config file.

cv-on-hub:
Users are accustomed to many hundreds of prefs files created or updated in their home folder without their knowledge including, as it happens, several by Patterson and TGTools plugins! (Plus JW's config files in the Finale folder). It's true that creating prefs file to reflect active user choices is more justifiable/reasonable, but otherwise the only thing I find concerning is an old prefs file overwriting revised script behaviour. For which I'd rename the prefs file to sync with the new script. If the user changed the original prefs they can also change the new one.

rpatters1:
Based on new pull request by @jwink75, I think I see a way out of this. We should distinguish between configuration and user settings. Configuration will continue to reside in `script_settings` and only users will write to it. (Scripts never will.) User settings will reside in the user's settings directory and be fully writable all the time. Personally, I think @cv-on-hub should change his scripts that write settings to put the file in the user settings rather than the config directory.

That said, I think we need to have a naming convention for them, because writing random filenames in a preferences folder is very bad form. I would propose the following:

```
com.finalelua.<script-file-name>.config.txt
```

cv-on-hub:
That all sounds very sensible. To clarify, by "user's settings directory" do you mean what MacOS calls `~/Library/Preferences/`?

cv-on-hub:
Should I now close or delete this PR?

rpatters1:
I would just close it.

PR #230, opened 2022-07-07
Harp Pedal Wizard function

jwink75:
I'll add documentation sometime in the future!

### DIFF
--- a/src/harp_pedal_wizard.lua
+++ b/src/harp_pedal_wizard.lua
@@ -9,17 +9,17 @@ function plugindef()
     return "Harp Pedal Wizard", "Harp Pedal Wizard", "Creates Harp Diagrams and Pedal Changes"
 end
 
-  if finenv.IsRGPLua == false then
+if finenv.IsRGPLua == false then
     local ui = finenv.FCUI
     ui:AlertInfo("This script requires RGP Lua to function.", NULL)
     local str = finale.FCString()
     str.LuaString = "https://robertgpatterson.com/-fininfo/-rgplua/rgplua.html"
     ui:DisplayWebURL(str)
-  end
+end
 
 local library = require("library.general_library")
 --local configuration = require("library.configuration")
-  
+
 local config_filename = "com.harp_pedal_wizard.text"
 
 -- finalelua library functions - copied from library\configuration.lua
@@ -161,11 +161,12 @@ end
 ---
 function harp()
     finenv.RetainLuaState = true
-    local default_music_font = get_def_mus_font() 
     local partial = false
     local changes = false
     local stack = true
     local pedal_lanes = true
+    local direct = false
+    local override = false
     local context =
     {
         window_pos_x = window_pos_x or nil,
@@ -177,6 +178,7 @@ function harp()
     description = finale.FCString()
     local changes_str = finale.FCString()
     changes_str.LuaString = ""
+    local default_music_font = get_def_mus_font()
     local diagram_font = "^fontTxt("..default_music_font..")"
     --
     local flat_char = ""
@@ -186,6 +188,7 @@ function harp()
     local desc_prefix = finale.FCString()
     desc_prefix.LuaString = ""
     local ui = finenv.UI()
+    local direct_notes = {0, 0, 0, 0, 0, 0, 0}
 
     local config = harp_config_load()
 
@@ -203,6 +206,7 @@ function harp()
 
     function process_return(harpnotes)
         local error = false
+        direct_notes = {0, 0, 0, 0, 0, 0, 0}
         --
         local harp_tbl = split(harpnotes, ",")
         local count = 0 
@@ -234,18 +238,25 @@ function harp()
             ---- Assign to strings...
             if harp_tbl[i]:sub(1,1) == "A" then 
                 harpstrings[7] = harp_tbl[i]
+                direct_notes[7] = harp_tbl[i]
             elseif harp_tbl[i]:sub(1,1) == "B" then 
                 harpstrings[3] = harp_tbl[i]
+                direct_notes[3] = harp_tbl[i]
             elseif harp_tbl[i]:sub(1,1) == "C" then
                 harpstrings[2] = harp_tbl[i]
+                direct_notes[2] = harp_tbl[i]
             elseif harp_tbl[i]:sub(1,1) == "D" then
                 harpstrings[1] = harp_tbl[i]
+                direct_notes[1] = harp_tbl[i]
             elseif harp_tbl[i]:sub(1,1) == "E" then
                 harpstrings[4] = harp_tbl[i]
+                direct_notes[4] = harp_tbl[i]
             elseif harp_tbl[i]:sub(1,1) == "F" then
                 harpstrings[5] = harp_tbl[i]
+                direct_notes[5] = harp_tbl[i]
             elseif harp_tbl[i]:sub(1,1) == "G" then
                 harpstrings[6] = harp_tbl[i]
+                direct_notes[6] = harp_tbl[i]
             else
 --                error = true
                 --goto error1               
@@ -395,6 +406,19 @@ function harp()
         else
             new_pedals = harpstrings
         end
+
+        if (changes == false) and (partial == true) then
+            if direct == true then
+                local yesno = ui:AlertYesNo("There are no pedal changes required.\rAdd anyway?", nil)
+                if yesno == 3 then
+                    override = true
+                elseif yesno == 2 then
+                    new_pedals = direct_notes
+                    changes = true
+                end
+            end
+        end
+
         changes_update()
         for i = 1, 7, 1 do
             if diag == true then
@@ -471,7 +495,6 @@ function harp()
             elseif (pedal_lanes == false and partial and diagram_string.LuaString == "") then
                 diagram_string.LuaString = left_strings.LuaString
             end
-
             diagram_string.LuaString = string.gsub(diagram_string.LuaString, "n", "^natural()")
             diagram_string.LuaString = string.gsub(diagram_string.LuaString, "b", "^flat()")
             diagram_string.LuaString = string.gsub(diagram_string.LuaString, "#", "^sharp()")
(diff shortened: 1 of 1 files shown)


src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1591 @@
...
+    -- are both reserved for the plug-in definition.
+    finaleplugin.RequireSelection = true
+    finaleplugin.Author = "Jacob Winkler"
+    finaleplugin.Copyright = "2022"
+    finaleplugin.Version = "2.0"
+    finaleplugin.Date = "2022-07-06"
+    return "Harp Pedal Wizard", "Harp Pedal Wizard", "Creates Harp Diagrams and Pedal Changes"
+end
+
+if finenv.IsRGPLua == false then
```
rpatters1, line 12: This is a perfect example why you should never explicitly compare boolean values to `true` or (especially) `false`. This code does not do what you expect it to in JW Lua. That's because in JW Lua, `IsRGPLua` is `nil`, not `false`. So in JW Lua this expression will evaluate to false (that is `IsRGPLua` ~= `false`.)

Replace this with `if not finenv.IsRGPLua then`.

src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1591 @@
...
+
+local config_filename = "com.harp_pedal_wizard.text"
+
+-- finalelua library functions - copied from library\configuration.lua
+local strip_leading_trailing_whitespace = function (str)
+    return str:match("^%s*(.-)%s*$") -- lua pattern magic taken from the Internet
+end
+
+local parse_parameter -- forward function declaration
+
```
rpatters1, line 31: As as a matter of interest, Lua does not require a forward function declaration. The only thing that is required is that `parse_parameter` be defined as a function by the first time you try to call it in `parse_table`. This "forward declaration" actually does nothing.

src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1591 @@
...
+            local name = strip_leading_trailing_whitespace(string.sub(line, 1, delimiter_at-1))
+            local val_string = strip_leading_trailing_whitespace(string.sub(line, delimiter_at+1))
+            parameters[name] = parse_parameter(val_string)
+        end
+    end
+    return parameters
+end
+
+-- Modified from library
+function get_parameters(file_name, parameter_list)
```
rpatters1, line 78: Why are you replicating the library function? Instead, if the library function doesn't do exactly what you need, enhance it to do that. Don't duplicate the code.

src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1591 @@
...
+    local path = finale.FCString()
+    local path_delimiter = finale.FCString()
+    local ui = finenv.UI()
+    path:SetUserOptionsPath()
+    if ui:IsOnMac() then
+        path_delimiter.LuaString = "/"
+    elseif ui:IsOnWindows() then
+        path_delimiter.LuaString = "\\"
+        --path_delimiter.LuaString = "/" -- apparently Windows can use either! Go figure!
+    end
```
rpatters1, line 102: Window recognizes "/" as a path delimiter. (This is a lesser-known fact.) But if you want to explicitly use "\\" instead, at least remove the comment.

src/harp_pedal_wizard.lua
rpatters1: It appears the only reason you've duplicated the library code is to be able to use a different file path. Instead of duplicating the code, add a function to the config library called something like `get_parameters_with_path`. Actually, the most correct steps would be:

- Change `get_parameters` to `get_parameters_with_full_path`
- Add a new `get_parameters` that calls `get_parameters_with_full_path` with the expected `script_settings` path.

This expands the library for everyone instead of being a one-off.

src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1591 @@
...
+    for i,k in pairs(init_settings) do
+        init_count = init_count + 1
+    end
+    --
+    local file_r = io.open(path.LuaString, "r")
+
+    if file_r == nil then
+        config_settings = init_settings
+        harp_config_save(config_settings)
+        file_r = io.open(path.LuaString, "r")
```
rpatters1, line 122: similar comment for saving as for getting. This should be in the library.

src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1591 @@
...
+    local fontinfo = finale.FCFontInfo()
+    if fontinfo:LoadFontPrefs(finale.FONTPREF_MUSIC) then
+        return getUsedFontName(fontinfo:GetName())
+    end
+end
+
+---
+function harp()
+    finenv.RetainLuaState = true
+    local partial = false
```
rpatters1, line 164: Recommendation: do not set `finenv.RetainLuaState = true` until after you successfully run `ShowModeless`. That's because if an error occurs while building the the dialog box, it leaves your Lua state in an unknown condition and the only way to recover is restart Finale.

Ultimately, this code will be greatly simplified by using the `RunModeless` mixin function that handles all of these details for you. I really wish we could start merging code that uses it.

src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1591 @@
...
+    local changes_str = finale.FCString()
+    changes_str.LuaString = ""
+    local default_music_font = get_def_mus_font()
+    local diagram_font = "^fontTxt("..default_music_font..")"
+    --
+    local flat_char = ""
+    local nat_char = ""
+    local sharp_char = ""
+    local cross_char = ""
+    local desc_prefix = finale.FCString()
```
rpatters1, line 188: When RGP Lua 0.63 comes out (soon I think), this needs to be changed to, e.g.:

```lua
local flat_char = utf8.char(0xe360) -- standard SMuFL flat
```
Embedding SMuFL characters directly in the text file is risky because text fonts used by most editor will not be able to display them, so you can't tell from the code what the character is.

src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1591 @@
...
+                    new_pedals[i] = harpstrings[i]
+                    changes = true
+                end
+            end
+        else
+            new_pedals = harpstrings
+        end
+
+        if (changes == false) and (partial == true) then
+            if direct == true then
```
rpatters1, line 411: Again with the direct comparisons to boolean values. This is a habit I hope you will break.

```lua
if not changes or partial then
    if direct then
```
is cleaner

rpatters1: The only times you might compare explicitly are

- to distinguish between `false` and `nil`. In this case I would compare to `nil`: `if x == nil then`
- to distinguish between `true` and some other non-nil non-boolean type. In this case, I would rewrite my code not to do this or put an assert on the type to guarantee it is a boolean type.

src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1591 @@
...
+                dialog:ShowModeless()
+
+--                if dialog:ExecuteModal(nil) == finale.EXECMODAL_OK then
+--                end -- if dialog:Execute...
+
+            end -- get_harp
+            harp_dialog()
+        end -- harp
+
+        harp()
```
rpatters1, line 1591: I'm being a bit picking on this review because I really like this script. It's a great idea for a script, and I want to it to be as robust as possible. When RGP Lua 0.63 comes out I'm probably going to enhance it to use the `utf8` library for SMuFL handling (unless you want to). Also, once we can deploy mixins correctly, this definitely needs to transition to `RunModeless`.

src/harp_pedal_wizard.lua
rpatters1: As a followup, I am going to take a stab at enhancing the configuration library for user settings. So look for a pull request from me on this.

jwink75:
I will work on cleaning this up, thanks Robert! I do want to point out that the ‘forward function declaration’ bit was another thing I copied/pasted from the library configuration.lua file (line 60), and it’s still in there.

I did try switching the way I do booleans and was still having trouble, so I figured there would be no downside to being as explicit as possible… I will work on breaking myself of that habit, I promise :)

-Jake

jwink75:
I went through and as carefully as I could changed all the explicit comparisons to true, false, or nil to ‘if’ and ‘if not’ type statements. I was as exacting as I could be, but it still comes out broken at the other end, so I’m not sure what is going wrong. 

Also: I’m trying to use the existing configuration functions, and they’re just not working. I don’t know if I’m loading them wrong or what… I’m using the exact same approach as the general library (local configuration = require(“library.configuration”), but I get an error message. That’s why I copied the relevant functions into the main script in the first place as well.

Given that my next inclination is to toss my computer out the window, I think I need to take a step back from this for a bit!

rpatters1:
If you want to provide a specific example that was working before and no longer works now, I can probably help you figure it out. Please post before and after snippets.

src/harp_pedal_wizard.lua
jwink75: Removed! I used the minimum version as you suggested elsewhere

src/harp_pedal_wizard.lua
jwink75: Added the utf 8 codes but commented out, waiting for 0.63

src/harp_pedal_wizard.lua
jwink75: I cleaned this all up, but had to leave the one in line 598

rpatters1:
Now that RGP Lua 0.63 is released and the `configuration.lua` changes are merged, let's change the SMuFL character definitions to use `utf8.char`. And remove the duplicated configuration code in favor of `get/set_user_settings`. If there is something missing, let me know.

One thing I realized is that it might be handy to have a library to calculate the script name to pass in. It is possible to calculate it from `finenv.RunningLuaFilePath()`. Is this something you would find useful?

jwink75:
Yes, I’ve already changed to utf8.char internally. I’ll play around with the configuration stuff more on my plane ride home to try to get that working.

I do think it would be handy to have a library function to calculate the script name, yes!

PR #231, opened 2022-07-07
add user_settings to configuration library

rpatters1:
This function modifies `configuration.lua` to add the concept of _user settings_. User settings are written by the scripts themselves and reside in the user's preferences folder in an appropriately-named location. (The naming convention is a detail that the configuration library handles for the caller.) If the user settings are to be changed from their defaults, the script itself should provide a means to change them.

User settings are in contrast to _configuration_ settings, which only the user may modify and then only if they wish to.

The changes are as follows:

- added `configuration.get_user_settings` function
- added `configuration.save_user_settings` function
- removed `configuration.save_parameters` function

@cv-on-hub @jwink75

### DIFF
--- a/src/library/configuration.lua
+++ b/src/library/configuration.lua
@@ -81,11 +81,11 @@ parse_parameter = function(val_string)
     return tonumber(val_string)
 end
 
-local get_parameters_from_file = function(file_path)
-    local parameters = {}
+local get_parameters_from_file = function(file_path, parameter_list)
+    local file_parameters = {}
 
     if not file_exists(file_path) then
-        return parameters, false
+        return false
     end
 
     for line in io.lines(file_path) do
@@ -97,11 +97,18 @@ local get_parameters_from_file = function(file_path)
         if nil ~= delimiter_at then
             local name = strip_leading_trailing_whitespace(string.sub(line, 1, delimiter_at - 1))
             local val_string = strip_leading_trailing_whitespace(string.sub(line, delimiter_at + 1))
-            parameters[name] = parse_parameter(val_string)
+            file_parameters[name] = parse_parameter(val_string)
         end
     end
 
-    return parameters, true
+    for param_name, _ in pairs(parameter_list) do
+        local param_val = file_parameters[param_name]
+        if nil ~= param_val then
+            parameter_list[param_name] = param_val
+        end
+    end
+
+    return true
 end
 
 --[[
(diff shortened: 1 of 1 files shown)


rpatters1:
Also, once PR #231 is merged, `save_parameters` will go away and be replaced by `save_user_settings`. This will save anything you pass to it, overwriting what was there before. So you can use `get_user_settings` to merge an existing file into your table of parameters in memory. Then you can save your table of parameters in memory back out, and it will overwrite what was there before, replacing any discontinued fields and adding any added fields.

It seems to me like you are overthinking this. Just trust the get/save to do it, and it will probably work for you. If it doesn't we'll fix it so it does.

cv-on-hub:
This looks great.

jwink75:
Hey folks, I’m still struggling to understand how the functions in the configuration file are supposed to work. 

The one that I am hung up on in at the moment is get_parameters().

As I see it, this function is fed a filename and a set of parameters (parameter_list). Then the function calls up a file with get_parameters_from_file(), compares the values and updates the original set of parameters. That’s great, but then all it returns is is a boolean on whether it is successful. I don’t see the point if it’s not returning the updated parameter_list itself (which is the way I modified it in my harp pedal script). What is the boolean supposed to be used for?

I’m tempted to rewrite this function so that it
1) can take a third input parameter to override the path it calculates for itself,
2) outputs the boolean AND the parameter list.

If I were to do that, am I correct in understanding that as long as I keep the boolean as the FIRST variable returned, I won’t be breaking anybody else’s scripts that use this script?

rpatters1:
It modifies the parameter list in place. The return value just tells you whether it found the file. Please do not add a second return value.

I prefer you not add a parameter to specify a path. The last thing we need is for config files to be scattered all over the place. They should be in a consistent location.

Personally, at least with respect to the harp pedaling script, I think you should be looking at `get/set_user_settings`. The way the parameter list works is identical to `get_parameters` but the script can save new values with user settings.

jwink75:
What is the purpose of modifying the parameter list in place?

rpatters1:
P.S. You can probably ignore the return value. It usually doesn't matter on a get function. It's there mainly to allow an existing script to transition from using config parameters to user settings.

rpatters1:
P.P.S. I have rewritten the documentation to show before and after examples. Perhaps that will help.

PR #234, opened 2022-07-09
Update staff_explode.lua

cv-on-hub:
Fixing staff_explode to input only user-created config parameters.

### DIFF
--- a/src/staff_explode.lua
+++ b/src/staff_explode.lua
@@ -3,20 +3,21 @@ function plugindef()
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "http://carlvine.com"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "v1.46"
-    finaleplugin.Date = "2022/07/09"
+    finaleplugin.Version = "v1.47"
+    finaleplugin.Date = "2022/07/11"
     finaleplugin.Notes = [[
-        This script explodes a set of chords from layer one on one staff onto single lines on subsequent staves. 
+        This script explodes a set of chords on one staff into single lines on subsequent staves. 
         The number of staves is determined by the largest number of notes in any chord.
         It warns if pre-existing music in the destination will be erased. 
         It duplicates all markings from the original and resets the current clef on each destination staff.
 
-        By default this script respaces the selected music after running. 
-        To change this behaviour you must create a `configuration` file. 
+        This script doesn't respace the selected music after it completes. 
+        If you want automatic respacing you must create a `configuration` file. 
         If it does not exist, create a subfolder called `script_settings` in the folder containing this script. 
-        In that folder create a text file  called `staff_explode.config.txt` containing the line:  
+        In that folder create a plain text file  called `staff_explode.config.txt` containing the line:  
         ```
-        fix_note_spacing = false
+        fix_note_spacing = true -- respace music when the script finishes
+
         ```
     ]]
     return "Staff Explode", "Staff Explode", "Staff Explode onto consecutive single staves"
@@ -25,7 +26,7 @@ end
 local configuration = require("library.configuration")
 local clef = require("library.clef")
 
-local config = {fix_note_spacing = true}
+local config = { fix_note_spacing = false }
 configuration.get_parameters("staff_explode.config.txt", config)
 
 function show_error(error_code)
--- a/src/staff_explode.lua
+++ b/src/staff_explode.lua
@@ -1,32 +1,30 @@
 function plugindef()
     finaleplugin.RequireSelection = true
     finaleplugin.Author = "Carl Vine"
-    finaleplugin.AuthorURL = "http://carlvine.com"
+    finaleplugin.AuthorURL = "http://carlvine.com/?cv=lua"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "v1.46"
-    finaleplugin.Date = "2022/07/09"
+    finaleplugin.Version = "v1.47"
+    finaleplugin.Date = "2022/07/11"
+    finaleplugin.AdditionalMenuOptions = [[ Staff Explode (spaced) ]]
+    finaleplugin.AdditionalUndoText = [[    Staff Explode (spaced) ]]
+    finaleplugin.AdditionalPrefixes = [[    music_spacing = true ]]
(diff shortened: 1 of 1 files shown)


src/staff_explode.lua
```diff
@@ -3,18 +3,20 @@ function plugindef()
...
-        This script allows for the following configuration:
-
+        By default this script respaces the selected music after running. 
+        To change this behaviour you must create a `configuration` file. 
+        If it does not exist, create a subfolder called `script_settings` in the folder containing this script. 
+        In that folder create a text file  called `staff_explode.config.txt` containing the line:  
         ```
-        fix_note_spacing = true -- to respace music automatically when the script finishes
+        fix_note_spacing = false
         ```
```
rpatters1, line 20: I would add back the comment here:

```
fix_note_spacing = false -- to skip respacing music when the script finishes
```

Also, add a blank line between the last line of text and the three tick marks. Sometimes the documentation generator gets confused if there isn't a blank line.

rpatters1 (changes requested):

src/staff_explode.lua
Nick-Mazuk: The lack of a blank line isn't causing an issue here. However, it's generally good practice to prevent ambiguous markdown.

cv-on-hub: A neater and more useful variation - no user settings or config files. Just a simple second menu option. If this is approved I'll duplicate the process with my other 3 EXPLODE scripts.

rpatters1:
I have concerns about proliferating menu options to accommodate preferences. The permutations could quickly become unmanageable. FWIW: I load the entire `./src` into RGP Lua, and the number of menu options has exploded already.

Personally, I don't think spacing should even be an option. Finale itself has an automatic spacing option that may already take care of this. (Have you tested it?) Unfortunately, plugins cannot read this value, much as I have wanted to be able to through the years. `FCGeneralPrefs` seems never to have been expanded since about Finale 2000.

In my opinion, there are two camps of Finale users regarding automatic spacing: people who want it all the time and people who never want it. (I'm in the latter camp.) Therefore, I think it should remain an config parameter.

Nick-Mazuk:
+1 to Robert's points. I currently have over 250 items in the menu bar. Any more becomes unwieldy—especially if it's for a preferences script that I'll only ever use once (if that).

I'd be open to having one script that can update all the preferences for all scripts, I'm not a fan of a different menu item for each script.

cv-on-hub:
All comments noted. (Though to be fair, 23 of your menu items arise from your "4/4 meter" script!) Things like Layer Hide/Unhide used to be individual scripts - complementary function, so I'd maintain still useful as separate menu items. 

Anyway, my last commit reverts to simple User .config file, with default behaviour "no music spacing".

Nick-Mazuk:
> (Though to be fair, 23 of your menu items arise from your "4/4 meter" script!)

Yeah… it works well when using Keyboard Maestro. However, using a script that I don't have a macro for is substantially slower.

rpatters1:
> (Though to be fair, 23 of your menu items arise from your "4/4 meter" script!)

This gets to the question of what should be a menu and what should be a config option and what should be a user setting. After thinking about this a second, here is a first draft of some ideas.

## Menu Items

Menu items:

- Are functions that a user could reasonably be expected to use on a regular or at least semi-regular basis.
- Are functions the should be easily accessible to keyboard maestro and the like.
- Do not change.

The 23 menu items in "4/4" seem to be clearly in this category (though I blanched at the number of items when Nick originally proposed them.) Even if I personally don't use every one of them, I can imagine using them and I have already used more of them than I could list. Also, they correspond to a menu in Finale, so they are literally making a Finale menu available to keyboard maestro that was not available to it before.

## Configuration Items

Configuration items:

- Embody behavior that reflects the preferences of the user.
- Rarely be expected to change for a given user.
- Are set or changed only by users comfortable with editing text files (at least currently).

I see the spacing option in this PR as in that category (as I commented above). Most users either want or do not want automatic spacing, and that preference almost never changes. If Finale provided access to the automatic spacing program option, we could dispense with the script's configuration option altogether and base the script's behavior on Finale's option. But alas, that road is closed to us.

## User settings

User settings:

- Allow scripts to persist data across calls, such as the last data entered into a dialog box.
- May change frequently.
- Are written and read by the script itself.
- Can replace configuration items.

cv-on-hub:
Having an agreed set of criteria will save a lot of discussion time down the track, and this is an excellent set of orthogonal formulae to start with. 

A few suggestions ...
**Menu Items**:
1) "at least" is redundant
2) "that should" not "the should"

**Configuration Items:**
2) Are expected to remain consistent for each user

**User Settings:**
4) .. if there is a clear distinction between _configurations_ and _settings_ then I don't think settings should be able to "replace" configurations.

rpatters1:
I addressed the typos in the original comment.

I see user settings as being an "upgrade" or superset of configuration options. If a script has user settings, it has no need for configuration options. Basically, configuration options are the legacy and user settings are the future. But it's a little more complicated than that because a script may want to have some configuration options without going to the trouble of creating a user interface to edit them.

Here is a case in point: all the various nudger scripts. Currently the nudge amount is either a configuration option or a hard-coded value that could become a configuration option. It has occurred to me that I could add an optional dialog to them, accessed by invoking with Option/Shift (see #246), to allow the user to edit the nudge amount. This converts a configuration option to a user setting.

cv-on-hub:
Not to be picky, but aren't the most salient element of UserSettings that:
1) they are chosen by the user within the script
2) the same user is likely to make different choices depending on the context (unlike configs)

Furthermore that they should (could?) be realised using RetainLuaState rather than saved to file unless there are many parameters, or it's important to retain choices between Finale restarts?

rpatters1:
From my perspective, the most salient point of user settings is that the script must provide a UI (or other means if you can think of one) to modify them. Once you are providing a UI, does it really matter how often one of the fields changes?

RetainLuaState is fine as far as it goes, but retaining user choices across Finale calls is even better. I notice that in your other scripts you dropped RetainLuaState in favor of user settings. That's an inclination I support. But it's also good to have a variety of tools at our disposal. RetainLuaState is still an excellent tool for saving the result of an expensive calculation. (For example, the class index the the RGP Lua Class Browser creates.) But with user settings it is perhaps less useful than user settings for persisting user choices (including modeless window position).

cv-on-hub:
PS: just noting that my final commit d833a65 (not yet merged) addressed your initial concerns.

rpatters1:
I just had one tiny comment left: the blank line in the Notes should be before the first triple-tick mark not before the 2nd triple-tick mark (as it is now).

cv-on-hub:
Of course. Sorry - I misunderstood.

rpatters1 (approved):

PR #235, opened 2022-07-09
Update cue_notes_create.lua

cv-on-hub:
Revision to take advantage of the new configuration.save_user_settings() function, and removes access to user parameter file. Also revised Notes.


rpatters1 (changes requested):
The same comments as for cross_staff_offset.lua.

rpatters1 (approved):

PR #236, opened 2022-07-09
Update cross_staff_offset.lua

cv-on-hub:
Revision taking advantage of the new configuration.save_user_settings() function, and replacing the option-key detection of "dialog-free" operation with a secondary menu that is infinitely more useful to my workflow.

### DIFF
--- a/src/cross_staff_offset.lua
+++ b/src/cross_staff_offset.lua
@@ -2,26 +2,43 @@ function plugindef()
     finaleplugin.RequireSelection = true
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
     finaleplugin.Author = "Carl Vine"
-    finaleplugin.AuthorURL = "http://carlvine.com"
-    finaleplugin.Version = "v1.21"
-    finaleplugin.Date = "2022/05/23"   
+    finaleplugin.AuthorURL = "http://carlvine.com/?cv=lua"
+    finaleplugin.Version = "v1.26"
+    finaleplugin.Date = "2022/07/09"
+    finaleplugin.AdditionalMenuOptions = [[  CrossStaff Offset No Dialog  ]]
+    finaleplugin.AdditionalUndoText = [[     CrossStaff Offset No Dialog  ]]
+    finaleplugin.AdditionalPrefixes = [[     cross_staff_nodialog = true  ]]
+    finaleplugin.AdditionalDescriptions = [[ Offset horizontal position of cross-staff note entries - NO DIALOG ]]
     finaleplugin.Notes = [[
-        When creating cross-staff notes using the option-downarrow shortcut, the stems of 'crossed' notes
-        are reversed (on the wrong side of the notehead) and so appear too far to the right (if shifting downwards)
-        by the width of one notehead (typically 24EVPU). This script lets you set a horizontal offset for all cross-staff
-        notes in the currently selected region, with a different offset for non-crossed notes. For crossing to the
-        staff below use (-24,0) or (-12,12). For crossing to the staff above use (24,0) or (12,-12). Also specify
-        which layer number (1-4) or "all layers" (0). (This also offers a simple way to reset the horizontal offset
-        of all notes in the selection to zero).
-
-        If you invoke the plugin with Option key (macOS) or Shift key pressed, the script uses the last settings
-        from the dialog.
-    ]]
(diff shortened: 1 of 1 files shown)


src/cross_staff_offset.lua
```diff
@@ -2,26 +2,43 @@ function plugindef()
...
+   return "CrossStaff Offset", "CrossStaff Offset", "Offset horizontal position of cross-staff note entries"
 end
- 
+
+local config = {
+    cross_staff_offset  = 0,
+    non_cross_offset    = 0,
+    layer_number        = 0,
+    script_name         = "cross_staff_offset"
+}
```
rpatters1, line 34: Is it wise to include the script name in the configuration? It could potentially be changed by the configuration library, and I don't see the purpose of including it in the configuration. (FWIW: when the settings file is written, there is a comment at the top stating which script_name was passed in, and of course the filename itself contains the string as well.)

I just realized it would be possible to create a library function to return this by parsing it out of `finenv.RunningLuaFilePath()`. Is that something you would use instead?

src/cross_staff_offset.lua
```diff
@@ -73,34 +94,28 @@ function is_out_of_range(horiz_offset)
...
-        -- use global result values so that RetainLuaState remembers them
-        dialog_result, cross_offset, non_cross_offset, layer_number = user_selects_offset()
-        if dialog_result ~= finale.EXECMODAL_OK then -- user cancelled
+    if not cross_staff_nodialog then
+        if not user_selects_offset() then -- user cancelled
             return
-        end
-        if nil ~= finenv.RetainLuaState then
-            finenv.RetainLuaState = true
+        else -- check new selections
```
rpatters1, line 100: One of the things that I have been using `RetainLuaState` for is to remember where the dialog box is and move it there before opening it. (You can see an example of how to do this in the current master branch of transpose_by_step.lua.) Now that we have user settings, you could do the same with user settings. I find that the UX is greatly improved when a dialog box opens at the same location where you left it. (Once all the mixin stuff is checked in, this will be handled automatically by mixin code for modeless dialogs.)

src/cross_staff_offset.lua
```diff
@@ -2,26 +2,43 @@ function plugindef()
...
+        For crossing to the staff below use (-24,0) or (-12,12).  
+        For crossing to the staff above use (24,0) or (12,-12).  
+        Also specify which layer number to act upon (1-4) or "all layers" (0). 
+        (This also offers a simple way to reset the horizontal offset of all notes in the selection to zero).
+    
+        Under RGPLua (version 0.62+) the script adds an extra menu item that repeats the last 
+        chosen offset without presenting a dialog, for fast changes duplicating the last settings. 
+]]
+   return "CrossStaff Offset", "CrossStaff Offset", "Offset horizontal position of cross-staff note entries"
 end
```
rpatters1, line 27: The menu option that opens a dialog should include "..." at the end. (But only the menu string, not the undo string.)

rpatters1 (changes requested):

src/cross_staff_offset.lua
cv-on-hub: Thanks as always for your great advice Robert. All taken on board, and saving window position in user settings will be excellent. But I'm having trouble with dialog:RestorePosition() in a modal dialog. I'll submit the current script revision as work-in-progress, with a dummy `test_txt` variable to show behaviour. Position not saving.

(PS: `transpose_by_step` is my favourite transposition tool).

cv-on-hub: 2nd re-try to follow, using modeless dialog, is more promising, but the sequencing isn't right yet.

rpatters1:
If repositioning the modal inside the InitWindow handler didn't work, it may not work for modal dlgs. Tbh, I don't recall if I tested it.

cv-on-hub:
Got there at last with Modeless. Dialog window position saved in `user_settings`. Don't know why but setting these to `nil` for the initial default fails, and (as Robert already knew) none of it works without a local UndoBlock.

rpatters1 (approved):

rpatters1:
FWIW: Dialog positioning by the Lua script is not dependent on the local Undo block. It is dependent on the dialog being modeless. At least on macOS, the Framework always centers modal dialogs, overriding the scripts positioning settings.

cv-on-hub:
Good to know. I couldn't get the script to behave properly as modeless until properly implementing the Undo Block.

PR #238, opened 2022-07-09
adds general_library.calc_script_name

rpatters1:
This PR adds a function to calculate the script name for passing to `configuration.get_user_settings` (and the setter). I thought about building it directly into the configuration library, but I think could be a useful stand-alone function, and I didn't want to add a library dependency in `configuration.lua`.

### DIFF
--- a/src/library/general_library.lua
+++ b/src/library/general_library.lua
@@ -442,4 +442,36 @@ function library.system_indent_set_to_prefs(system, page_format_prefs)
     return system:Save()
 end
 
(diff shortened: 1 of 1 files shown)


rpatters1:
@jwink75 @cv-on-hub  may be interested in this as well

cv-on-hub:
I think I'll find that useful.

Nick-Mazuk (approved):
Looks good to me.

PR #239, opened 2022-07-09
add general_library.get_smufl_font_list function

rpatters1:
This is a function that I suspect @jwink75 is going to find to be very useful.


Nick-Mazuk (approved):
Looks good to me

PR #240, opened 2022-07-10
Enable bundling of mixin scripts

Nick-Mazuk:
This PR enables the bundling of the mixin scripts. To do so, it replaces the custom bundler I wrote with https://github.com/Benjamin-Dobell/luabundle. Additionally, since it results in larger file sizes in general, I minified the resulting bundles using https://github.com/mathiasbynens/luamin.

Some things to note:

- The majority of the bundling now depends on 3rd-party dependencies, so if there are bugs we'd need to fix them in the upstream repos
- The mixins (e.g., things located in `src/mixin`) are not included unless the mixin library is required (e.g., `require("library.mixin")`. The exact heuristic is that if the text `library.mixin` is used anywhere in the file, all the mixins will be bundled.
  - This means we don't need to statically determine which mixins are used, which is a much more complex issue
  - It also means that anytime you use a mixin, you must put `library.mixin` somewhere in the file (usually the require statement is sufficient).
- Imported lua files are no longer required to be located in `src/library`. They can now be located in any subfoler inside the `src` folder (including folders added in the future).
- I cannot entirely test things out locally, so it's possible there will be build errors when this PR is merged

Overall, I think this is a net win.

Allows the merging of #183.

Feel free to merge this PR when approved.


rpatters1 (approved):
I'm not really qualified to review this. If you are comfortable with merging it, go for it. We can always address any issues with another PR.

rpatters1:
One reason I'm punting on mergin is the merge conflicts with the dist files. I'm wondering if they should even be tracked.

Nick-Mazuk:
> One reason I'm punting on merging is the merge conflicts with the dist files. I'm wondering if they should even be tracked.

Right now they should be tracked.

When someone clicks "Download" on the website, they download the bundled file directly from GitHub. For instance, if someone wants to download the "simplify accidentals" script, they are essentially doing the following:

```
<a download href="https://raw.githubusercontent.com/finale-lua/lua-scripts/master/dist/accidental_simplify.lua">Download</a>
```

It's slightly more complex than that to meet more edge cases, but that's the idea.

I'm open to not checking in these bundled files. It just means we'll need to figure out another place to store the bundled scripts.

rpatters1:
I figured there might be an issue that they couldn't be on github if not tracked. I see removing them as very low priority.

PR #243, opened 2022-07-11
update system nudgers to use SpaceAbove

### LINKED ISSUE #87: [BUG] System Up/Down Nudgers always use top system margin
I would like to modify the system up/down nudgers to use Distance Between instead of Top Margin. In general, a non-zero system margin offers much less flexibility with page layout than leaving the vertical system margins at zero.

If others feel that Top Margin is preferable, we could implement it so as to use the margin when it is non-zero or the distance between when it is zero or when the result would be negative.

What do y'all think? (It's a simple change.)

rpatters1:
and consolidate to a single script.

Resolves #87


Nick-Mazuk (approved):
Looks good to me

PR #247, opened 2022-07-11
File comparison script

rpatters1:
This script implements a way to easily find substantive changes between two Finale documents. It emits all the notational elements into a text file that can be used with a text-file comparison tool to find differences. (More details in the Notes section of the script.)

I see this as being useful for composers to track changes they've made as well as ultimately by musicologists who want to study those changes later. The format of the text file itself is not so important as the fact that it can point you where to look in the scores to identify the changes.

I have requested a review by @ThistleSifter to see if I have fully utilized mixins to the extent that they can be used without adding a new `FCM` or `FCX` module. I am particularly interested in `FCMFileSaveAsDialog` but if you see any other easy opportunities, by all means alert me.


Nick-Mazuk (approved):
LGTM

src/document_save_as_text.lua
```diff
@@ -0,0 +1,377 @@
...
+end
+
+function do_save_as_dialog(document)
+    local path_name = finale.FCString()
+    local file_name = finale.FCString()
+    local file_path = finale.FCString()
+    document:GetPath(file_path)
+    file_path:SplitToPathAndFile(path_name, file_name)
+    local full_file_name = file_name.LuaString
+    local extension = finale.FCString()
```
ThistleSifter, line 58: Lines 58-60 could be shortened to the following, if you want:
`local extension = mixin.FCMString():SetLuaString(file_name.LuaString):ExtractFileExtension()`

ThistleSifter (commented):
I think you've done about as much as you can with `FCMFileSaveAsDialog` without adding an extra module. I just left one small suggestion for possible `FCMString` usage.

PR #252, opened 2022-07-13
Create hairpin_creator.lua

cv-on-hub:
Brand new script incorporating a lot of different elements. But no direct user interaction!

### DIFF
--- a/src/hairpin_creator.lua
+++ b/src/hairpin_creator.lua
@@ -1,0 +1,520 @@
+function plugindef()
+    finaleplugin.RequireSelection = true
+    finaleplugin.Author = "Carl Vine after CJ Garcia"
+    finaleplugin.AuthorURL = "http://carlvine.com/lua"
+    finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
+    finaleplugin.Version = "v0.49"
+    finaleplugin.Date = "2022/07/13"
+    finaleplugin.AdditionalMenuOptions = [[
+        Hairpin create diminuendo
+        Hairpin create swell
+        Hairpin create unswell
+    ]]
+    finaleplugin.AdditionalUndoText = [[
+        Hairpin create diminuendo
+        Hairpin create swell
+        Hairpin create unswell
+    ]]
+    finaleplugin.AdditionalDescriptions = [[
+        Create diminuendo spanning the selected region
+        Create a swell (messa di voce) spanning the selected region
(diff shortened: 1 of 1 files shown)


cv-on-hub:
Sorry about the quick re-commit. On reflection the config file was altogether too messy, and is likely to need changing for different uses. So it's all gone into user_settings with modeless position saving. 

Also fixed one placement error with dynamic matching at end of measure.

cv-on-hub:
OK - so I'm taking advantage of the lull to submit final corrections and tweaks. But I'm putting it to bed now until you identify any mistakes requiring correction.

Nick-Mazuk (approved):
Overall, looks good with me. However, I'd like to get some clarification on how this script relates to CJ's existing hairpin script before merging.

https://github.com/finale-lua/lua-scripts/blob/master/src/standalone_hairpin_adjustment.lua

Some potential questions:

- How is this different from the existing script?
- Is this script supposed to replace the existing script?
- Could the script replace the existing script (even if that wasn't the intention)
- If this is different from the existing script, how does the usage differ?

I don't foresee any issues with merging this script, but would just like some clarity to ensure that we're not duplicating work accidentally (as that would get unmaintainable).

cv-on-hub:
1) Massively different from CJ's `standalone` script, which only shifts existing hairpins and dynamics. This is specifically designed to CREATE hairpins from scratch, and shift dynamics to match only if they need it. It is not a dynamic adjuster like CJ's, but a hairpin CREATOR. I pulled his script apart at the outset to work out how to make things join up, but have then moved some distance beyond his original.
2) It is not a replacement, but it does, IMHO, a better job of adjusting vertical heights of hairpins and markings.
3) It cannot replace `standalone`
4) It makes hairpins. I have reverse engineered several of CJ's old scripts, including a much older one, since withdrawn, that DID make hairpins and then balance them with dynamics. That WOULD have been in direct competition, and CJ is using a greatly improved version of that old function within JetStream, but AFAIK hasn't made a separate version for the repository. My version, again IMHO, still does a better job than CJ's revision, even though it uses some of his assumptions and methods. It is more robust, more detailed, and easier to work out how the various elements relate. There is no other `hairpin creator` in the repo.

Nick-Mazuk:
Thanks for the clarifications!

Merging.

cv-on-hub:
Thanks Nick. `hairpin_creator` only provides configuation via option-key, and us KM-freaks really hate selecting menu items manually. The following companion program allows direct configuration of its `user_settings` file, and requires just one more menu item, just for those who need it.

PR #253, opened 2022-07-16
Enable bundling of mixin scripts #2

Nick-Mazuk:
This PR is a redo of #240 which was ultimately buggy and had to be rolled back in #245.

Some notable changes from the previous PR:

1. Instead of relying on a 3rd-party bundler, I just rewrote our existing bundler
2. This change is actually much simpler, resulting in more human-friendly bundles (great for debugging)
3. This PR should be much more stable as it has fairly extensive test coverage
4. Added some quick docs to show how the bundling works. Since most won't need to know how bundling works, I just added a markdown file in the action's folder. That means it won't appear on the website, but will be noticeable if you ever start working on the bundler.
5. Much better heuristic for importing mixins. How, the mixins will be imported whenever the `src/library/mixin.lua` file is imported. You don't need to explicitly write `require("library.mixin")`. Requiring any string that resolves to that mixin file (including `library/mixin` or `library/mixin.lua`) will all cause the mixins to be imported.

Like before, this PR should add support for mixins (both public ones in the `src/mixin` folder and private ones in the `src/personal_mixin` folder.

cc @rpatters1 @ThistleSifter @cv-on-hub


rpatters1 (approved):
I'm not really qualified to review this (lol) but it lgtm.

rpatters1:
If you are ready to merge it, be my guest.

PR #254, opened 2022-07-16
fix bug in trumpet numbering when creating a brass quintet score

Nick-Mazuk:
Title says it all.

### DIFF
--- a/src/score_create_brass_quintet.lua
+++ b/src/score_create_brass_quintet.lua
@@ -36,7 +36,7 @@ local function score_create_brass_quintet_score()
     local staves = {}
 
     staves.trumpet_1 = score.create_staff("Trumpet in C 1", "Tpt. 1", finale.FFUUID_TRUMPETC, "treble")
-    staves.trumpet_2 = score.create_staff("Trumpet in C 1", "Tpt. 1", finale.FFUUID_TRUMPETC, "treble")
+    staves.trumpet_2 = score.create_staff("Trumpet in C 2", "Tpt. 2", finale.FFUUID_TRUMPETC, "treble")
     -- Yes, I know, alto clef for horns. Trust me, though, it's better
     -- to read horns in alto clef when viewing the score in concert pitch.
     -- Alto clef fits the horns' concert pitch range perfectly.


rpatters1 (approved):
Lol I just saw your comment about writing the horn in alto clef. If you *really* want it to be pristine, put the horn in mezzo-soprano clef. Then the lines & spaces are the same as F transposition. (This was a suggestion from one of my comp. teachers: one that I definitely did not take. :-p)

Also, alto clef only gets you so far. To write in the bottom third of the horn's range, you really are going to need bass clef. (I mean, the horn and trombone cover pretty much exactly the same range, being the same length and all.) That said, you rarely encounter those low range notes in a brass quintet.

Alto clef as the horn clef seems definitely like the solution of a trombone player, y'all being so used to it. As a horn player, my C scores alternate the horns between treble and bass clef. I had another comp teacher who said all C clefs should be abolished. (Another suggestion I have ignored, not least because the violists would kill me.)

Cheers

jwink75:
This made me giggle, because I remember testing a version of this for Nick earlier and remember thinking "WHY are the horns in alto clef?" Nick's reasoning was both sound... and hardly conventional :)

Regarding what you said about violists and C clefs:

I just had a chorister graduate whose mother is the principal violist for the Seattle Symphony. Like many violists, she started on violin. Her daughter once confided to me that her mother STILL prefers reading in treble clef, and is apparently STILL 'converting' alto to treble clef in her head!

-Jake
On 7/16/2022 3:43:47 PM, rpatters1 ***@***.***> wrote:
@rpatters1 approved this pull request.
Lol I just saw your comment about writing the horn in alto clef. If you really want it to be pristine, put the horn in mezzo-soprano clef. Then the lines & spaces are the same as F transposition. (This was a suggestion from one of my comp. teachers: one that I definitely did not take. :-p)
Also, alto clef only gets you so far. To write in the bottom third of the horn's range, you really are going to need bass clef. (I mean, the horn and trombone cover pretty much exactly the same range, being the same length and all.) That said, you rarely encounter those low range notes in a brass quintet.
Alto clef as the horn clef seems definitely like the solution of a trombone player, y'all being so used to it. As a horn player, my C scores alternate the horns between treble and bass clef. I had another comp teacher who said all C clefs should be abolished. (Another suggestion I have ignored, not least because the violists would kill me.)
Cheers

rpatters1:
> Like many violists, she started on violin. 

Conversation overheard between professional cellist and 2 professional violists:

Cellist: My son says he wants to play the viola. (Her son is only 6 or 7.) Should he switch from violin?
Both Violists (in unison): Absolutely not!

They went on to explain that it's because the entry-level orchestra part for viola are soul-crushingly boring, esp. compared to the fiddle parts. But I was think, "Somebody has to play them." I guess not the kids with promise, though.

Nick-Mazuk:
In my opinion, there should only be 5 clefs:

1. Treble clef
2. Treble clef up the octave with the 8
3. Bass clef
4. Bass clef down the octave with the 8
5. Alto clef

For 2 and 4, these are essentially for octave transposing instruments like piccolo and string bass. Then alto clef fits nicely in that middle zone between the two clefs. With these five clefs, you cover every standard instrumental range fairly well—both in C score and transposed.

And in my opinion, the only reason why people don't like alto clef is that almost no one uses it. It really is underused and is incredibly applicable if you're willing to break away from convention.

The other C clefs only provide an incremental benefit to alto clef (at best) while causing a lot more confusion and should be abolished.

> Alto clef as the horn clef seems definitely like the solution of a trombone player, y'all being so used to it.

To be honest, I'm actually more used to tenor. Only old alto trombone parts are written in alto clef. Everything else is either bass (most common by far) or tenor (only common in standard orchestra rep).



> To write in the bottom third of the horn's range, you really are going to need bass clef.

Yes, definitely. When I need bass clef, I'll use it. Alto clef just fits very well for the vast majority of horn parts. And probably more importantly, everyone I've convinced to write horn in alto clef (for C scores) tended to write better horn parts. The tendency is almost always to avoid ledger lines—especially notes with many ledger lines. So horn parts tend to drift towards the higher end of the horns' range because that's the most natural thing to do in concert treble clef. Now a good orchestrator won't exclusively write horn in that range, but it's easier to get in the habit of doing so if you read concert alto clef.

rpatters1:
You could get rid of all the C clefs by include treble-tenor in your pantheon and eliminating alto clef. Treble tenor (treble with the 8 *under* it) is only a half step off from alto in terms of lines and spaces, falling exactly between alto and tenor C clefs. Furthermore, I would argue that treble-tenor is more common than alto clef, being used exclusively in vocal writing.

But, boy, would *that* ever bring out the torch-wielding violists.

rpatters1:
P.S. I agree that writing horn parts on standard treble frequently produces shite horn parts. My mantra about contemporary horn parts is, they are almost invariably too high, don't have enough rests, and ignore the low range. (I can't tell you how many PhD composers have looked at me in astonishment when I tell them that Beethoven and his ilk, without hesitation, wrote Bb1—the bottom note on a bassoon—for the 2nd horn.)

rpatters1:
P.P.S. I would argue that the opposite has happened to the trombone. By making at a bass clef instrument, trombone parts are frequently too low.

Nick-Mazuk:
> You could get rid of all the C clefs by including treble-tenor in your pantheon and eliminating alto clef. Treble tenor (treble with the 8 _under_ it) is only a half step off from alto in terms of lines and spaces, falling exactly between alto and tenor C clefs. Furthermore, I would argue that treble-tenor is more common than alto clef, being used exclusively in vocal writing.

Oh yeah. I never thought of using that clef. I'll try it out.

> P.S. I agree that writing horn parts on standard treble frequently produces shite horn parts. My mantra about contemporary horn parts is, they are often too high, don't have enough rests, and ignore the low range. (I can't tell you how many PhD composers have looked at me in astonishment when I tell them that Beethoven and his ilk, without hesitation, wrote Bb1—the bottom note on a bassoon—for the 2nd horn.)

+1 to all of this

> P.P.S. I would argue that the opposite has happened to the trombone. By making it a bass clef instrument, trombone parts are frequently too low.

Again, +1 to this. Tenor trombone parts are too low, and bass trombone parts are too high. Trombone is best when it's sitting outside the bass clef staff (in either direction).

rpatters1:
I tried using treble-tenor for horn parts a few times, but it seemed weird to me. I ended up going back to plain old treble/bass alternation, awkward though it is.

PR #256, opened 2022-07-16
add swap staves script from Jari's website

Nick-Mazuk:
The title says it all. Surprised we haven't added this one yet.

There's room for potential improvement (e.g., resetting the original clefs), but thought I'd add it and we can improve it later.


rpatters1 (approved):
I guess I should investigate how `CopyMusic` and `PasteMusic` work. It may be that preserving the clefs is better accomplished there. But in the meantime, this is fine.

PR #261, opened 2022-07-17
Enable bundling of personal scripts while preventing committing them

Nick-Mazuk:
Previously, the bundler would ignore any scripts marked as personal (i.e., `personal_*`). However, this means it's nearly impossible to share scripts.

Now, you can run the bundler locally ([described here](https://github.com/finale-lua/lua-scripts/blob/master/.github/actions/bundle/README.md)) and it will bundle those scripts as well. Git will now also still ensure those scripts won't be committed into the repository.

As a side effect, this PR also allows one to create personal libraries. Any file in the library can be prefixed with `personal_` and it will not be committed into the repo.


Nick-Mazuk:
I know we should really do code reviews, but this one seems minor enough that I'm going to go ahead and merge it. That way I don't accidentally commit a bunch of personal scripts I shouldn't make public yet if I get around to doing another PR today.

If there is an issue, leave a comment. Happy to fix it in a follow-up PR.

PR #262, opened 2022-07-17
Harp Pedal Wizard

jwink75:
- Now uses library functions the way they were intended.
- Updated GUI to account for disparity between Mac/Windows

### DIFF
--- a/src/harp_pedal_wizard.lua
+++ b/src/harp_pedal_wizard.lua
@@ -1,0 +1,1506 @@
+function plugindef()
+  -- This function and the 'finaleplugin' namespace
+  -- are both reserved for the plug-in definition.
+  finaleplugin.RequireSelection = false
+  finaleplugin.Author = "Jacob Winkler"
+  finaleplugin.Copyright = "2022"
+  finaleplugin.Version = "2.0"
+  finaleplugin.Date = "2022-07-17"
+  finaleplugin.MinJWLuaVersion = 0.63 -- https://robertgpatterson.com/-fininfo/-rgplua/rgplua.html
+  return "Harp Pedal Wizard", "Harp Pedal Wizard", "Creates Harp Diagrams and Pedal Changes"
+end
+
+local library = require("library.general_library")
+local configuration = require("library.configuration") -- Having trouble getting this to work!
+
+--------------------------------------
+-- borrowed JetStream functions - will replace later with repo equivalents!
+function get_used_font_name(standard_name)
+  local font_name = standard_name
+  if string.find(os.tmpname(), "/") then
+    font_name = standard_name
+  elseif string.find(os.tmpname(), "\\") then
+    font_name = string.gsub(standard_name, "%s", "")
+  end
+  return font_name
+end
+
+function get_def_mus_font()
+  local fontinfo = finale.FCFontInfo()
+  if fontinfo:LoadFontPrefs(finale.FONTPREF_MUSIC) then
+    return get_used_font_name(fontinfo:GetName())
+  end
+end
+--------------------------------------
+function harp()
+  local script_name = "harp_pedal_wizard"
+  local config = {root = 2, accidental = 1, scale = 0, scale_check = 1, chord = 0, chord_check = 0, diagram_check = 1, names_check = 0, partial_check = 0, stack = 1, pedal_lanes = 1, last_notes = "D, C, B, E, F, G, A"}
+  local partial = false
+  local changes = false
+  local stack = true
+  local pedal_lanes = true
+  local direct = false
+  local override = false
+  local context =
+  {
+    window_pos_x = window_pos_x or nil,
+    window_pos_y = window_pos_y or nil
+  }
+
+  local SMuFL = library.is_font_smufl_font(nil)
+  harpstrings = {}
+  diagram_string = finale.FCString()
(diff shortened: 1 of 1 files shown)


src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1506 @@
...
+
+      add_expression:SetStaff(staff_num)
+      add_expression:SetMeasurePos(measure_pos)
+      add_expression:SetID(diag_ted)
+      add_expression:SaveNewToCell(and_cell)
+
+-------
+      ::error1::
+      finenv.EndUndoBlock(true)
+      if error == true then -- Tried this as if error then, it broke the script
```
rpatters1, line 495: The reason `if error` broke the script is that `error` is a pre-defined global function in Lua, and unlike everywhere else you are using `error` as a variable, in this function you never defined it as a local variable. This code will never take the error path, because error will never be `true`.

Generally speaking I would avoid naming variables `error`. But for this case, if you add a `local error = false` at the top of the function, it should fix problem with `if error`.

rpatters1 (changes requested):

src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1506 @@
...
+        end
+
+--                if dialog:ExecuteModal(nil) == finale.EXECMODAL_OK then
+--                end -- if dialog:Execute...
+
+      end -- get_harp
+      harp_dialog()
+    end -- harp
+
+    harp()
```
Nick-Mazuk, line 1506: Seems like there might be some "end" statements missing, or a few "unindents"? It seems weird to have a file end with lines still indented.

src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1506 @@
...
+--
+      function format_ctrl(ctrl, h, w, st)
+        ctrl:SetHeight(h)
+        ctrl:SetWidth(w)
+        str.LuaString = st
+        ctrl:SetText(str)
+      end -- function format_ctrl
+--
+      local dialog = finale.FCCustomLuaWindow()
+      if nil ~= context.window_pos_x and nil ~= context.window_pos_y then
```
Nick-Mazuk, line 766: A little nit, but `context.window_pos_x ~= nil` and `context.window_pos_y ~= nil` would be better here.

The way you originally wrote it is called a [yoda condition](https://en.wikipedia.org/wiki/Yoda_conditions). While they are valid syntax and there are a few minor advantages of them, they can in general make it harder to read code.

src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1506 @@
...
+            end -- for j
+            if not found then
+              error = true
+              goto error
+            end
+            scale_deg = scale_deg + 1
+            last = "B"
+          end
+          if last == "B"  and scale_deg <= 7 then
+            local found = false
```
Nick-Mazuk, line 608: Guess this review is going to be me being picky about minor code style.

One general rule of thumb with booleans is that the variable should signify that it's a boolean. Often, you'll want to prefix them with `is_` or `should_`, or something else that describes the purpose of the variable. That way, you can read `is_found` and instantly know that variable is a boolean.

It also makes if statements read more like English:

```lua
if is_found then
```

src/harp_pedal_wizard.lua
```diff
@@ -0,0 +1,1506 @@
...
+        local result = ui:AlertYesNo("There seems to be a problem with your harp diagram. \n Would you like to try again?", NULL)
+--                if result == 2 then harp_dialog() end
+      end -- error
+      if diag_asn then
+        ui:AlertInfo("There is already a harp diagram assigned to this region.", NULL)
+      end
+    end -- function add_pedals
+
+    -------
+    function harp_scale(root, scale, diag, chd, partial)
```
Nick-Mazuk, line 506: It would be best to have more descriptive variable names—especially for function arguments. I assume `diag` means diagram? However, I have no idea what `chd` is from looking at the code.

src/harp_pedal_wizard.lua
jwink75: renamed all instances of error to variables according to the function they are in (e.g. scale_error)

Nick-Mazuk (approved):

rpatters1 (approved):

rpatters1:
I'm cool with merging it.

Nick-Mazuk:
My comments are very nit picky, so I'll go ahead and merge this PR. @jwink75, if you want to resolve my comments, you can go ahead and create a follow-up PR.

src/harp_pedal_wizard.lua
jwink75: I don't remember where I copied this from, but it is definitely from some other script that I stole it from. I never really got this working (retaining the window position), but I'll go ahead and change

src/harp_pedal_wizard.lua
jwink75: What are unindents?

All I know is that if I add any more 'end' statements it breaks...

src/harp_pedal_wizard.lua
jwink75: yoda condition... I just got it!

```
if you_become == 900 then
    look_so_good = false
end
```

PR #265, opened 2022-07-18
Update hairpin_creator.lua

cv-on-hub:
Revised script to use mixin config dialog and ScriptGroupName. Can't get Mixin to deliver dialog:RestorePosition() so it's implemented literally.

Please don't commit to /dist until the library inclusion corruption is fixed!

### DIFF
--- a/src/hairpin_creator.lua
+++ b/src/hairpin_creator.lua
@@ -25,7 +25,7 @@ function plugindef()
         hairpin_type = -1 -- "swell"
         hairpin_type = -2 -- "unswell"
     ]]
-    finaleplugin.MinJWLuaVersion = 0.62
+    finaleplugin.MinJWLuaVersion = 0.63
     finaleplugin.ScriptGroupName = "Hairpin creator"
     finaleplugin.Notes = [[
         This script creates hairpins spanning the currently selected music region. 
@@ -544,7 +544,7 @@ function create_dialog_box()
         local function make_static(msg, horiz, vert, width, sepia)
             local static = dialog:CreateStatic(horiz, vert):SetText(msg)
             static:SetWidth(width)
-            if sepia then
+            if sepia and static.SetTextColor then
                 static:SetTextColor(204, 102, 51)
             end
         end
--- a/src/hairpin_creator.lua
+++ b/src/hairpin_creator.lua
@@ -3,8 +3,8 @@ function plugindef()
     finaleplugin.Author = "Carl Vine after CJ Garcia"
     finaleplugin.AuthorURL = "http://carlvine.com/lua"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "v0.58"
-    finaleplugin.Date = "2022/07/22"
+    finaleplugin.Version = "v0.59"
+    finaleplugin.Date = "2022/07/24"
     finaleplugin.AdditionalMenuOptions = [[
         Hairpin create diminuendo
         Hairpin create swell
@@ -51,7 +51,7 @@ function plugindef()
         first move one or more dynamic to the lowest point you need. 
         
         To change the script's default settings hold down the `shift` or `alt` (option) key when selecting the menu item. 
-        (This may need special treatment when using a keystroke macro program like KeyboardMaestro). 
+        (This might need special treatment when using a keystroke macro program like KeyboardMaestro). 
         For simple hairpins that don't mess around with trailing barlines and dynamics try selecting 
         `dynamics_match_hairpin` and de-selecting the other options.
     ]]
@@ -288,10 +288,12 @@ local function lowest_note_element(rgn)
 end
 
 local function expression_is_dynamic(exp)
-    if not exp:IsShape() and exp.StaffGroupID == 0 then
+    if not exp:IsShape() and exp.Visible and exp.StaffGroupID == 0 then
         local cd = finale.FCCategoryDef()
         local text_def = exp:CreateTextExpressionDef()
-        if text_def and cd:Load(text_def.CategoryID) then
+        local font_info = text_def:CreateTextString():CreateLastFontInfo() -- ignore hidden expressions
+
+        if text_def and cd:Load(text_def.CategoryID) and not font_info.Hidden then
             if text_def.CategoryID == finale.DEFAULTCATID_DYNAMICS or string.find(cd:CreateName().LuaString, "Dynamic") then
                 return true
             end
@@ -573,9 +575,10 @@ function create_user_dialog() -- attempting MODELESS operation
         end
     end
     -- measurement unit options
-    dialog:CreateStatic(0, (#dialog_options + 1) * y_step + 24 - mac_offset):SetText("Units:")
+    local y_current = (#dialog_options + 1.6) * y_step
+    dialog:CreateStatic(x_offset[2] - 40, y_current ):SetText("Units:") -- + mac_offset
     dialog:SetMeasurementUnit(config.measurement_unit)
-    dialog:CreateMeasurementUnitPopup(40, (#dialog_options + 1) * y_step + 20)
+    dialog:CreateMeasurementUnitPopup(x_offset[2], y_current)
 
     -- InitWindow: set config values
     dialog:RegisterInitWindow(function(self)
@@ -631,6 +634,7 @@ function action_type()
             global_dialog:SetRestorePositionOnlyData(config.window_pos_x, config.window_pos_y)
             global_dialog:RestorePosition()
         end
+        global_dialog.OkButtonCanClose = true
         finenv.RegisterModelessDialog(global_dialog)
         global_dialog:ShowModeless()
     else
--- a/src/hairpin_creator.lua
+++ b/src/hairpin_creator.lua
@@ -3,8 +3,8 @@ function plugindef()
     finaleplugin.Author = "Carl Vine after CJ Garcia"
     finaleplugin.AuthorURL = "http://carlvine.com/lua"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "v0.55" -- (now with Mixin)
-    finaleplugin.Date = "2022/07/18"
+    finaleplugin.Version = "v0.58"
+    finaleplugin.Date = "2022/07/22"
     finaleplugin.AdditionalMenuOptions = [[
         Hairpin create diminuendo
         Hairpin create swell
@@ -35,8 +35,8 @@ function plugindef()
         Hairpins are positioned vertically to avoid colliding with the lowest notes, down-stem tails, 
         articulations and dynamics on each staff in the selection. 
         Dynamics are shifted vertically to match the calculated hairpin positions. 
-        Dynamics in the middle of a hairpin span will also be levelled, so 
-        giving them an opaque background will make them appear to sit "above" the hairpin. 
+        Dynamics in the middle of a hairpin will also be levelled, so 
+        give them an opaque background to sit "above" the hairpin. 
         The script also considers `trailing` notes and dynamics, just beyond the end of the selected music, 
         since a hairpin is normally expected to end just before the note with the destination dynamic. 
 
@@ -50,9 +50,9 @@ function plugindef()
         It will find the lowest acceptable vertical offset for the hairpin, but if you want it lower than that then 
         first move one or more dynamic to the lowest point you need. 
         
-        To change the script's default settings hold down the `alt` / `option` key when selecting the menu item. 
-        (This may not work when invoking the menu with a keystroke macro program). 
-        For simple hairpins that don't mess around with trailing barlines try selecting 
+        To change the script's default settings hold down the `shift` or `alt` (option) key when selecting the menu item. 
+        (This may need special treatment when using a keystroke macro program like KeyboardMaestro). 
+        For simple hairpins that don't mess around with trailing barlines and dynamics try selecting 
         `dynamics_match_hairpin` and de-selecting the other options.
     ]]
     return "Hairpin create crescendo", "Hairpin create crescendo", "Create crescendo spanning the selected region"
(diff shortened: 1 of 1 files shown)


ThistleSifter:
Hmm, what's the issue with position restoring? It's meant to be completely automatic when using `FCXCustomLuaWindow`.

cv-on-hub:
yes - I thought it was automatic and really can't say what's wrong. I tried to imitate the way Robert set it up in `transpose_by_step.lua` but it refuses to retain dialog position unless I set it up explicitly using config values the way it is in this script. I'd be very grateful if you can spot what I'm doing wrong.

ThistleSifter:
Ah, the issue is that `prefs_dialog` is a local variable, so every time the script runs it's creating a fresh dialog. If you make it a global variable (and don't overwrite it) then it will persist across script executions. I'll take a thorough look tomorrow and I can give you some more detailed suggestions then.

If you have any questions about mixins, feel free to ping me. I know they're still new and can take some getting used to.

cv-on-hub:
That's great @ThistleSifter - many thanks. I realise now that the **other** reason it hasn't retained position is that the dialog is only invoked when option/shift are used, so that the "retained" state may or may not include the dialog state.

Also the dialog must disappear after activation, so if I use:`prefs_dialog = prefs_dialog or create_dialog_box()` then the dialog thinks it's still active and doesn't re-activate. Also for `OkButtonCanClose` to be recognised seems like option/shift has to be active, which caused me some confusion early on!

I might have to keep using the literal position setting for this "occasional" use of dialog windows. (RetainLuaState also doesn't retain across re-boots).

ThistleSifter:
Do you specifically __not__ want the dialog to show when the shift key is pressed? Maybe @rpatters1 can chime in here, but my understanding was that the behaviour we were going for was showing the dialog when Shift _or_ Alt was pressed.
&nbsp;
>Also for OkButtonCanClose to be recognised seems like option/shift has to be active, which caused me some confusion early on!
>
Hmm, it should be recognised for either Alt _or_ Shift, in `RunModeless`. At least that's what the code does. Is this a KM thing? I've been relying on @rpatters1 for testing on Mac since I only have Windows, and I'm pretty sure it worked as expected for him (well, it was originally his code anyway).

Retaining dialog position across reboots is an interesting conundrum. On the one hand, I can see that it would be nice to have it always in the same place. But on the other hand, it would only be for scripts that have a config file. Does this mean that every script with a dialog box should have a config file, just to ensure consistent placement? That could get a bit heavy. I'd be interested to see what others think or if there is anything that RGPLua can do.

rpatters1:
I am not eager to try to build persistence of dialog position into RGP Lua. It would actually have to be built into the PDK Framework, and I try to avoid breaking that as much as I can, especially the U.I. features. The code is very complex.

I lean towards making window persistence across Finale executions optional, based on whether a user settings file exists. The good news is, the `configuration` library tells you this, and the `general_library` has a generic function that calculates the name of the script from `finenv.RunningLuaFilePath`. So it should be possible to build this into `RunModal` fairly easily. We could standardize the `x` and `y` variable names. The only tricky part is writing them out to the user settings when the window closes.

rpatters1:
BTW: the latest version of RGP Lua fixed the issue of `RestoreSize` not being able to restore smaller than an arbitrary size. The fix was simply making the minimum size really tiny so that it would never be an issue. I didn't try to muck around in the PDK Framework to address the problem directly, but it works now for any reasonable case. Whether we want to add it back into `FCXCustomLuaWindow` is an open question.

cv-on-hub:
> Do you specifically **not** want the dialog to show when the shift key is pressed?

No - I just thought offering two options added no functionality and was more confusing than one. (I had worked out that `mixin` will accept either, in line with Robert's assumptions).

> every script with a dialog box should have a config file ...

There is a `user_settings` file in this case, so that's an easy choice. I hadn't even thought of consistent window placement before Robert mentioned it re `RetainLuaState`, and although it's a nice feature is hardly crucial. For me, `RetainLua` is perfectly reasonable and functional  for scripts without config files.

rpatters1:
About ALT/SHIFT, one of the advantages of using mixin is to get consistent behavior. :-)

cv-on-hub:
> I figured it would be easier to just attach the whole file.

Thank you so much @ThistleSifter. That's absolutely brilliant, especially `:SetText():SetWidth()`. Will rework appropriately, and use shift/alt for consistency. (Now that we have color I'm going to stick with sepia to delineate explanations from titles - but a darker version!) I was so obsessed with conforming to modeless protocol that I forgot that what I REALLY need here is a modal dialog.

I'm not convinced about changing measurement units since that will mean alternately retrieving floats or integers .. or does `mixin` handle all that automatically and provide well-behaved integer EVPU at the end?

cv-on-hub:
Just tried with your mods. The "Units" popup system is amazing. 

Just one question: when converting from another unit back to EVPU, which is what my script requires, it leaves approximated float values (as you say - just like Finale's own units system). I assume it's up to my script to convert these back to EVPUs? (or at least round them up to corresponding integers using math.floor(x + 0.5)?) I can't quite see where `mixin` provides that integer re-conversion.

ThistleSifter:
My mistake. If you replace `SetMeasurement` and `GetMeasurement` with `SetMeasurementInteger` and `GetMeasurementInteger` respectively, it will take care of rounding for you.

cv-on-hub:
Equally brilliant. Now I just can't work out the name of the `CreateMeasurementUnitPopup` to access the value of the `MeasurementUnit` to store in the config table (and user_settings) and preset the "Unit" popup for the next call.

ThistleSifter:
The measurement unit comes from the window itself (the popup is just a way of allowing the user to change it) so if you want to get the window's current unit, you can use the `GetMeasurementUnit` method of `FCXCustomLuaWindow` (you can also use `SetMeasurementUnit` to set the unit programmatically). If you want to give the popup a name, you can do so in the same way as other controls by adding a third value to `CreateMeasurementUnitPopup` but there's nothing useful you can do with it other than change its size/position, so I didn't specify one in the file I posted.

cv-on-hub:
Since I need a modal dialog, how do I preserve the user's preferred unit at each invocation? The window doesn't have the user's "current unit" until they've chosen to progress, and so their choice needs to be brought in from the config at startup and stored back in config after action. This seems to require setting either the popup selection or the window unit value at the start and getting the popup or window state back afterwards.

(And sorry to be such a pest!)

ThistleSifter:
Could you post your updated code?

cv-on-hub:
Got it. At start of dialog:
`dialog:SetMeasurementUnit(config.measurement_unit)`
and at in the OK handler:
`config.measurement_unit = self:GetMeasurementUnit()`

Only problem now is that when choosing Picas it makes negative numbers positive! (All other units seem OK). Still a work in progress, but attached:
[hairpin_creator.txt](https://github.com/finale-lua/lua-scripts/files/9146347/hairpin_creator.txt)

Can't get modal window to restore window position.

ThistleSifter:
This should work (I also left a couple more comments with `@TS` ).
[hairpin_creator.txt](https://github.com/finale-lua/lua-scripts/files/9155674/hairpin_creator.txt)

A couple more points:
- Automatic position restoration won't work unless you're using `finenv.RetainLuaState` (maybe that should be documented better). Since you're loading the position from a config file, you will still need to handle that yourself. I'll look into a way of handling automatic saving/loading from a config file in the mixin library, but for now this will work.
- `prefs_dialog` only needs to be global under two scenarios: 1) If it's a modeless dialog. 2) If you're using `finenv.RetainLuaState` (Or both). Any other situation, like this one, and you can leave it local.

Thanks for pointing out the issue with Picas, I'll look into it once I've finished up the control state PR.

ThistleSifter:
Hmm, I'm really not sure. I just tested it again (using the file you just posted) and it works fine for me (v26.3, Win). Are you using the latest mixin code from the repo? Maybe someone else could test it and check?

Also, the issue with Picas doesn't happen for me either (the sign stays the same whether switching in or out). The switching code just uses the standard `FCCtrlEdit` `GetMeasurement` and `SetMeasurement` methods to do it, so if it doesn't work, I'm thinking there might be a bug in the PDK or Finale. @rpatters1 ?

cv-on-hub:
I just downloaded `library` and `mixin` files from:
https://github.com/finale-lua/lua-scripts/tree/master/src

Still won't follow `RestorePosition()` and pica polarity problem still occurring on Mac.

ThistleSifter:
Here's quick test script. Can you run it and let me know what sort of results it gives you? Enter a pica value in the text box and press the button to display it in EVPUs.
```lua
local str = finale.FCString()

str.LuaString = "Enter Picas"
dialog:CreateStatic(0, 0):SetText(str)

local edit = dialog:CreateEdit(0, 30)

local btn = dialog:CreateButton(0, 60)
str.LuaString = "to EVPUs"
btn:SetText(str)

	if ctrl:GetControlID() == btn:GetControlID() then
		finenv.UI():AlertInfo(tostring(edit:GetMeasurement(finale.MEASUREMENTUNIT_PICA)), "EVPUs")
	end
end)

dialog:ExecuteModal(nil)
```

cv-on-hub:
I get the dialog window pasted below. The button does nothing
pica_window

ThistleSifter:
Does it work if you take out the second argument to `AlertInfo`? (the string `"EVPUs"`)

If not, try this:
```lua
local str = finale.FCString()

str.LuaString = "Enter Picas:"
dialog:CreateStatic(0, 0):SetText(str)

local edit = dialog:CreateEdit(70, 0)

local static = dialog:CreateStatic(0, 30)

    str.LuaString = "EVPUs: " .. edit:GetMeasurement(finale.MEASUREMENTUNIT_PICAS)
    static:SetText(str)
end)

dialog:ExecuteModal(nil)
```

rpatters1:
I'm not seeing a problem (on Mac) with either version of the test code. The sign is always the same for me.

cv-on-hub:
Yes - that works fine now. (Pic attached). Mysterious why rpatters1 has no problem on Mac. What could be problematic in my setup? (macOS 12.4, Fin 27.2, RGPLua 0.63)

pica_window2

cv-on-hub:
The problem, at least from this end, is that `CreateMeasurementUnitPopup` is using the "textified" version of Picas, like "-0p3.5" (-14 EVPU). When converting to any other unit, this becomes positive.

Regarding window positioning, it seems (on my system) that `RunModeless()` ignores requested window position but `RegisterModelessDialog()` ... `ShowModeless()` obeys the request.

cv-on-hub:
Here's a revised (modeless) version that does what I want, respects window position and I think incorporates all of @ThistleSifter's advice:

[hairpin_creator_cv058.txt](https://github.com/finale-lua/lua-scripts/files/9163466/hairpin_creator_cv058.txt)

ThistleSifter:
@cv-on-hub If it's okay with you (and you have the time), I'd like to debug the issues with Picas and RestorePosition further with you and see if we can get to the bottom of it (I don't like an unsolved mystery!) If you're keen, can you shoot me an email?

cv-on-hub:
messaging off-thread

ThistleSifter:
@cv-on-hub did some more in depth testing and these were the results:

**Picas**
A negative value in Picas is incorrectly read as a positive value on his setup:
- From `GetMeasurement`, `GetMeasurementInteger` and `GetMeasurementEfix`
- In both `FCCtrlEdit` and `FCString` (no mixins involved)
- Using both `RGPLua` and `JWLua`
- When the value in Picas is between `0` and `-1` (ie it starts with `-0p`)
Eg -`0p3.25` is returned as `13` EVPUs instead of `-13`

Finale's in-built dialogs (eg Document Options) don't display this issue and convert correctly.

I can include a fix for this in the mixins for `FCMCtrlEdit`, `FCMString` and `FCXMeasurementEdit` but I wonder if it's able to be addressed in RGPLua?

**RestorePosition**
Doesn't work for dialogs run with `ExecuteModal` (either standard `FCCustomLuaWindow` or mixin `FCXCustomLuaWindow`) on his setup. Dialogs run with `ShowModeless` have their position restored correctly (either standard or mixin).

This one I can't address in the mixins, so if it can be fixed, it would need to be in `RGPLua`.

@rpatters1 Any thoughts or suggestions?

rpatters1:
Allowing `RestorePosition` for modal dialogs requires a change that could affect a lot more than RGP Lua, so I am going to suggest we put that on the back burner.

For the Pica conversion issue I need a sample script that displays the problem. If an RGP Lua fix is needed, I'll fix it.

ThistleSifter:
Oh? `RestorePosition` seemed to work okay for me on Windows with modal dialogs.

Here is a simple script for Picas
```lua
local str = finale.FCString()
str.LuaString = "-0p3.25"
print(str:GetMeasurement(finale.MEASUREMENTUNIT_PICAS))
```
It should give `-13` but it gives `13` instead.

ThistleSifter:
There was also the one I posted here: https://github.com/finale-lua/lua-scripts/pull/265#issuecomment-1191366256
Which should display `-13` if you type `-0p3.25` in the text box, but for @cv-on-hub it gives `13`. (for me, it works correctly)

rpatters1:
Yes, I think it probably is a mac-only issue.

ThistleSifter:
Actually, the `FCString` script two comments up doesn't work for me either.

rpatters1:
The issue with pica conversion is the zero:

-**0**p3.25

When the PDK Framework code splits the two chunks apart it loses the sign because "-0" isn't negative. So, e.g., "-1p3.25" should work correctly.

I'm going to submit a change and hopefully get this fix into the next version of RGP Lua.

cv-on-hub:
Re: https://github.com/finale-lua/lua-scripts/pull/265#issuecomment-1191366256 ... worth noting (on Mac) that anything following the "p" is given the wrong polarity. So "1p3" gives 60 EVPU (48 + 12), but "-1p3" gives -36 (-48 + 12)!!!

rpatters1:
Yes, that too. But there should be no difference between mac and win.

cv-on-hub:
And noting that this error has been corrected within Finale where multiple measurement units are employed in dialogs.

rpatters1:
Finale uses different code. Right now the conversion code for RGP Lua is in the PDK Framework.

rpatters1:
Never mind, I found the reason windows works and mac doesn't.

rpatters1:
I should add, windows works with `FCCtrlEdit::GetMeasurement`. The behavior is the same on both platforms for `FCString::GetMeasurement`. This has to do with the utterly different implementations of dialog boxes on mac and win.

rpatters1:
`FCString::SetMeasurement` looks broken for picas as well. Can someone please test?

ThistleSifter:
It seems to work okay for me. Do you have an example of a broken Pica/EVPU value pair that I could check with?

rpatters1:
No, on further analysis `SetMeasurement` works correctly. The problem is just `FCString::GetMeasurement` which affects `FCCtrlEdit` on Mac.

rpatters1:
Ok, I've submitted a fix. I recommend that `mixin` just let it be wrong for pre-0.64 versions of RGP Lua.

cv-on-hub:
This has been a great journey of wonder and discovery, but somewhere at its start was a script to be considered for commitment. The commit following incorporates whatever it can from the above, ignores invisible or hidden dynamics, and has shifted the "Units" menu. Pica measurements will remain faulty, like all other such facilities, until RGP 0.64.

ThistleSifter:
Looks good. The only thing I have to add is that `FCXCustomLuaWindow` automatically registers a modeless dialog for you (https://github.com/finale-lua/lua-scripts/blob/master/src/mixin/FCXCustomLuaWindow.lua#L417), so you can save yourself a line if you want.

Nick-Mazuk:
> > ... why it is a good idea to bury that call ...
> 
> From your phrasing it felt you were building up to say NOT a good idea!

I can't speak to Robert's actual thoughts, but my guess is that `RegisterModelessDialog` needs to be called for technical reasons, but it's overhead for a developer. So it works best when in a library function. Developers won't need to know to call it, and by using the library you can be sure it's always called!

Nick-Mazuk:
Hey @cv-on-hub, just wanted to check in to see if this PR was ready for a review and merge, or if there were still things you wanted to work on.

> Please don't commit to /dist until the library inclusion corruption is fixed!

This is fixed with #271

cv-on-hub:
Thanks for checking in @Nick-Mazuk. All ready to go.

src/hairpin_creator.lua
```diff
@@ -25,7 +25,8 @@ function plugindef()
         hairpin_type = -1 -- "swell"
         hairpin_type = -2 -- "unswell"
     ]]
-    finaleplugin.MinJWLuaVersion = 0.62
+    finaleplugin.MinJWLuaVersion = 0.63
+    finaleplugin.ScriptGroupName = "Hairpin creator"
```
Nick-Mazuk, line 29: I also added an option `finaleplugin.ScriptGroupDescription` which would then be the short description on the website instead of "Create crescendo spanning the selected region". Want to add it?

Other than that, LGTM.

Nick-Mazuk (approved):

src/hairpin_creator.lua
cv-on-hub: > Want to add it?

Yes please!

rpatters1:
The current version in the master branch is throwing a lot of lint errors. Are these fixed in the PR?

rpatters1:
Whoops, my mistake. It's the harp pedal wizard throwing lint errors. No lint problems here.

jwink75:
Uh oh… I’m going to need to look up lint errors :/

Anything I need to do?

Nick-Mazuk:
> Uh oh… I’m going to need to look up lint errors :/ Anything I need to do?

Need to go to bed so don't have much time to explain, but here are the docs:

https://www.finalelua.com/docs/getting-started/style-guide#automated-styling-with-vs-code-linter

PR #266, opened 2022-07-18
Fix harp pedal wizard

jwink75:
- Resolved a problem where incorrect values were being written to 'config.last_notes'

### DIFF
--- a/src/harp_pedal_wizard.lua
+++ b/src/harp_pedal_wizard.lua
@@ -1,13 +1,13 @@
 function plugindef()
-    -- This function and the 'finaleplugin' namespace
-    -- are both reserved for the plug-in definition.
-    finaleplugin.RequireSelection = false
-    finaleplugin.Author = "Jacob Winkler"
-    finaleplugin.Copyright = "2022"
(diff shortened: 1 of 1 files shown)


jwink75:
Sure.

line 144: This is probably an unnecessary addition, but in case it encounters a situation where the various harpstrings are not defined, it will force a reset to natural-position pedals.

line 385: In most cases I think a "problem with the harp diagram" is going to result from wonkiness with the "last_notes" setting. I added this to force a reset to natural-position pedals.

lines 850-880: did some additional formatting of these checkbox controls to deal with more Windows UI stuff.

line 1467: This was the underlying root of the problem, and I don't know why it started acting wonky. I had it set up to iterate through the first 6 values of a table and build up a string from that, and for some reason it was always starting with the 7th element (so if I had {1, 2, 3, 4, 5, 6, 7} the string would end up "71234567". I changed it to just brute force the string together!


On 7/18/2022 12:34:34 PM, Nick Mazuk ***@***.***> wrote:
Looks like the entire file was reformatted, making the diff noisy. Could you real quickly mention the parts you edited?

PR #268, opened 2022-07-22
Create 'region_replicate_music.lua'

jwink75:
Submitting on behalf of Michael McLennan.

Jake notes: This is a simple but elegant little script! I wish it actually worked like the Sibelius/Dorico version, where the repetition is at any level, and not just at the measure level. Maybe it wouldn't be hard to adapt to that... But for now, this is pretty slick! I renamed it region_replicate_music rather than just replicate_music, to try to make it fit the repo style better.

### DIFF
--- a/src/region_replicate_music.lua
+++ b/src/region_replicate_music.lua
@@ -3,14 +3,16 @@ function plugindef()
     finaleplugin.Version = 1.0
     finaleplugin.Copyright = "2022/01/03"
     finaleplugin.HandlesUndo = true
+    finaleplugin.Notes = [[
+    Inspired by the 'r' key in Sibelius, this script copies the selected music, and pastes it directly to the right.
+Works with a single or multiple measures.
+When activated with a shortcut or hotkey, ultra fast replication is possible.
+]]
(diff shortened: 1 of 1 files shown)


src/region_replicate_music.lua
```diff
@@ -0,0 +1,32 @@
...
+    finaleplugin.Copyright = "2022/01/03"
+    finaleplugin.HandlesUndo = true
+return "Replicate Music", "Replicate Music", "Inspired by the 'r' key in Sibelius, this script copies the selected music, and pastes it directly to the right"
+end
+
+--[[
+Inspired by the 'r' key in Sibelius, this script copies the selected music, and pastes it directly to the right.
+Works with a single or multiple measures.
+When activated with a shortcut or hotkey, ultra fast replication is possible.
+--]]
```
Nick-Mazuk, line 13: Could we move this to the `finaleplugin.Notes` field so it shows up on the website?

When that's done, LGTM and I'll merge.

Nick-Mazuk (changes requested):

Nick-Mazuk (approved):
Looks good to me.

PR #269, opened 2022-07-24
Added SMuFL utility scripts

jwink75:
- Bravura: Use larger noteheads
- Finale Maestro: Use wide noteheads
- Fix arpeggio articulation for non-MakeMusic SMuFL fonts

### DIFF
--- a/dist/hairpin_creator.lua
+++ b/dist/hairpin_creator.lua
@@ -289,654 +289,6 @@ __imports["library.configuration"] = function()
 
 end
 
-__imports["library.client"] = function()
-    --[[
-    $module Client
-
-    Get information about the current client. For the purposes of Finale Lua, the client is
-    the Finale application that's running on someones machine. Therefore, the client has
-    details about the user's setup, such as their Finale version, plugin version, and
-    operating system.
-
-    One of the main uses of using client details is to check its capabilities. As such,
-    the bulk of this library is helper functions to determine what the client supports.
-    All functions to check a client's capabilities should start with `client.supports_`.
-    These functions don't accept any arguments, and should always return a boolean.
-    ]] --
-    local client = {}
-
-    --[[
-    % get_raw_finale_version
-    Returns a raw Finale version from major, minor, and (optional) build parameters. For 32-bit Finale
-    this is the internal major Finale version, not the year.
-
-    @ major (number) Major Finale version
-    @ minor (number) Minor Finale version
-    @ [build] (number) zero if omitted
-
-    : (number)
-    ]]
-    function client.get_raw_finale_version(major, minor, build)
-        local retval = bit32.bor(bit32.lshift(math.floor(major), 24), bit32.lshift(math.floor(minor), 20))
-        if build then
-            retval = bit32.bor(retval, math.floor(build))
-        end
-        return retval
-    end
-
-    --[[
-    % supports_smufl_fonts()
-
-    Returns true if the current client supports SMuFL fonts.
-
-    : (boolean)
-    ]]
-    function client.supports_smufl_fonts()
-        return finenv.RawFinaleVersion >= client.get_raw_finale_version(27, 1)
-    end
-
-    --[[
-    % supports_category_save_with_new_type()
-
-    Returns true if the current client supports FCCategory::SaveWithNewType().
-
-    : (boolean)
-    ]]
-    function client.supports_category_save_with_new_type()
-        return finenv.StringVersion >= "0.58"
-    end
-
-    --[[
-    % supports_finenv_query_invoked_modifier_keys()
-
-    Returns true if the current client supports finenv.QueryInvokedModifierKeys().
-
-    : (boolean)
-    ]]
-    function client.supports_finenv_query_invoked_modifier_keys()
-        return finenv.IsRGPLua and finenv.QueryInvokedModifierKeys
-    end
-
-    --[[
-    % supports_retained_state()
-
-    Returns true if the current client supports retaining state between runs.
-
-    : (boolean)
-    ]]
-    function client.supports_retained_state()
-        return finenv.IsRGPLua and finenv.RetainLuaState ~= nil
-    end
-
-    --[[
-    % supports_modeless_dialog()
-
-    Returns true if the current client supports modeless dialogs.
-
-    : (boolean)
-    ]]
-    function client.supports_modeless_dialog()
-        return finenv.IsRGPLua
-    end
-
-    --[[
-    % supports_clef_changes()
-
-    Returns true if the current client supports changing clefs.
-
-    : (boolean)
-    ]]
-    function client.supports_clef_changes()
-        return finenv.IsRGPLua or finenv.StringVersion >= "0.60"
-    end
-
-    --[[
-    % supports_custom_key_signatures()
-
-    Returns true if the current client supports changing clefs.
-
-    : (boolean)
-    ]]
-    function client.supports_custom_key_signatures()
-        local key = finale.FCKeySignature()
-        return finenv.IsRGPLua and key.CalcTotalChromaticSteps
-    end
-
-    return client
-
-end
-
-__imports["library.general_library"] = function()
-    --[[
-    $module Library
-    ]] --
-    local library = {}
-
-    local client = require("library.client")
-
-    --[[
-    % group_overlaps_region
-
-    Returns true if the input staff group overlaps with the input music region, otherwise false.
-
-    @ staff_group (FCGroup)
-    @ region (FCMusicRegion)
-    : (boolean)
-    ]]
-    function library.group_overlaps_region(staff_group, region)
-        if region:IsFullDocumentSpan() then
-            return true
-        end
-        local staff_exists = false
-        local sys_staves = finale.FCSystemStaves()
-        sys_staves:LoadAllForRegion(region)
-        for sys_staff in each(sys_staves) do
-            if staff_group:ContainsStaff(sys_staff:GetStaff()) then
-                staff_exists = true
-                break
-            end
-        end
-        if not staff_exists then
-            return false
-        end
-        if (staff_group.StartMeasure > region.EndMeasure) or (staff_group.EndMeasure < region.StartMeasure) then
-            return false
-        end
-        return true
-    end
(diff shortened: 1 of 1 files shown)


src/smufl_arpeggio_fix.lua
```diff
@@ -0,0 +1,35 @@
+function plugindef()
+  finaleplugin.RequireSelection = false
```
Nick-Mazuk, line 2: FYI, if `finaleplugin.RequireSelection` is `false`, you don't need this line of code.

src/smufl_bravura_larger_noteheads.lua
```diff
@@ -0,0 +1,46 @@
...
+    finaleplugin.Copyright = "2022"
+    finaleplugin.Version = "1.0"
+    finaleplugin.Date = "4/18/2022"
+    finaleplugin.Notes = [[
+Replaces the noteheads in a 'Bravura' document with the larger glyphs included in the Stylistic Alternates category. These alternate glyphs (which are about 8% larger than the ones loaded by default) are the ones that the font is actually designed to use, and are the ones used by Dorico. For a discussion about these larger noteheads and why they are not used by default, see this thread in the MuseScore forum: https://musescore.org/en/node/68461
+    ]]
+    return "Bravura: Large Noteheads", "Bravura: Large Noteheads", "Replaces noteheads in Bravura with the intended larger size."
+end
+
+--JETSTREAM-- should add to library!
```
Nick-Mazuk, line 14: What is this comment for?

If you think it should be added to the library, you can just add it to the library.

src/smufl_maestro_wide_noteheads.lua
```diff
@@ -0,0 +1,45 @@
...
+    finaleplugin.Version = "1.0"
+    finaleplugin.Date = "7/24/2022"
+    finaleplugin.Notes = [[
+Replaces the noteheads in a 'Finale Maestro' document with the 'Maestro Wide' glyphs included in the Alternates & Extras category.
+    ]]
+    return "Finale Maestro: Wide Notes", "Finale Maestro: Wide Notes", "Replaces noteheads in Finale Maestro with the wide variation."
+end
+
+--JETSTREAM-- should add to library!
+function getUsedFontName(standard_name)
```
Nick-Mazuk, line 15: …especially because it's also used here

jwink75:
Ha! You know that “adding it to the library” is not so simple! It needs to be thought out where it should go, documentation needs to be created, conventions need to be adhered to, etc etc. Think of that as a placeholder note to self, especially as it doesn’t currently impact whether or not it works!

src/library/general_library.lua
```diff
@@ -526,5 +526,37 @@ function library.calc_script_name(include_extension)
...
+        font_name = standard_name
+    elseif string.find(os.tmpname(), "\\") then
+        font_name = string.gsub(standard_name, "%s", "")
+    end
+    return font_name
+end
+
+--[[
+% get_default_music_font
+
```
rpatters1, line 549: I would prefer this function be called `calc_default_music_font_name`. That makes it clear that what is returned is a string and not an `FCFontInfo` instance.  Also, see my question at `get_used_font_name`. We may need to add a prefix like `os`, e.g., `calc_default_music_font_os_name`. Note the use of the `calc` prefix. We should be migrating towards using the `get` prefix for getting properties while `calc` is used for calculating values that are not properties.

src/library/general_library.lua
```diff
@@ -526,5 +526,37 @@ function library.calc_script_name(include_extension)
     return retval
 end
 
+--[[
+% get_used_font_name
+
```
rpatters1, line 531: I am unclear why this function is needed. Can you give an example of an issue it addresses? Also, I think I a clearer name would be `calc_font_os_name` and I think it should take `FCFontInfo` as its input parameter.

src/library/general_library.lua
```diff
@@ -526,5 +526,37 @@ function library.calc_script_name(include_extension)
...
+% get_default_music_font
+
+Fetches the default music font from document options and processes the name into a usable format.
+
+: (string) The name of the defalt music font.
+]]
+function get_default_music_font()
+    local fontinfo = finale.FCFontInfo()
+    if fontinfo:LoadFontPrefs(finale.FONTPREF_MUSIC) then
+        return getUsedFontName(fontinfo:GetName())
```
rpatters1, line 557: The documentation for `GetName` specifically disrecommends its use, because it is not Unicode. Use `GetNameString` instead (and move the call to `calc_font_os_name`.)

rpatters1 (changes requested):

src/library/general_library.lua
jwink75: Great, will do!

src/library/general_library.lua
jwink75: This is code that I am abstracting from JetStream that CJ wrote initially. In looking at it, it seems to me like he is calling up a temporary file and analyzing its path to see whether it uses / or \ as a delimiter to determine OS... Does that seem right? Wouldn't it be better to use UI:IsMac() or whatever it is to determine the OS?

In any case, you may be right and this particular step may be unnecessary. I just confirmed that if I set a document to Engraver Font Set on my Mac, fontinfo:GetName() returns "Engraver Font Set", and then when I open the same file in Windows I get "EngraverFontSet". In other words, the transformation that this function does where it strips out spaces in the font name isn't really needed.

I think that I may test this on my JetStream setup without this step just because I'm a little reluctant to change it when it's already working... But if we're right I'll remove this entire function and just go with calc_default_music_font_name() as a single function.

rpatters1: If it's just converting from FCString to lua string, I'd say the prefix can stay `get_`.

FWIW: believe it or not, Windows works equally well with the `/` (forward slash). I only learned this within the last 12 months, despite having worked with it for almost 40 years going back to when it was only DOS. All this code to use backslash on Windows and forward slash on Unix/macOS is completely unnecessary.

rpatters1: P.S. Also this would ideally be a `FCMFontInfo` mixin, but I'd say we can save that for another day.

src/library/general_library.lua
```diff
@@ -526,5 +526,37 @@ function library.calc_script_name(include_extension)
...
 
+--[[
+% get_used_font_name
+
+Processes an input font name into Mac or Windows specific versions. This function is used by get_defualt_music_font().
+
+@ [standard_name] (string) The name of the font as returned by FCFontInfo:GetName().
+: (string) The OS-specific name of the font.
+]]
+function get_used_font_name(standard_name)
```
Nick-Mazuk, line 537: This needs to have `library.` prefixed to the name for it to be exported.

```lua
function library.get_used_font_name(standard_name)
```

src/library/general_library.lua
```diff
@@ -526,5 +526,37 @@ function library.calc_script_name(include_extension)
...
+end
+
+--[[
+% get_default_music_font
+
+Fetches the default music font from document options and processes the name into a usable format.
+
+: (string) The name of the defalt music font.
+]]
+function get_default_music_font()
```
Nick-Mazuk, line 554: Same here:

```lua
function library.get_default_music_font()
```

src/smufl_bravura_larger_noteheads.lua
```diff
@@ -0,0 +1,29 @@
...
+    finaleplugin.Version = "1.0"
+    finaleplugin.Date = "4/18/2022"
+    finaleplugin.Notes = [[
+Replaces the noteheads in a 'Bravura' document with the larger glyphs included in the Stylistic Alternates category. These alternate glyphs (which are about 8% larger than the ones loaded by default) are the ones that the font is actually designed to use, and are the ones used by Dorico. For a discussion about these larger noteheads and why they are not used by default, see this thread in the MuseScore forum: https://musescore.org/en/node/68461
+    ]]
+    return "Bravura: Large Noteheads", "Bravura: Large Noteheads", "Replaces noteheads in Bravura with the intended larger size."
+end
+
+function bravura_notehead_update()
+    local library = require("library.general_library")
```
Nick-Mazuk, line 15: We typically add these library imports to the top-level of the file.

See https://github.com/finale-lua/lua-scripts/blob/master/src/layers_swap_1_2.lua

src/smufl_maestro_wide_noteheads.lua
```diff
@@ -0,0 +1,28 @@
...
+    finaleplugin.Version = "1.0"
+    finaleplugin.Date = "7/24/2022"
+    finaleplugin.Notes = [[
+Replaces the noteheads in a 'Finale Maestro' document with the 'Maestro Wide' glyphs included in the Alternates & Extras category.
+    ]]
+    return "Finale Maestro: Wide Notes", "Finale Maestro: Wide Notes", "Replaces noteheads in Finale Maestro with the wide variation."
+end
+
+function maestro_wide_update()
+    local library = require("library.general_library")
```
Nick-Mazuk, line 15: Same here. This `require` should be top-level.

Nick-Mazuk (changes requested):

jwink75:
Gentlemen, I have made all the suggested fixes, but before I update the pull request, can we talk names for a moment?

Right now I have these named according to the font they affect. For example, the script that will affect Bravura documents appears in the menu as “Bravura: Large Noteheads”, and the Finale Maestro as “Finale Maestro: Wide Noteheads”. I’m not sure that’s the best approach, and would appreciate your thoughts and wisdom!

Nick-Mazuk:
My honest opinion is that the names are good as is. Though if you think they should be improved we should continue the discussion on this issue about naming scripts. Then, we can go ahead and merge these changes while figuring out a better long-term strategy for naming scripts in this repo.

https://github.com/finale-lua/lua-scripts/issues/270

Nick-Mazuk:
Also, yesterday I merged PR https://github.com/finale-lua/lua-scripts/pull/199 that updated the `library/general_library.lua` file. So now it looks like there are merge conflicts.

To fix this, you can run the following:

```bash
git pull --rebase upstream master
```

or use the "Resolve conflicts" button in this PR.

Screen Shot 2022-07-29 at 9 26 48 AM

PR #272, opened 2022-07-29
Generate mixin documentation

Nick-Mazuk:
The mixin files themselves have very good documentation comments. For instance, take a look at FCMCtrlEdit.lua.

https://github.com/finale-lua/lua-scripts/blob/master/src/mixin/FCMCtrlEdit.lua

However, these are currently not being parsed or displayed on the website. This PR should hopefully change that.

Since there's no good way for me to test this locally, I'm going to go ahead and merge it immediately. Then I can check if there are any bugs when deploying the website and rollback if needed.

@rpatters1 FYI

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -38,6 +38,11 @@ jobs:
               with:
                   input: './lua/src/library'
                   output: './lua/docs/library'
+            - name: Generate the docs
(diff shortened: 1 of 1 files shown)


Nick-Mazuk:
Looks like everything is working as expected.

https://www.finalelua.com/docs/mixin

PR #275, opened 2022-07-29
Update staff_explode.lua

cv-on-hub:
3 parallel "Explosion" scripts - all being changed to default to NO NOTE SPACING with User Config File and Option Key override. This is the first one.

### DIFF
--- a/src/staff_explode.lua
+++ b/src/staff_explode.lua
@@ -11,16 +11,19 @@ function plugindef()
         It warns if pre-existing music in the destination will be erased. 
         It duplicates all markings from the original and resets the current clef on each destination staff.
 
-        This script doesn't respace the selected music after it completes. 
-        If you want automatic respacing you must create a `configuration` file. 
+        By default this script doesn't respace the selected music after it completes. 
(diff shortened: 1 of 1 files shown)


src/staff_explode.lua
```diff
@@ -72,6 +75,12 @@ function ensure_score_has_enough_staves(slot, note_count)
 end
 
 function staff_explode()
+    if finenv.QueryInvokedModifierKeys and
+    (finenv.QueryInvokedModifierKeys(finale.CMDMODKEY_ALT) or finenv.QueryInvokedModifierKeys(finale.CMDMODKEY_SHIFT))
```
Nick-Mazuk, line 79: Going to approve and merge these scripts, though we might consider moving this into the library if it'll be used often.

For instance, I just added the "client" library to help check what a machine is capable of (https://www.finalelua.com/docs/library/client). Something similar might be good for this like `client.is_user_pressing_modifier_keys()` or similar.

Not that it saves much typing, but if this behavior will be used elsewhere, moving it to a library function will ensure consistency amongst all scripts that use it.

Nick-Mazuk (approved):

PR #282, opened 2022-07-29
Bug fixes for repositioning measure numbers [2nd try]

rpatters1:
Items in this PR:

- Refine repositioning of measure numbers to take into account if the number is centered or left-justified.
- Added function to calculate whether to request parts or score when getting `FCMeasureNumberRegion` properties.


Nick-Mazuk (approved):
Looks good to me.

PR #283, opened 2022-07-29
Parse new finaleplugin fields to script metadata

### LINKED ISSUE #122: Add minimum plugin version to the metadata parser
Scripts can have the `finaleplugin.MinJWLuaVersion` metadata field set in their `plugindef`. https://github.com/Nick-Mazuk/jw-lua-scripts/pull/121 Brought this to my attention, and this field should be displayed on the website where users download scripts under "More details" (or "Hide details" when the details are shown, as in the screenshot). This means this field must be parsed as part of the metadata json for the scripts.

Screen Shot 2022-01-03 at 6 20 26 PM

I'll fix this.

### LINKED ISSUE #258: RFC: Script group names
## The issue

This is in response to #257.

#257 adds a new script to help configure the [hairpin_creator.lua](https://github.com/finale-lua/lua-scripts/blob/master/src/hairpin_creator.lua) script. The configuration script is named "Hairpin creator configuration", which makes sense. However, on the website the original script is named "Hairpin create crescendo".

https://www.finalelua.com/scripts?search=Hairpin%20create%20crescendo

This has a few issues:

1. Someone who reads "Hairpin create crescendo" might not realize it creates other hairpins
2. Someone looking at the website might not realize the two scripts are related

I suggest we figure out a better way of naming scripts with multiple menu items.

## How names are currently determined

Names are currently determined from the return value in the plugindef:

```lua
function plugindef()
    return "Hairpin create crescendo", "Hairpin create crescendo", "Create crescendo spanning the selected region"
end
```

This script's name will be shown as "Hairpin create crescendo" since that string is the first value in the return statement. This made sense when we didn't have additional menu options since the first value in the return statement would be what's displayed in the Finale menu. However, this doesn't make sense for scripts taking advantage of the recently added `finaleplugin.AdditionalMenuOptions` option.

```lua
function plugindef()
    finaleplugin.AdditionalMenuOptions = [[
        Hairpin create diminuendo
        Hairpin create swell
        Hairpin create unswell
    ]]
    finaleplugin.AdditionalDescriptions = [[
        Create diminuendo spanning the selected region
        Create a swell (messa di voce) spanning the selected region
        Create an unswell (inverse messa di voce) spanning the selected region
    ]]
    finaleplugin.AdditionalPrefixes = [[
        hairpin_type = finale.SMARTSHAPE_DIMINUENDO
        hairpin_type = -1 -- "swell"
        hairpin_type = -2 -- "unswell"
    ]]
    return "Hairpin create crescendo", "Hairpin create crescendo", "Create crescendo spanning the selected region"
end
```

Now, the name "Hairpin create crescendo" doesn't accurately represent what this script file accomplishes.

## Proposal

For scripts with multiple menu items, we allow a new item in the plugin def for the script group name.

```lua
function plugindef()
    finaleplugin.ScriptGroupName = "Hairpin creator"
    return "Hairpin create crescendo", "Hairpin create crescendo", "Create crescendo spanning the selected region"
end
```

To prevent abuse, the new `finaleplugin.ScriptGroupName` field will be ignored if there aren't any additional menu options. However, when there are additional menu options, the script group name will be displayed on the website instead of the first value in the return statement.

No changes will be needed to RGP Lua, only to the metadata generator. I hope to tackle this and a few other metadata generator backlog items within the next week or two. Though opening this up for comments and suggestions in the meantime.

cc @cv-on-hub @rpatters1

Nick-Mazuk:
This PR parses extra plugindef fields to be displayed on the website. After this I need to actually go update the site, but it should be fairly easy.

cc @rpatters1, could you make sure the JWLua and Finale versions are in the correct form?

Fix #258
Fix #122


rpatters1:
I'm not sure what you mean about the JW Lua vs RGP Lua versions. There shouldn't be any difference if you are talking about the `plugindef` function.

Nick-Mazuk:
More like is this the correct syntax:

```lua
function plugindef()
    finaleplugin.MinJWLuaVersion = "0.61"
    finaleplugin.MaxJWLuaVersion = "0.61"
    finaleplugin.MinFinaleVersion = "2014.5"
    finaleplugin.MaxFinaleVersion = "2014.5"
    return "Hairpin create crescendo", "Hairpin create crescendo", "Hairpin create crescendo"
end
```

Just been such a long time since I've used those, and you're likely more familiar with them than I am.

rpatters1:
I don't think the versions have to be strings. It may be permitted, though.

Nick-Mazuk:
Ok, I'll update the PR to use numbers instead.

Issue #279, opened 2022-07-29
RFC: Supports feature or exit


Nick-Mazuk:
Yesterday we merged the `client` library (https://www.finalelua.com/docs/library/client).

One thing I realized is that this means we're abstracting away the versions of JW/RGP Lua and Finale that are needed for a script. Overall, I think this is a net good thing as it should improve reliability and stability amongst programs. Developers don't need to remember what versions of JW/RGP Lua and Finale that are required, they just need to ensure the client can do the required things.

However, I could see this making us less likely to define our required software versions in the `plugindef` (which to be fair, most of us aren't doing anyways). This means it's less likely that users can know why things aren't working as expected.

Now in some places we are using the capabilities as a form of progressive enhancement:

```lua
if client.supports_this_feature() then
    do_something()
else
    do_something_else()
end
```

However, other parts of the codebase just check if something is supported:

```lua
if client.supports_this_feature() then
    do_something()
end
-- what happens here if that feature is not supported?
-- nothing will happen and the user won't get any errors
```

For these cases, I suggest we add a new type of functions to the client library: `client.supports_x_or_exit()`. These functions will determine if `x` feature is supported. If it's not, it will exit the program and give the user instructions on how to use the script (e.g., upgrade your version of Lua, use a newer version of Finale, etc.).

Additionally, every "supports" function in the `client` library will have an `or_exit` equivalent.

```lua
function client.supports_modeless_dialog()
    return finenv.IsRGPLua
end

function client.supports_modeless_dialog_or_exit()
    if client.supports_modeless_dialog() then
        return true
    end
    -- show dialog "This script uses modeless dialogs which are not available on JW Lua. Please use RGP Lua instead to use this script."
end
```

From this, we can give great, standardized error messages to users while improving developer experience.

@rpatters1 @cv-on-hub @ThistleSifter

rpatters1:
Is it possible for a library function to exit the script? Otherwise every script will have to exit exit exit. (It could be layers deep.)

Nick-Mazuk:
You can exit by throwing an error:

```
error("message")
```

rpatters1:
In that case I recommend that the library func throw the error rather than displaying a dialog. I'm pretty sure RGP Lua will then show that error in a dialog.

rpatters1:
Anyway, it's worth a test.

Nick-Mazuk:
Throwing the error did bring up a dialog:

```lua
[...]

PR #289, opened 2022-07-30
Update client capabilities api

Nick-Mazuk:
This PR is based on our discussion in #279.

I'm really curious for feedback on this API and the syntax around it, since this will likely become one of our heavily used APIs as RGP Lua keeps getting updated (and Finale every few decades).

Reviews requested from @rpatters1 and @ThistleSifter. I will not merge until I get a LGTM from both.

### DIFF
--- a/src/library/client.lua
+++ b/src/library/client.lua
@@ -8,10 +8,75 @@ operating system.
 
 One of the main uses of using client details is to check its capabilities. As such,
 the bulk of this library is helper functions to determine what the client supports.
-All functions to check a client's capabilities should start with `client.supports_`.
-These functions don't accept any arguments, and should always return a boolean.
 ]] --
 local client = {}
+
+local function to_human_string(feature)
+    return string.gsub(feature, "_", " ")
+end
+
+local function requires_later_plugin_version(feature)
+    if feature then
+        return "This script uses " .. to_human_string(feature) .. "which is only available in a later version of RGP Lua. Please update RGP Lua instead to use this script."
+    end
+    return "This script requires a later version of RGP Lua. Please update RGP Lua instead to use this script."
+end
+
+local function requires_rgp_lua(feature)
+    if feature then
+        return "This script uses " .. to_human_string(feature) .. " which is not available on JW Lua. Please use RGP Lua instead to use this script."
+    end
+    return "This script requires RGP Lua, the successor of JW Lua. Please use RGP Lua instead to use this script."
+end
+
+local function requires_plugin_version(version, feature)
+    if version <= 0.54 then
+        if feature then
+            return "This script uses " .. to_human_string(feature) .. " which requires RGP Lua or JW Lua version " .. version ..
+                       " or later. Please update your plugin to use this script."
+        end
+        return "This script requires RGP Lua or JW Lua version " .. version .. " or later. Please update your plugin to use this script."
+    end
+    if feature then
+        return "This script uses " .. to_human_string(feature) .. " which requires RGP Lua version " .. version .. " or later. Please update your plugin to use this script."
+    end
+    return "This script requires RGP Lua version " .. version .. " or later. Please update your plugin to use this script."
+end
+
+local function requires_finale_version(version, feature)
+    return "This script uses " .. to_human_string(feature) .. ", which is only available on Finale " .. version .. " or later"
+end
+
+local features = {
+    clef_change = {
+        test = finenv.StringVersion >= "0.60",
+        error = requires_plugin_version("0.58", "a clef change"),
+    },
+    ["FCKeySignature::CalcTotalChromaticSteps"] = {
+        test = finenv.IsRGPLua and finale.FCKeySignature.__class.CalcTotalChromaticSteps,
+        error = requires_later_plugin_version("a custom key signature"),
+    },
+    ["FCCategory::SaveWithNewType"] = {
+        test = finenv.StringVersion >= "0.58",
+        error = requires_plugin_version("0.58"),
+    },
+    ["finenv.QueryInvokedModifierKeys"] = {
+        test = finenv.IsRGPLua and finenv.QueryInvokedModifierKeys,
+        error = requires_later_plugin_version(),
+    },
+    ["FCCustomLuaWindow::ShowModeless"] = {
+        test = finenv.IsRGPLua,
+        error = requires_rgp_lua("a modeless dialog")
+    },
+    ["finenv.RetainLuaState"] = {
+        test = finenv.IsRGPLua and finenv.RetainLuaState ~= nil,
+        error = requires_later_plugin_version(),
+    },
+    smufl = {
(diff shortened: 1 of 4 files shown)


src/library/client.lua
```diff
@@ -33,81 +115,45 @@ function client.get_raw_finale_version(major, minor, build)
...
-
-: (boolean)
-]]
-function client.supports_custom_key_signatures()
-    local key = finale.FCKeySignature()
-    return finenv.IsRGPLua and key.CalcTotalChromaticSteps
+function client.assert_supports(feature)
+    if not client.supports(feature) then
+        if features[feature].error then
+            error(features[feature].error, 0)
```
ThistleSifter, line 151: What about having level `2` if in debug mode and level `0` otherwise? If I was writing a plugin and I got this error, I'd want to know exactly where it was coming from.

src/library/client.lua
```diff
@@ -13,6 +13,88 @@ These functions don't accept any arguments, and should always return a boolean.
 ]] --
 local client = {}
 
+function to_human_string(feature)
```
ThistleSifter, line 16: Should this (and the ones below) be a `local` functions?

ThistleSifter (changes requested):
Don't forget to update the doc block at the top of the file to reflect the new structure.

src/library/client.lua
Nick-Mazuk: Ah yes. Thanks for reminding me!

src/library/client.lua
Nick-Mazuk: That makes sense. Do you happen to know how to determine if you're running in debug mode?

ThistleSifter: `finenv.DebugEnabled and 2 or 0`

src/library/client.lua
```diff
@@ -33,81 +113,46 @@ function client.get_raw_finale_version(major, minor, build)
...
-
-Returns true if the current client supports changing clefs.
-
-: (boolean)
-]]
-function client.supports_custom_key_signatures()
-    local key = finale.FCKeySignature()
-    return finenv.IsRGPLua and key.CalcTotalChromaticSteps
+function client.assert_supports(feature)
+    local error_code = finenv.DebugEnabled and 2 or 0
```
ThistleSifter, line 147: Apologies for the nit, but for correctness this should be named `error_level`

Nick-Mazuk: True. In other languages it's an error "code", but it does look like it's "level" in Lua.

https://www.lua.org/manual/5.1/manual.html#pdf-error

Nick-Mazuk: And no worries about the small nits. Things like this matter in the long run.

src/library/client.lua
```diff
@@ -33,81 +113,46 @@ function client.get_raw_finale_version(major, minor, build)
...
+function client.assert_supports(feature)
+    local error_level = finenv.DebugEnabled and 2 or 0
+    if not client.supports(feature) then
+        if features[feature].error then
+            error(features[feature].error, error_level)
+        end
+        -- Generic error message
+        error("Your Finale version does not support " .. to_human_string(feature), error_level)
+    end
+    return true
```
ThistleSifter, line 155: Why does this function need to return anything?

Nick-Mazuk: That way the two functions can be a drop-in replacement for one another. You can change between `client.supports` and `client.assert_supports` without needing to change any code for the success case.

Otherwise you may get the following unintuitive behavior:

```lua
-- original
if client.supports("x") then
    -- do something here
end


-- refactored
if client.assert_supports("x") then
    -- this never runs for some reason…
end
```

Now is the refactored option the kind of thing we'd want to save in this repo? Probably not. However I can imagine there being some unintuitive bugs as you're trying to work through how best to organize the code.

rpatters1 (approved):
Looks good to me.

src/library/client.lua
Nick-Mazuk: Is this ok with you? If so, I'd like to merge unless you have any other nits.

rpatters1:
I found one issue that's already merged. The problem is that `FCKeySignature` cannot be constructed in Lua. So you should rewrite the custom key sig check like this:

```lua
function client.supports_custom_key_signatures()
    -- FCKeySignature cannot be constructed in Lua, so check __class table instead
    return finenv.IsRGPLua and finale.FCKeySignature.__class.CalcTotalChromaticSteps
end
```

Nick-Mazuk:
Done.

Nick-Mazuk:
Small update:

I realized we don't need to define the `test` as a function for every item since it's always going to return the same thing. Doing so simply adds unnecessary lines of code and increases the runtime cost.

If in the future we need to do something more complex to test a capability, we can use an immediately invoked function, which I in Lua uses the following syntax:

```lua
(function()
    print("hello world")
end)()
```

src/library/client.lua
```diff
@@ -33,81 +98,46 @@ function client.get_raw_finale_version(major, minor, build)
...
-Returns true if the current client supports FCCategory::SaveWithNewType().
+For a list of valid features, see the [`features` table in the codebase](https://github.com/finale-lua/lua-scripts/blob/master/src/library/client.lua#L52).
 
+@ feature (string) The feature the client should support.
 : (boolean)
 ]]
-function client.supports_category_save_with_new_type()
-    return finenv.StringVersion >= "0.58"
+function client.supports(feature)
+    return features[feature].test
```
ThistleSifter, line 114: With the latest change, this will no longer throw an error if the feature doesn't exist, is that intentional?

src/library/client.lua
```diff
@@ -8,11 +8,76 @@ operating system.
...
+local features = {
+    clef_change = {
+        test = finenv.StringVersion >= "0.60",
+        error = requires_plugin_version("0.58", "a clef change"),
+    },
+    custom_key_signature = {
+        test = finenv.IsRGPLua and finale.FCKeySignature.__class.CalcTotalChromaticSteps,
+        error = requires_later_plugin_version("a custom key signature"),
+    },
+    ["FCCategory::SaveWithNewType"] = {
```
ThistleSifter, line 59: I like this. Can the others be changed too where applicable? Eg `modeless_dialog` > `FCCustomLuaWindow::ShowModeless`, `retained_state` > `finenv.RetainLuaState`, etc, etc

It'll be more intuitive, where possible, rather than having to remember/look up what the feature has been called.

src/library/client.lua
ThistleSifter: It's probably not what I would do, but if you prefer it that way, go for it.

src/library/client.lua
Nick-Mazuk: I agree.

src/library/client.lua
Nick-Mazuk: Thanks for catching that. I added a check that'll throw an error if the feature is not defined.

ThistleSifter (approved):
I think it's good to go.

PR #290, opened 2022-07-31
Added Expression scaling script

jwink75:


src/expression_scale_noscale.lua
```diff
@@ -0,0 +1,40 @@
...
+    Expressions: Scaling ON
+    ]]
+    finaleplugin.AdditionalDescriptions = [[
+    Turns on expression scaling in the selected region.
+    ]]
+    finaleplugin.AdditionalPrefixes = [[
+    scale_bool = true
+    ]]
+--    finaleplugin.MinJWLuaVersion = 0.63
+--    finaleplugin.ScriptGroupName = ""
```
Nick-Mazuk, line 21: It would be worth it to add the `finaleplugin.ScriptGroupName` and `finaleplugin.ScriptGroupDescription` to make clearer for users on what the script does on the download page.

https://github.com/finale-lua/lua-scripts/issues/258

These fields directly appear on the website's download page.

Screen Shot 2022-07-30 at 6 05 15 PM

PR #295, opened 2022-07-31
Create smufl_maestro_wide_noteheads

RLJunchaya:
Replaces Finale Maestro noteheads with the former Maestro Wide equivalents found in the Alternates & Extras category.
Based on a script by Jacob Winkler

### DIFF
--- a/src/smufl_maestro_wide_noteheads
+++ b/src/smufl_maestro_wide_noteheads
@@ -1,0 +1,703 @@
+local __imports = {}
+
+function require(item)
+    if __imports[item] then
+        return __imports[item]()
+    else
+        error("module '" .. item .. "' not found")
+    end
+end
+
+__imports["library.client"] = function()
+    --[[
+    $module Client
+
+    Get information about the current client. For the purposes of Finale Lua, the client is
+    the Finale application that's running on someones machine. Therefore, the client has
+    details about the user's setup, such as their Finale version, plugin version, and
+    operating system.
+
+    One of the main uses of using client details is to check its capabilities. As such,
+    the bulk of this library is helper functions to determine what the client supports.
+    All functions to check a client's capabilities should start with `client.supports_`.
+    These functions don't accept any arguments, and should always return a boolean.
+    ]] --
+    local client = {}
+
+    --[[
+    % get_raw_finale_version
+    Returns a raw Finale version from major, minor, and (optional) build parameters. For 32-bit Finale
+    this is the internal major Finale version, not the year.
+
+    @ major (number) Major Finale version
+    @ minor (number) Minor Finale version
+    @ [build] (number) zero if omitted
+
+    : (number)
+    ]]
+    function client.get_raw_finale_version(major, minor, build)
+        local retval = bit32.bor(bit32.lshift(math.floor(major), 24), bit32.lshift(math.floor(minor), 20))
+        if build then
+            retval = bit32.bor(retval, math.floor(build))
+        end
+        return retval
+    end
+
+    --[[
+    % supports_smufl_fonts()
+
+    Returns true if the current client supports SMuFL fonts.
+
+    : (boolean)
+    ]]
+    function client.supports_smufl_fonts()
+        return finenv.RawFinaleVersion >= client.get_raw_finale_version(27, 1)
+    end
+
+    --[[
+    % supports_category_save_with_new_type()
+
+    Returns true if the current client supports FCCategory::SaveWithNewType().
+
+    : (boolean)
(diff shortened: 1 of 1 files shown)


src/smufl_maestro_wide_noteheads
```diff
@@ -0,0 +1,703 @@
...
+        if fontinfo:LoadFontPrefs(finale.FONTPREF_MUSIC) then
+            fontinfo:GetNameString(default_music_font_name)
+            return default_music_font_name.LuaString
+        end
+    end
+
+    return library
+
+end
+
```
Nick-Mazuk, line 674: Could you delete everything above this line?

In this repository, we have a much of shared code. That way, we can reuse a ton of code across multiple scripts, speeding up the time it takes for us to write scripts and increasing the stability of our scripts (since we don't need to worry about every edge case every time).

You'll notice down below that Jacob imports our general library:

```lua
local library = require("library.general_library")
```

This is one of our files with shared code: https://github.com/finale-lua/lua-scripts/blob/master/src/library/general_library.lua

As you can see, this shared code is split across multiple files.

However, we don't want to make users download multiple files to use our scripts. Instead, we have it setup that when we update a script, all the used shared files are "bundled" into a single file. Everything above this line is the auto-generated code that bundles the general library into the script file. That's what users download for the website.

All that to say you can safely delete the above lines and everything will continue to work.

For reference, here's the original script Jacob wrote:

https://github.com/finale-lua/lua-scripts/blob/master/src/smufl_maestro_wide_noteheads.lua

And here's the bundled version that users download from the website:

https://github.com/finale-lua/lua-scripts/blob/master/dist/smufl_maestro_wide_noteheads.lua

Nick-Mazuk (changes requested):
Hey @RLJunchaya, welcome to the repository!

I made one small comment for something to fix, but overall looks good.

If at any time you have questions or comments, feel free to ask.

Nick-Mazuk:
Upon further reflection, since this is so similar to Jake's original script, it might be better to just incorporate the changes into his script. I'm talking with him offline and will follow-up here.

Nick-Mazuk:
After talking with Jake, this is what he said:

> The "FinaleMaestro" thing probably isn't actually needed!
>
> When we started working on JetStream I think we started running into trouble with the Engraver Font Set font because it will report itself as "EngraverFontSet" on Windows and "Engraver Font Set" on Mac. That's because the different OSs are pulling that info from different fields in the font. CJ compensated for that by writing the "use actual name" function or whatever it was that I initially put into the general_library but then we decided to take out.
>
> Finale Maestro actually reports itself as "Finale Maestro" on both operating systems. When I first made the script, it actually didn't work because CJ's function was stripping out that space, which was actually necessary! I compensated by adding 'or default_music_font == "Finale Maestro"` since that was the condition that actually needed to be met... I left the FinaleMaestro variation just to cover my bases. But again, probably unnecessary.
> 
> I think it is also unnecessary to include a version for non-SMuFL fonts. The reason this script was even written is that Rafael preferred the 'Maestro Wide' font... For non-SMuFL folks, that's the appropriate solution, just change the font! 
>
> The 'Maestro Wide' noteheads are included as stylistic alternates within Finale Maestro, so using this script you get to retain all the other SMuFL goodies while still using those wider notes.

Given that, I'm opting to close this PR. It seems like these changes don't do much and for non-SMuFL users the more sustainable way is to change the font through the Finale UI.

With that said, I may be missing something so feel free to comment. We can always reopen this PR.

I appreciate the effort to get started contributing, Rafael! If you have any more scripts you'd like to contribute now or in the future, feel free to create a new PR. And as always, feel free to ask questions if you need any help.

RLJunchaya:
Hi, Nick, thanks for your email. There was actually a misunderstanding from
my side as Jacob commented he had developed the Bravura script and I
thought I could use it as a model to create one to get the Maestro Wide
noteheads. I downloaded the Bravura script and modified it unaware that
Jacob had made the same thing for Maestro, so I uploaded it unnecessarily
as it was already in the depository. Already talked to Jacob over the
matter and there is no need to have the almost identical thing in the
depository so it is ok to delete it (I tried to do it myself but probably
you have already addressed the situation)
Thank you very much (and also Jacob). I think I will now wipe the dirt of
my programming experience and try to make a more general noteheads changer
script. Let's see if I succeed..

Thanks a lot for all the effort with the depository and your videos.
Hopefully MakeMusic will recognise all those effort more effectively in
future

Rafael J

Nick-Mazuk:
Glad everything got sorted out.

PR #304, opened 2022-08-01
Rgp/bug fixes

rpatters1:
Changes the bariolage script to start with layer 2 (rather than layer 1) when the stem directions warrant it.

### DIFF
--- a/src/note_bariolage.lua
+++ b/src/note_bariolage.lua
@@ -1,8 +1,8 @@
 function plugindef()
     finaleplugin.Author = "Jacob Winkler"
     finaleplugin.Copyright = "2022"
-    finaleplugin.Version = "1.0"
-    finaleplugin.Date = "3/18/2022"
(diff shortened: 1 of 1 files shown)


src/note_bariolage.lua
```diff
@@ -26,29 +28,38 @@ local note_entry = require("library.note_entry")
 function bariolage()
     local region = finenv.Region()
     layer.copy(region, 1, 2)
-    local layer1_ct = 1
-    local layer2_ct = 1
+    local oddlayer_ct = 1
+    local evenlayer_ct = 1
```
Nick-Mazuk, line 32: Nit: these should be `odd_layer_ct` and `even_layer_ct`

Nick-Mazuk (changes requested):
Could you also update the revision date and number?

After these two things, LGTM.

Nick-Mazuk (approved):
LGTM.

PR #306, opened 2022-08-01
Update notes_tie.lua

cv-on-hub:
Fixing more plugindef() hiccups.

### DIFF
--- a/src/notes_tie.lua
+++ b/src/notes_tie.lua
@@ -2,18 +2,28 @@ function plugindef()
     finaleplugin.RequireSelection = true
     finaleplugin.Author = "Carl Vine and Robert Patterson"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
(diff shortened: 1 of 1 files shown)


cv-on-hub:
The only weird thing (now!) is I keep getting messages like the attached after each submission ..
unhappy_github

Nick-Mazuk:
Oh, hmm…

It seems like the GitHub actions to build and deploy everything to the site are also running on your fork. Because of that, they're breaking (which is a good thing). However, they shouldn't be running on your fork.

Things seem to be running fine in the main repository, so it's not an issue as far as breaking the actual site.

Since it may be a few weeks since I'm able to really debug this, perhaps you can turn off notifications on your fork file an issue here to keep track of this?

Either way, this PR looks good to me.

rpatters1:
I get those too. You can ignore them on your fork.

cv-on-hub:
Great news all. I'll turn off the warnings.

PR #308, opened 2022-08-02
Update layer_clear_selective.lua

cv-on-hub:
All of this hoo-hah so that the MODAL dialog retains its screen position in RGP 0.64+. Still stable on earlier RGP versions, and not using mixin until it is fully stabilised.

### DIFF
--- a/src/layer_clear_selective.lua
+++ b/src/layer_clear_selective.lua
@@ -3,58 +3,71 @@ function plugindef()
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "http://carlvine.com/?cv=lua"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "v1.04"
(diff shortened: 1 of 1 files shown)


rpatters1:
If you want it to retain position always, use `RegisterHandleCloseWindow` instead of `RegisterHandleOkButtonPressed`. Right now it only remembers the window position if the user presses `OK`.

cv-on-hub:
Good to know, but I was thinking specifically that a "cancel" would NOT retain window position.

rpatters1:
Not how I would do it, but it's your baby.

rpatters1:
`RegisterHandleCloseWindow` also handles the case of it being closed with the close button (upper left small button).

cv-on-hub:
OK you've convinced me!

Nick-Mazuk:
I'll let Robert review and approve this one as he's much more familiar with the dialogs.

Nick-Mazuk (approved):
Looks good to me.

PR #309, opened 2022-08-02
fix a couple bugs in client.lua

rpatters1:
- `requires_plugin_version` sometimes receives `version` as a string.
- `features` table needs `get_raw_finale_version` to be defined before it is.

### DIFF
--- a/src/library/client.lua
+++ b/src/library/client.lua
@@ -30,7 +30,7 @@ local function requires_rgp_lua(feature)
 end
 
 local function requires_plugin_version(version, feature)
-    if version <= 0.54 then
+    if tonumber(version) <= 0.54 then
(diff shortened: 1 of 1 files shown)


Nick-Mazuk (changes requested):
Overall looks good. Could you also update the line number of the `features` table in the doc comments URL down below? It appears twice:

https://github.com/finale-lua/lua-scripts/blob/master/src/library/client.lua#L52

to 

https://github.com/finale-lua/lua-scripts/blob/master/src/library/client.lua#L69

Unfortunately GitHub won't let me make a comment directly on any lines you didn't edit.

Nick-Mazuk (approved):
Good catch. Looks good to me.

PR #310, opened 2022-08-02
Update layer_hide.lua

cv-on-hub:
Added save-position modal dialogs with RetainLuaState, ready for RGP 0.64.


rpatters1 (approved):
Looks good to me. Once 0.64 is released we should create a RunModal in mixin to do this. But for now this is fine.

PR #311, opened 2022-08-02
Update layer_mute.lua

cv-on-hub:
Added save-position modal dialogs with RetainLuaState, ready for RGP 0.64.


rpatters1 (approved):
Lgtm. Same comment about mixin as the other PR.

PR #312, opened 2022-08-02
Update note_ends.lua

cv-on-hub:
Fixed an error condition (with quarter notes) where one note is followed by more than one rest.

### DIFF
--- a/src/note_ends.lua
+++ b/src/note_ends.lua
@@ -60,6 +60,7 @@ function expand_note_ends()
             and entry:Next():IsRest() -- that is a rest
             and entry.Duration < note_value -- this note is too short
             and note_boundary == 0 -- on a beat or half-beat
+            and entry.Duration + entry:Next().Duration >= note_value
             and
             (    (beat_duration >= position_in_beat + note_value) -- enough space for expanded endpoint
              or  (not is_compound_meter and not eighth_notes and start_beat % 2 == 0) -- special case quarter note on an odd beat
--- a/src/note_ends.lua
+++ b/src/note_ends.lua
@@ -3,13 +3,13 @@ function plugindef()
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "http://carlvine.com/?cv=lua"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "v0.54"
+    finaleplugin.Version = "v0.55"
     finaleplugin.Date = "2022/08/02"
     finaleplugin.AdditionalMenuOptions = [[
-        Note ends eighths
+        Note Ends Eighths
     ]]
     finaleplugin.AdditionalUndoText = [[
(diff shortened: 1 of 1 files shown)


src/note_ends.lua
```diff
@@ -60,6 +60,7 @@ function expand_note_ends()
             and entry:Next():IsRest() -- that is a rest
             and entry.Duration < note_value -- this note is too short
             and note_boundary == 0 -- on a beat or half-beat
+            and entry.Duration + entry:Next().Duration >= note_value
             and
             (    (beat_duration >= position_in_beat + note_value) -- enough space for expanded endpoint
```
rpatters1, line 65: This is a style issue, but it is important. You should *never* end up with two `end` statements vertically aligned. It is a recipe for a long frustrating slog sometime in the future. I would tab the entire `then`...`end` block over one to the right.

rpatters1 (changes requested):

src/note_ends.lua
```diff
@@ -63,19 +63,19 @@ function expand_note_ends()
...
+                then
+                local duration_with_rest = entry.Duration + entry:Next().Duration
+                entry.Duration = note_value	-- expand target note
+                if duration_with_rest == note_value then
+                    should_delete_next = true -- just delete the following rest
+                elseif duration_with_rest > note_value then -- some duration left over
+                    entry:Next().Duration = duration_with_rest - note_value -- make rest smaller
+                    should_delete_next = false -- and don't delete it
+                end
         end
```
rpatters1, line 77: So now you went one tab too far. The final `end` statements should be a procession of tab stops:

```lua
                end
            end
        end
    end
end
```

No gaps and no vertical alignement or overlaps.

rpatters1 (changes requested):

src/note_ends.lua
cv-on-hub: My last submission PR# acb1acb now looks balanced on GitHub.

rpatters1: I'm sorry to be a pain, but this is important. The `then` and `end` are now aligned correctly, but the lines of code *between* them need an indent.

rpatters1 (changes requested):

rpatters1:
This still has a gap in the `end` statements at the bottom. I had to think about it way too hard before I understood it. Perhaps the cleanest approach would be to create a function that does all those checks and then call that function.

```lua
local valid_note = function(entry)
    if not entry:IsNote() then
        return false
    end
    if not entry:Next() then
        return false
    end
    if not entry:Next():IsRest() then
        return false
    end
    .
    .
    .
    return true
end
```

Nick-Mazuk:
Nit on Robert's suggestion, the function name should be `is_valid_entry`. The `is_` to signify that it returns a boolean, and `entry` instead of `note` because we are validating an entry which may contain several notes.

cv-on-hub:
All taken on board. New version follows.

rpatters1 (approved):
Looks good to me

PR #314, opened 2022-08-03
Update rest_offsets.lua

cv-on-hub:
Re-jigged for saved-position modal dialog (RGP 0.64) with *NEW IMPROVED* vertical rest positioning provided by @rpatters1

### DIFF
--- a/src/rest_offsets.lua
+++ b/src/rest_offsets.lua
@@ -2,7 +2,7 @@ function plugindef()
     finaleplugin.RequireSelection = true
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
     finaleplugin.AuthorURL = "http://carlvine.com/lua/"
-    finaleplugin.Version = "v1.33"
+    finaleplugin.Version = "v1.35"
     finaleplugin.Date = "2022/08/03"
     finaleplugin.Notes = [[
     Several situations including cross-staff notation (rests should be centred between the staves) 
@@ -55,9 +55,10 @@ function user_choices()
     end
 
     texts = { -- offset number / horizontal offset / description /  vertical position
-        { "0", 5, "= top staff line", 0},
-        { "-4", 0, "= middle staff line", 15 },
-        { "-8", 0, "= bottom staff line", 30 },
+        {  "4", 5, "= top staff line", 0},
+        {  "0", 5, "= middle staff line", 15 },
+        { "-4", 0, "= bottom staff line", 30 },
+        { "", 0, "(for 5-line staff)", 45 },
     }
     for _, v in ipairs(texts) do -- static text information lines
         str.LuaString = v[1]
@@ -83,21 +84,17 @@ function user_choices()
 end
 
 function make_the_change()
+    local staff_spec = finale.FCCurrentStaffSpec()
     if finenv.RetainLuaState ~= nil then
         finenv.RetainLuaState = true
     end
-    local current_staff = nil
-    local staff_spec = finale.FCCurrentStaffSpec()
 
     for entry in eachentrysaved(finenv.Region(), config.layer) do
         if entry:IsRest() then
-            --if config.offset == 0 then
-            --    entry:SetFloatingRest(true)
-            --else
-                if current_staff ~= entry.staff then -- need a new staff spec
-                    current_staff = entry.staff
-                    staff_spec:LoadForEntry(entry)
-                end
+            if config.offset == 0 then
+                entry:SetFloatingRest(true)
+            else
+                staff_spec:LoadForEntry(entry)
                 local rest_prop = "OtherRestPosition"
                 local duration = entry.Duration
                 if duration >= finale.BREVE then
@@ -112,7 +109,7 @@ function make_the_change()
                 local rest = entry:GetItemAt(0)
                 local curr_staffpos = rest:CalcStaffPosition()
                 entry:SetRestDisplacement(entry:GetRestDisplacement() + total_offset - curr_staffpos)
-            --end
+            end
         end
 	end
 end
--- a/src/rest_offsets.lua
+++ b/src/rest_offsets.lua
@@ -2,7 +2,7 @@ function plugindef()
     finaleplugin.RequireSelection = true
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
     finaleplugin.AuthorURL = "http://carlvine.com/lua/"
-    finaleplugin.Version = "v1.31"
+    finaleplugin.Version = "v1.32"
     finaleplugin.Date = "2022/08/03"
     finaleplugin.Notes = [[
     Several situations including cross-staff notation (rests should be centred between the staves) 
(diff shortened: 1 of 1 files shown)


rpatters1:
I have a suggestion about your pull requests which is to combine them into a single request. If you are "re-jigging" several scripts to use a new feature, it is much easier to review them all at once in a single PR. No need this time, but it will cut down a lot on administrative overhead.

src/rest_offsets.lua
```diff
@@ -64,34 +71,51 @@ function get_user_choices()
...
-            entry:SetRestDisplacement(rest_offset + base_offset)
+    for entry in eachentrysaved(finenv.Region(), config.layer) do
+         if entry:IsRest() then
+            if config.offset == 0 then
+                entry:SetFloatingRest(true)
+            else
+                entry:MakeMovableRest()
+                local rest = entry:GetItemAt(0)
+                local curr_staffpos = rest:CalcStaffPosition()
+                entry:SetRestDisplacement(entry:GetRestDisplacement() + config.offset - curr_staffpos - 4)
```
rpatters1, line 98: Instead of hard-coding `-4`, you should be using the correct rest position for the rest from `FCCurrentStaffSpec`. The properties are `DoubleWholeRestPosition`, `WholeRestPosition`, `HalfRestPosition`, and `OtherRestPosition`.

Also, it might be nice to add this to the `note_entry` library as a callable function.

rpatters1 (changes requested):

src/rest_offsets.lua
cv-on-hub: > might be nice to add this to the `note_entry` library as a callable function

True. I'll work this up in due course to match rest duration to StaffSpec. Not confident enough to start editing a library file! For now this script works miles better than the old version currently in the repo.

src/rest_offsets.lua
```diff
@@ -87,23 +86,40 @@ function make_the_change()
...
+                if duration % 3 == 0 then -- it's dotted
+                    duration = duration * 2 / 3 -- get the un-dotted version
+                end
+                local power_count = 1
+                while duration > 1024 and power_count <= #rest_type do
+                    duration = duration / 2
+                    power_count = power_count + 1
+                end
+                local rest_type_offset = staff_spec[rest_type[power_count]] + 4 -- adjusted for middle line
                 entry:MakeMovableRest()
```
rpatters1, line 112: Generally speaking, a good rule of thumb is to avoid hard-coding numbers other than 0. It is impossible to make that a hard-and-fast rule, but you can get close. Most of these hard-coded numbers could have been avoided with finale constants. It would be much clearer to write:

```lua
if duration >= finale.DOUBLE_WHOLE then
   rest_prop = "DoubleWholeRestPosition"
elseif duration >= finale.WHOLE_NOTE then
   rest_prop = "WholeRestPosition"
elseif duration >= finale.HALF_NOTE then
   rest_prop = "HalfRestPosition"
else
   rest_prop = "OtherRestPosition",
end

local rest_type_offset = staff_spec[rest_prop] + 4 -- WHY 4??!! (see below)
```

That's it. No need to check dotted (unless you need it for something else). And no hard-coded numbers.

The one that really bugs me is line 111. What is that magic 4? The `staff_spec` for  standard staves has a rest position of -4 for all rests. So in the most common case  `rest_type_offset` will end up zero. Why are you adding 4? Have you tested this code?

rpatters1 (changes requested):

src/rest_offsets.lua
Nick-Mazuk: https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants

cv-on-hub: Will amend with chained elseif's. Got you both on the un-magic of fixed constants. I decided (arbitrarily) to call the middle staff position zero (input by user) in this script, since that's where rests "appear" to sit. `rest_type_offset` is subtracted from the final offset so the plus four moves it down the staff. Yes - its was tested and works with all basic staves and rest durations. Should I leave zero to represent the top staff line?

rpatters1:
There are more staff properties that tell you where the reference line is in relation to the top and bottom staves. Recall that Finale allows staves with any number of lines from 0 on up, and the reference line can be any of them. Heck, there can be more than 2 steps between each line (though it has to be an even number). Perhaps the most use case for these weird staff configurations in percussion instruments.

If you want to go from the middle line, look at the staff properties to get the reference line and the top and bottom lines. Then calculate the middle position and then offset that from the reference line. (The reference line is always zero.)

(This rigamarole is why we don't hard-code.)

cv-on-hub:
Do you think this new version (following) is more sensible? “0” = top staff; “-8” = bottom staff; “-4” = middle (“normal” position on 5-line staff).

I note that the built-in version (Note, Beam, and Rest Editing -> Move Rests...) only changes relatively - it doesn’t offer an absolute position except “Clear”. Although the “-4” constitutes an explicit constant, it is also the number of half-spaces between the top and centre lines of the staff. Another option is to count the number of lines in the staff, subtract one, divide by two, multiply by number of half-spaces to get the centre-point - i.e. -4 for a 5-line staff!

src/rest_offsets.lua
```diff
@@ -88,34 +88,33 @@ function make_the_change()
...
+                elseif duration >= finale.HALF_NOTE then
+                    rest_prop = "HalfRestPosition"
                 end
-                local power_count = 1
-                while duration > 1024 and power_count <= #rest_type do
-                    duration = duration / 2
-                    power_count = power_count + 1
-                end
-                local rest_type_offset = staff_spec[rest_type[power_count]] + 4 -- adjusted for middle line
+                local total_offset = staff_spec[rest_prop] + config.offset
```
rpatters1, line 110: This computation is now correct. Basically, the `...RestPosition` properties are *telling* you the middle line. So by adding it in your are offsetting from the middle line. So your text above should remain {-4, 0, 4}. I just tested your code, so I know I'm right. I'm concerned that you haven't seen this in your own testing.

Also, another issue I noticed is that you must reload `FCCurrentStaffSpec` for every entry. Don't try to be smart about it. This is because staff styles can change the shape of the staff in the middle of a measure. That's the whole reason `FCCurrentStaffSpec` exists instead of plain old `FCStaff`. It is very efficient to load, so just do it.

rpatters1 (changes requested):

cv-on-hub:
Yes - I'm also now getting correct behaviour at {-4, 0, 4}. I must have somehow been exercising the wrong version of the script. Not to worry. Seems now to be working properly!

cv-on-hub:
PS: thanks for all your patience and above-and-beyond help.

rpatters1:
No problem. I will await those final 2 fixes, and hopefully it is finished.

cv-on-hub:
My last commit (8b6b004) actually included those changes. But I re-commit now with corrected menu capitalisation.

rpatters1:
On Github, the `end` statement on line 114 is vertically aligned with the one on line 113. It needs to be one tabstop to the left.

Also, I recommend moving line 87

```lua
local staff_spec = finale.FCCurrentStaffSpec()
```
 immediately before line 106 so that it looks like this:

```lua
local staff_spec = finale.FCCurrentStaffSpec()
staff_spec:LoadForEntry(entry)
```

The reason is that you are reusing the same instance of `FCCurrentStaffSpec` for every entry, whereas if you do the way I'm suggesting you'll get a freshly initialized one. It may work as-is, but there's a risk that it won't and that you'll pick up crud from a previous iteration.

cv-on-hub:
Thanks Robert. I think this nails it.

rpatters1 (approved):

PR #315, opened 2022-08-04
Rgp/bug fixes

rpatters1:
More bugs found due to real-world utilization of scripts for actual work.

- Prevent standalone hairpin script from crashing if smart shapes fail to load
- Always use score when generating a text file tracking document
-


Nick-Mazuk (approved):
Looks good to me.

PR #317, opened 2022-08-04
Revision of 2 MIDI data manipulation scripts

cv-on-hub:
Revision of 2 scripts using my new template for saved-position modal dialogs (RGP 0.64+). Various other minor improvements.

### DIFF
--- a/src/midi_duration.lua
+++ b/src/midi_duration.lua
@@ -79,7 +79,7 @@ function make_the_change()
     end
     for entry in eachentrysaved(finenv.Region(), config.layer) do
         local perf_mod = finale.FCPerformanceMod()
-		if entry:IsNote() then
+        if entry:IsNote() then
     		perf_mod:SetNoteEntry(entry)
     		for note in each(entry) do
     		    perf_mod:LoadAt(note)     -- don't change durations of tied notes!
--- a/src/midi_duration.lua
+++ b/src/midi_duration.lua
@@ -1,44 +1,51 @@
 function plugindef()
     finaleplugin.RequireSelection = true
     finaleplugin.Author = "Carl Vine"
-    finaleplugin.AuthorURL = "http://carlvine.com"
+    finaleplugin.AuthorURL = "http://carlvine.com/lua/"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "v1.3"
-    finaleplugin.Date = "2022/06/13"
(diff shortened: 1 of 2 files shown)


Nick-Mazuk:
> This is more a question for @Nick-Mazuk. I'm starting to see over and over this pattern:
> 
> ```lua
> if finenv.RetainLuaState ~= nil then
>     finenv.RetainLuaState = true
> end
> ```
> 
> I am wondering if this should be in `client.lua` if it isn't already. A counter argument could be that much of this code could disappear into `mixin` once RGP Lua 0.64 is released. (And that would hopefully include these lines, but it isn't guaranteed I suppose.)

At the very least, this should be:

```lua
if client.supports("finenv.RetainLuaState") then
    finenv.RetainLuaState = true
end
```

However, it seems like just a simple assignment should work, since I don't see how it'll cause runtime errors.

```lua
finenv.RetainLuaState = true
```

If we're really worried about this potentially throwing errors, I'd actually say the best pattern overall would be to encapsulate the if statement into its own library function. This probably won't fit into the `client` library as I see that more as a library to get information about the current client.

```lua
-- library function
function library.retain_state()
    if client.supports("finenv.RetainLuaState") then
        finenv.RetainLuaState = true
    end
end

--  script file

library.retain_state()
```

rpatters1:
Because `finenv` is userdata rather than a table, you can just assign to `RetainLuaState` if it isn't there. Or even if `finenv` is technically a table, its `_metatable` is controlled by LuaBridge which doesn't like it. So we'll need a check. I would propose `library.retain_state(bool)` because we need to reset as well as set it.

Nick-Mazuk:
I see, perhaps default the input to true?

```lua
-- library function
function library.retain_state(enabled)
    enabled = enabled or true
    if client.supports("finenv.RetainLuaState") then
        finenv.RetainLuaState = enabled
    end
end

--  script file

library.retain_state() -- turns it on
library.retain_state(true) -- turns it on
library.retain_state(false) -- turns it off
```

cv-on-hub:
> the same visceral revulsion it arouses in me.

Not quite revulsion but certainly displeasure. These must be ancient artefacts of using a much dumber text editor in original scripts, though (of course) they looked fine in Visual Studio Code before I submitted them. Had also checked (thought) in the Hub, but missed that last subroutine in both cases.

cv-on-hub:
So this is annoying. Yet one more tab mismatch from VSCode (`midi_duration`)

rpatters1 (approved):
lgtm

PR #318, opened 2022-08-04
Rgp/support bar crossing for hairpins

rpatters1:
This is an enhancement to `standalone_hairpin_adjustment.lua`. The motivation for the change is that I frequently (for example) have a cresc. hairpin from pp to ff where the ff occurs on the downbeat of the following bar. If I extend the hairpin to that downbeat, then if that next bar occurs on a new system, I end up with an ugly and stubby hairpin continuation. I prefer not to have these, so I usually set my hairpins to the end of the preceding bar in these situations.

This enhancement extends hairpins that end at the end of a bar to a dynamic at the beginning of the next bar if they are on the same systems and otherwise behaves as now if they are on different systems. It is triggered by an option that I have added to `config`. Both bars have to be selected before it does this.

Besides the usual code review, I would like to ask whether this new behavior should be the default. Right now I have it set not to engage, so it has to be activated with a config file.

### DIFF
--- a/src/standalone_hairpin_adjustment.lua
+++ b/src/standalone_hairpin_adjustment.lua
@@ -47,9 +47,6 @@ end
 local staff_systems = finale.FCStaffSystems()
 staff_systems:LoadAll()
 
-local distance_prefs = finale.FCDistancePrefs()
-distance_prefs:Load(1)
-
 function calc_cell_relative_vertical_position(fccell, page_offset)
     local relative_position = page_offset
     local cell_metrics = fccell:CreateCellMetrics()
--- a/src/standalone_hairpin_adjustment.lua
+++ b/src/standalone_hairpin_adjustment.lua
@@ -22,7 +22,8 @@ local config = {
     limit_to_hairpins_on_notes = true,          -- if true, only hairpins attached to notes are considered
     vertical_adjustment_type = "far",           -- possible values: "near", "far", "none"
     horizontal_adjustment_type = "both",        -- possible values: "both", "left", "right", "none"
-    vertical_displacement_for_hairpins = 12     -- alignment displacement for hairpins relative to dynamics handle (evpu)
+    vertical_displacement_for_hairpins = 12,    -- alignment displacement for hairpins relative to dynamics handle (evpu)
+    extend_to_expression_in_next_bar = false    -- if true, extends to an expression at the beginning of the next bar    
 }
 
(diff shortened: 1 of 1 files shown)


Nick-Mazuk:
I personally thing the default behavior is that they should always stop at the end of the measure in this case—regardless on if the two measures are on the same system or not. I do agree that it can get really ugly to have a hairpin stop at beat 1 on a new system.

The reason for always doing this is two-fold:

1. The system breaks are different in the score and the parts
2. I (and I presume many people) don't do any layout work until after the music is written. So the system breaks will likely change after applying the hairpins.

Don't have enough time right now to do an actual code review, but will look at this later.

rpatters1:
Just to clarify how *I* plan to use it, given what you said about layout. Before page layout I will only select to the end of bars where I don't want it to extend yet. In fact, I may choose *never* to extend in the score. Somehow the lack of extensions bothers me more in parts.

But the way the enhancement works is that if you don't select the following bar with the dynamic, it behaves the way it always did.

Nick-Mazuk:
I see. I do agree that it looks weird to have the hairpins stop at the barlines in the parts. Perhaps it's worth it to have a separate script you can run on all the hairpins on all the parts to do this extension? That way you can fix it in one go for every applicable hairpin after adjusting the layouts in the parts.

By the way, I've since adopted Carl Vine's Hairpin Creator script over this one since it seems to handle more edge cases better.

rpatters1:
It could be a setting to do it only in parts. I'll have to see how I use it. I might come back and add that.

cv-on-hub:
> Hairpin Creator script over this one since it seems to handle more edge cases better.

Elaine Gould ("Behind Bars") shows multiple options of hairpins either connecting across the barline or ending before it depending on very subtle context ... one of the reasons to employ configuration before each run of `hairpin_creator`.

rpatters1:
If I understand it correctly, the Hairpin Creator has a wider scope than this script. All this one does is reposition and align dynamics and hairpin endpoints. It doesn't create them.

Anyway, I've tweaked it enough that with my person config settings it works nearly all the time correctly for me.

rpatters1:
Any objections to me merging this? I'll leave the new behavior off by default.

rpatters1:
Without objection, I will merge this within a few hours or so. (I'm in housecleaning mode.)

Nick-Mazuk:
Merge away.

PR #324, opened 2022-08-06
Add 'notefinder' library function

jwink75:
- This library will return start/end notes for use with the smartshape library (or anything else that might find it useful!
- Currently there is a single function for returning a table of the starting/ending notes for each staff in a given selection.


rpatters1:
I think it is very risky to keep entries from an `eachentry` loop in a table outside that loop. Depending on when Lua garbage collection runs, I believe you could end up with entries pointing to deleted data. Finale will then crash if you try to access it. This is worth running a test to see if you can break it, but I'm fairly certain it is not safe.

Instead of a Lua table, this function might return an `FCNoteEntryLayers` instance. This allows the storing of entries from multiple measures. The way `tie.lua` does this is to return the FCNoteEntryLayer instance along with the 1st and last entry taken from that instance. (But read the caveats!) Because you potentially want multiple layers you'll need to return `FCNoteEntryLayers`.

rpatters1:
As a follow-on, I was not mistaken. This script crashes Finale, if you select at least three measure containing entries. (Maybe even just with 2, but 3 guarantees that Lua garbage collection has run on the first entry returned by `eachentry`.)

```lua

local extern
for entry in eachentry(finenv.Region()) do
    if not extern then
        extern = entry -- save the 1st entry we encounter
    end
end

-- this call crashes Finale because the entry has gone
-- out of scope in C++
local lownote = extern:CalcLowestNote(nil)
```

I strongly urge you to go look at the `eachentry` function and internalize how it works. You can find it [here](https://github.com/finale-lua/lua-source/blob/master/built-in-functions/rgplua_built_in_functions.lua).

jwink75:
Thanks, Robert… I’ll figure out another approach!

jwink75:
Out of curiosity I just ran the script I have that uses the function I wrote (where entries are stored in a table) on a 6+ measure passage… and it did not crash. Why do you suppose that is, if using tables for noteentry storage is an issue?

rpatters1:
Predicting Lua gc is difficult. You could run it successfully 100 times (or a 1000) then one day it could start crashing Finale every time. Perhaps after one small change.

Also, it matters what you do with the collection after you've made it.

jwink75:
I'm going back to the drawing board on this one... I'm thinking maybe a library function is not the right solution for what I want to do if I can't pass notes around in tables.

rpatters1:
You can pass them around in tables as long as you maintain a reference to their container. I recommended `FCNoteEntryLayers`, and I still think it would. You could take one of those as input and then return the start and end note. The problem arises when you try to take entries from the `eachentry` loop.

jwink75:
How would I got about maintaining a reference to the container? I'm not just interested in the start and end notes, that is just a jumping off point... Is there a good way to iterate through notes in FCNoteEntryLayers comparable to eachentry?

I read through the available functions and parameters for FCNoteEntryLayers and it didn't seem like the right tool for the job, but I'll go look at it again.

On 8/6/2022 4:06:15 PM, rpatters1 ***@***.***> wrote:
You can pass them around in tables as long as you maintain a reference to their container. I recommended FCNoteEntryLayers, and I still think it would. You could take one of those as input and then return the start and end note. The problem arises when you try to take entries from the eachentry loop.

rpatters1:
You maintain the reference by having a variable that's in scope, either a local variable in your code block or a global variable. You don't let it go out of scope until you are finished with it. Pretty much like any other Lua variable you ever use.  

The problem with `eachentry` is that it is creating variables under the covers that are hidden from you and feeding entries from them into your loop body, but then they go out of scope when the loop exits. It's really only good for procession entries one at a time.

One way to iterate `FCNoteEntryLayers` is

```lua
for note_entry_layer in each(note_entry_layers) do
   for entry in each(note_entry_layer) do
      ...
   end
end
```
If you only want entries inside a region:

```lua
for note_entry_layer in each(note_entry_layers) do
   for index = 0, note_entry_layer.Count-1 do
      local entry = note_entry_layer:GetRegionItemAt(index)
      if entry then
         ...
      end
   end
end
```
Both `FCNoteEntryLayers` and `FCNoteEntryLayer` are collections, so they have all the usual collection functions for iterating them.

Nick-Mazuk:
Don't have Finale open right now, but doesn't each note/entry have a unique ID? If so, you can get those IDs within the loop and store them even when the loop ends. You can then load a specific entry by ID.

I may be totally misremembering these classes, though.

rpatters1:
The EntryNumber is unique. Loading each one separately requires a good deal of redundant work, though.

Issue #313, opened 2022-08-03
Capitalization Rules for Menu Items


rpatters1:
The entire computer world of macOS and Windows uses Title Capitalization for menu items. I have been dismayed to see this repository straying further and further from this simple and near universal practice (in English). Title Capitalization is summed [here](https://www.grammarly.com/blog/capitalization-rules/) very nicely, but here is the salient quote:

>The capitalization rules for titles of books, movies, and other works vary a little between style guides. In general, you should capitalize the first word, all nouns, all verbs (even short ones, like is), all adjectives, and all proper nouns. That means you should lowercase articles, conjunctions, and prepositions—however, some style guides say to capitalize conjunctions and prepositions that are longer than five letters.

I am curious why we are straying from this, especially since the rest of Finale uses it. (Again, in English. Spanish and other languages have difference conventions.) My view is that we should use the same capitalization convention for our menu items as Finale does.

cv-on-hub:
It's true - I find I've been lower-casing recent menu items, and the only reason I can imagine is that it's a subconscious response to the universal snake_casing of the finalelua universe. This is an important point, which I'd never pondered, especially considering the universal capitalisation in Win/Mac.

Nick-Mazuk:
I actually never noticed that menu items on macOS are title case. By default, I always use sentence case because there's been quite a few large scale studies to show it's the easiest to read. But I agree, we should be consistent with the platforms.

Issue #270, opened 2022-07-27
Naming scripts


Nick-Mazuk:
> There’s not necessarily a conventional naming system for the scripts, so categories of things like Layers for example - You’ve got "Layers Hide" and "Layers Mute", which will end up next to each other alphabetically in the menu, but "Swap Layers" will end up way farther down.  But I can rename the last one "Layers Swap” and that will put that script next to the other Layers scripts in the menu.  Anyway, this is just my thought process here.  I’m trying to figure out the best organization for these.  Because… I suppose there is going to be an ever growing list of these and even just keeping track of them is gonna be a challenge at some point!

This is from an email I received from Jason Loffredo.

At some point we should probably figure out a standard for naming scripts. It seems like each of us has our own system, but they are not necessarily the same between each of us.

cc @rpatters1 @cv-on-hub @jwink75

rpatters1:
The Category string in `plugindef` seems to be how Jari was planning to organize this issue. (If the JW Lua plugin itself does anything with it, I haven't figured it out.)

Rather than trying to standardize file names (which will always be a tussle), I would recommend expanding or modifying the Category strings to an extend list (more than Jari's original list) and using those to present them to users on the website.

cv-on-hub:
I've tried to standardise at least my own script names, with a primary-affected-item-first convention that I thought I saw hinted at somewhere in the Hub documentation. **Category string** would certainly be the most sensible use of `plugindef`, though I keep forgetting to include the categories.

jwink75:
This is something that I think has evolved somewhat over time. For example, I would bet that 'Swap Layers' was mine initially, but if I were submitting it today I would 100% have named it layer_swap_1_2 or something for just that reason. I try to stick with the idea of going big to small... What does the script affect at the highest level, then getting more specific from there.

Perhaps a tidying pass is in order, and that's also something we should all pay close attention to in the future as well!

On 7/27/2022 2:13:34 PM, Nick Mazuk ***@***.***> wrote:
[...]

PR #330, opened 2022-08-10
Add docs on script names

Nick-Mazuk:
We've had two threads on naming scripts. I think we're in consensus on at least the basics of naming scripts. So I'm adding what we agreed upon to the docs.

Fix https://github.com/finale-lua/lua-scripts/issues/313
Fix https://github.com/finale-lua/lua-scripts/issues/270

Looking for a LGTM from both @rpatters1 and @cv-on-hub before merging. Merging this PR will close the related issues.


cv-on-hub:
I suggest, _"A script's name"_ should become _"A script's name as returned by plugindef()"_, and we seem to agree on Capital Case for script names appearing in Finale menus.

The _"Use a consistent file name"_ section remains solid and agrees with my previous assumption about: _"take things from general to specific"_

rpatters1 (approved):
Looks good to me.

cv-on-hub (approved):
LGTM also

PR #331, opened 2022-08-14
Update 'staff_rename', add 'systems_lock'

jwink75:


src/staff_rename.lua
```diff
@@ -163,40 +213,39 @@ function staff_rename()
...
         end
 --
-
         function add_ctrl(dialog, ctrl_type, text, x, y, h, w, min, max)
             str.LuaString = text
-            local ctrl
+            local ctrl = ""
             if ctrl_type == "button" then
-                ctrl = dialog:CreateButton(x, y)
+                ctrl = dialog:CreateButton(x, y + 2)
```
Nick-Mazuk, line 230: I don't think it's incredibly important to fix here, but this `2` is what's called a "magic number". A magic number is essentially just a hard-coded number like this where there's no real description of where it comes from. And with this number being the exact value it is, things somehow work like magic.

It's best to avoid magic numbers whenever possible. Though admittedly, here it makes sense. You're just increasing the vertical spacing by 2 pixels.

Here's a discussion link from one of Carl Vine's PRs about magic numbers:

https://github.com/finale-lua/lua-scripts/pull/314#discussion_r936174719

src/staff_rename.lua
```diff
@@ -206,30 +255,111 @@ function staff_rename()
...
+        local master_autonumber_popup = add_ctrl(dialog, "popup", "", col[3] + 60, row[1], row_h, col_w - col_gap, 0, 0)
+        for i, k in pairs(autonumber_style_list) do
+            str.LuaString = autonumber_style_list[i]
+            master_autonumber_popup:AddString(str)
+        end
+        add_ctrl(dialog, "horizontalline", "", 0, row[2] + 8, 0, col_w * 3.5 + 20, 0, 0)
+        str.LuaString = "*Custom*"
+        master_autonumber_popup:AddString(str)
+
         --local h_line = add_ctrl(dialog, "horizontalline", 0, row[1], 1, col_w * 3, 0, 0)
```
Nick-Mazuk, line 277: Likewise, delete this commented code.

src/staff_rename.lua
```diff
@@ -206,30 +255,111 @@ function staff_rename()
             return ctrl
         end
 
+        local autonumber_style_list = {"Instrument 1, 2, 3", "Instrument I, II, II", "1st, 2nd, 3rd Instrument",
+            "Instrument A, B, C", "1., 2., 3. Instrument"}
+        local autox_width = 40
```
Nick-Mazuk, line 260: Nit: `auto_x_nit` not `autox_nit`.

src/staff_rename.lua
```diff
@@ -206,30 +255,111 @@ function staff_rename()
...
+                edit_fullname[i]:GetText(str)
+                for j, l in pairs(staff_name) do
+                    if (staff_name[j] == str.LuaString) and (autonumber_check[i]:GetCheck() == 1) then
+                        is_match = true
+                        inst_num = inst_nums[j]
+                        inst_nums[j] = inst_nums[j] + 1
+                    end
+                end
+
+--                require('mobdebug').start()
```
Nick-Mazuk, line 342: Delete

src/staff_rename.lua
```diff
@@ -206,30 +255,111 @@ function staff_rename()
...
             str.LuaString = forms[i]
             form_select:AddString(str)
-        end
+        end   
+        local hardcode_autonumber_btn = add_ctrl(dialog, "button", "Hardcode Autonumbers", col[3] + autox_width, row[row_count + 2], row_h, col_w, 0, 0)
         --
         dialog:CreateOkButton()
         dialog:CreateCancelButton()
         --
+        function hardcode_autonumbers()
```
Nick-Mazuk, line 314: Not worth changing right now, but a general rule of thumb is to not define functions inside other functions (unless it's a short, simple, one-off function). It works, but it's harder to maintain.

The reason is that functions defined inside other functions are harder to reason about how they work. Ideally, we want a function to be as predictable as possible, and that often means writing "pure" functions.

A pure function is one that meets the following criteria:

1) It's behavior depends solely on it's arguments
2) It does not have side effects<sup>*</sup>

Here's a more rigorous definition: https://en.wikipedia.org/wiki/Pure_function

By defining a function inside another function, that inner function has access to all variables that exist in that outer function and can therefore mutate any of the outer function's variables. This can lead to incredibly unpredictable behavior whenever someone wants to update the script. So it's best not to mutate the outer function's variables.

The best to prevent that is to define functions at the top-level, so there is no nesting. Then you can't mutate an outer function's variables!

Likewise, it's best not to use the outer function's variables in the first place. Because then you essentially have implicit arguments to the function. It looks like I should be able to just call `hardcode_autonumbers()` to run this function. But if it depends on any variables in the outer function, then changing those variables outside the function will then change the behavior of `hardcode_autonumbers()`. That's very unpredictable! So it's best not to even use them.

And if you aren't accessing or mutating an outer function's variables, there's no reason to define a function inside another function.

<sup>*</sup> In Finale Lua code, it's difficult to not have any side effects since we are constantly updating FCNotes, FCEntries, and other Finale objects. In this repository, I think that's perfectly acceptable and actually a good practice to allow those mutations. But mutating variables outside the function's scope is a bad practice.

src/systems_lock.lua
```diff
@@ -0,0 +1,75 @@
...
+    local lock_score = true
+    local lock_parts = true
+
+    if lock == "Score" then
+        lock_parts = false
+    elseif lock == "Parts" then
+        lock_score = false
+    end
+
+    function systems_lock()
```
Nick-Mazuk, line 40: This looks like a good function to move to the top-level scope (see above comment on why).

src/systems_lock.lua
```diff
@@ -0,0 +1,75 @@
...
+            local freeze = sys:CreateFreezeSystem(true)
+            if freeze then
+                freeze:SetNextSysMeasure(next_sys)
+                freeze:Save()
+            end
+            sys:Save()
+        end
+    end
+
+    function parts_switch()
```
Nick-Mazuk, line 55: This looks like a good function to move to the top-level scope (see above comment on why).

Just make sure `lock_score` is an argument to the function.

src/systems_lock.lua
```diff
@@ -0,0 +1,75 @@
...
+        local part_current = finale.FCPart(1)
+        part_current:SetCurrent()
+        local parts = finale.FCParts()
+        parts:LoadAll()
+        for part in each(parts) do
+            if part:IsScore() and lock_score == true then
```
Nick-Mazuk, line 61:
```lua
if part:IsScore() and lock_score then
```

No need to check if a boolean is true. It's already either true or false.

src/systems_lock.lua
```diff
@@ -0,0 +1,75 @@
...
+                part:SwitchTo()
+                systems_lock()
+            end
+        end
+        part_current:SwitchTo()
+    end
+    parts_switch()
+end
+
+formatting_systems_lock(lock)
```
Nick-Mazuk, line 75: Nit: can you add a newline at the end of the file?

https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

https://www.quora.com/Why-is-it-a-common-practice-to-put-a-new-line-at-the-end-of-a-source-code-file

src/staff_rename.lua
```diff
@@ -317,15 +484,18 @@ function staff_rename()
...
             end
         end
-
     end -- function
-
     dialog("Rename Staves")
-
 end -- rename_staves()
 
 staff_rename()
```
Nick-Mazuk, line 501: Nit: can you add a newline at the end of the file? See comment below on why.

src/staff_rename.lua
```diff
@@ -3,8 +3,8 @@ function plugindef()
     -- are both reserved for the plug-in definition.
     finaleplugin.Author = "Jacob Winkler"
     finaleplugin.Copyright = "2022"
-    finaleplugin.Version = "1.1"
-    finaleplugin.Date = "3/18/2022"
```
Nick-Mazuk, line 7: Nit: it looks like this is for renaming selected staves? If so, does it require a selection? If so, please add the following:

```lua
finaleplugin.RequireSelection = true
```

https://www.finalelua.com/docs/rgp-lua/finaleplugin-properties#requireselection-boolean

Nick-Mazuk (changes requested):

jwink75:
Thanks Nick! I’m leading a music summer camp this week, I’ll putter away at these fixes as I can.

At your suggestion, I’ll move the Roman numeral and ordinal conversions into the  utilities library file. I’ll also move the alphabetical one I came up with in there.

src/staff_rename.lua
```diff
@@ -17,15 +17,55 @@ There are buttons for each instrument that will copy the full name into the abbr
...
 
 Speaking of the Bb Clarinet... Accidentals are displayed with square brackets, so the dialog will show "B[b] Clarinet". This is then converted into symbols using the appropriate Enigma tags. All other font info is retained.
-
-Note that this script is not currently able to address Finale's auto-numbering. When this feature is added to future versions of RGP Lua I will update the script to allow for some quick processing of these as well, such as being able to switch numbering systems for all instruments at once, or hard-coding the player number into the full/abbreviated names and removing auto-numbering.
 ]]
     return "Rename Staves", "Rename Staves", "Renames selected staves"
 end
 
+-- taken from https://exercism.org/tracks/lua/exercises/roman-numerals/solutions/Nia11 on 2022-08-13
+function calc_roman_numeral(n)
```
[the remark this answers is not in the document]
jwink75: I'll pull out all three of these... Roman, ordinal, and alphabetic (A1, B1, C1 etc) into the utilities.lua

src/staff_rename.lua
```diff
@@ -163,40 +213,39 @@ function staff_rename()
...
+                ctrl = dialog:CreateStatic(x, y + 4)
             elseif ctrl_type == "verticalline" then
                 ctrl = dialog:CreateVerticalLine(x, y, h)
             end
             if ctrl_type == "edit" then
                 ctrl:SetHeight(h-2)
                 ctrl:SetWidth(w - col_gap)
             elseif ctrl_type == "horizontalline" then
-                ctrl:SetY(y + h/2)
+--                ctrl:SetY(y + h/2)
```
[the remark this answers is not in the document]
jwink75: I need to look into why all of a sudden I started getting errors with that line when I hadn't before, and I wanted to move on with the interface... I don't know if I'm quite ready to take it out all the way yet :)

src/staff_rename.lua
```diff
@@ -163,40 +213,39 @@ function staff_rename()
...
             if ctrl_type == "button" then
-                ctrl = dialog:CreateButton(x, y)
+                ctrl = dialog:CreateButton(x, y + 2)
             elseif ctrl_type == "popup" then
                 ctrl = dialog:CreatePopup(x, y)
             elseif ctrl_type == "checkbox" then
                 ctrl = dialog:CreateCheckbox(x, y)
             elseif ctrl_type == "edit" then
-                ctrl = dialog:CreateEdit(x, y - 2)
+                ctrl = dialog:CreateEdit(x, y - 0)
```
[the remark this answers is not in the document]
jwink75: that 0 used to be a magic number that I was fiddling with ... it was easier to finetune with that there. I can take it out.

src/staff_rename.lua
jwink75: I totally understand what you're saying... And I wish I didn't need the magic numbers to make the UI look good, but the elements just don't line up without them. 

One of these days I'm going to do a super comprehensive comparison between the way MacOS and Windows draw everything and how they line up, and try to do this sort of thing more systematically...

src/systems_lock.lua
jwink75: Ha! Caught me! This is one I wrote before Robert schooled me on this!

src/staff_rename.lua
jwink75: If you make a selection, it processes the selected staves. If you do not, it processes them all.

src/staff_rename.lua
jwink75: Like hit return one more time?

ThistleSifter (commented):
I would write `calc_ordinal` like this:
```lua
    local units = num % 10
    local tens = num % 100
    if units == 1 and tens ~= 11 then
        return = num .. "st"
    elseif units == 2 and tens ~= 12 then
        return num .. "nd"
    elseif units == 3 and tens ~= 13 then
        return num .. "rd"
    end

    return num .. "th"
```
This handles the cases for 111, 112, 113 correctly.
It's also generallly not a good idea to have the same result (in this case `num .. "th"`) appear more than once. If that happens, it usually means that the conditions (what goes in the `if`, `elseif`, `else`) need rethinking.

(Also, for readability, put spaces between an operator and its operands)

src/library/utils.lua
```diff
@@ -72,4 +72,76 @@ function utils.round(num)
...
+
+This function was written to emulate the way Finale numbers saves when Autonumber is set to A, B, C... When the end of the alphabet is reached it goes to A1, B1, C1, then presumably to A2, B2, C2. 
+
+If anyone wants to take the time to figure out what Finale does once you reach Z9 staves I'd love to know!  
+Email me at jacob.winkler at mac dot com!
+
+@ num (number)
+: (string)
+]]
+function utils.calc_alphabet(num)
```
ThistleSifter, line 133: If I've read this correctly, then I think the following results should be given:
```lua
calc_alphabet(1)  -- Expected result: "A"  -- Actual result: "A"
calc_alphabet(26) -- Expected result: "Z"  -- Actual result: "Z"
calc_alphabet(27) -- Expected result: "A1" -- Actual result: "A1"
calc_alphabet(52) -- Expected result: "Z1" -- Actual result: "@2"
calc_alphabet(53) -- Expected result: "A2" -- Actual result: "A2"
```

Assuming I've understood it correctly, maybe try the following, which should give all of the expected results above:
```lua
    local letter = ((num - 1) % 26) + 1
    local n = math.floor((num - 1) / 26)

    return string.char(64 + letter) .. (n > 0 and n or "")
```

ThistleSifter:
One more small thing: the function names at the start of the doc blocks for the utility functions are missing the underscores. They should be exactly the same as the function name in the code.

PR #344, opened 2022-08-21
Create hairpin_creator_configuration.lua

cv-on-hub:
New script to configure `hairpin_creator` without using modifier key invocation. This has proven distinctly more convenient for a KM-freak than using the modifier key invocation. Configuration alone was provided with the original version of the script as an extra menu item whether you wanted it or not. Isolated like this it's only there if you want it.


rpatters1:
It should be an AdditionalMenuItem. Do not duplicate this code. In my setup, it will end up as a menu item either way, so just let it be a menu it.

rpatters1 (commented):
The "Configuration" menu option should have "..." to indicate it opens a dialog:

```
Configuration...
```

cv-on-hub:
> The "Configuration" menu option should have "..." to indicate it opens a dialog:

Of course. Will do.

rpatters1 (approved):
lgtm

Issue #343, opened 2022-08-20
Bug report


Nick-Mazuk:
**Description**

hello, trying to work with smufl_load_engraving_defaults plugin 
I get this answer: "...module 'lunajson.decoder' not found"
any help? thanks

**Metadata**

- Finale v27
- macOS 12
- Script: smufl_load_engraving_defaults.lua

_Created through the website by Alessandro Melchiorre amelchi@gmail.com_

rpatters1:
They can fix it now by downloading the luajson directory to be in the directory where their script is. Longer term, that could be added to the distribution packager.

Nick-Mazuk:
It should now be bundled. As of the bundler rewrite, it accepts any arbitrary static imports as long as it's in the `src` folder.

rpatters1:
I'm looking at `dist/smufl_load_engraving_defaults.lua` and it included `luajson.luajson` but not `luajson.decoder` and `luajson.encoder`. (This are required by `luajson/luajson.lua`.

Nick-Mazuk:
I see. It's likely because these require statements don't use parenthesis.

https://github.com/finale-lua/lua-scripts/blob/c04293031839b6d18759f5d0035c7522c4466edd/src/lunajson/lunajson.lua#L1-L3

This is how an import is currently defined in the parser:

https://github.com/finale-lua/lua-scripts/blob/c04293031839b6d18759f5d0035c7522c4466edd/.github/actions/bundle/src/helpers.ts#L2

Nick-Mazuk:
#345 (nice PR number) should fix this problem, but let's leave this issue open as the long-term fix is to update the bundler to support this edge case.

PR #345, opened 2022-08-21
fix lua json require statements

Nick-Mazuk:
Fixes the bug noticed in https://github.com/finale-lua/lua-scripts/issues/343.

Though let's leave that issue open for now since the long-term solution is to update the bundler to support that syntax for require.

Feel free to merge when ready.

### DIFF
--- a/src/lunajson/lunajson.lua
+++ b/src/lunajson/lunajson.lua
@@ -1,6 +1,6 @@
-local newdecoder = require 'lunajson.decoder'
-local newencoder = require 'lunajson.encoder'
-local sax = require 'lunajson.sax'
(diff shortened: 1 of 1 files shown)


rpatters1 (approved):
Did you check if any of those subfiles have offending require statements? Otherwise, lgtm.

amelchi:
sorry I am not a developer...
what should I do? :)
thanks

rpatters1:
As soon as this Pull Request is merged, you should be able to get the new version and it should work.

Nick-Mazuk:
Yes, none of the subfiles have any require statements. Merging.

PR #350, opened 2022-08-24
Tweaks to harp pedal wizard

jwink75:
- Improved pentatonic and whole tone scale algorithms
- Replaced one internal function with a library function
- Added locrian scale

### DIFF
--- a/dist/smufl_load_engraving_defaults.lua
+++ b/dist/smufl_load_engraving_defaults.lua
@@ -16,10 +16,1441 @@ function require(item)
     return __import_results[item]
 end
 
+__imports["lunajson.decoder"] = function()
+    local setmetatable, tonumber, tostring =
+          setmetatable, tonumber, tostring
+    local floor, inf =
+          math.floor, math.huge
+    local mininteger, tointeger =
+          math.mininteger or nil, math.tointeger or nil
+    local byte, char, find, gsub, match, sub =
+          string.byte, string.char, string.find, string.gsub, string.match, string.sub
+
+    local function _decode_error(pos, errmsg)
+    	error("parse error at " .. pos .. ": " .. errmsg, 2)
+    end
+
+    local f_str_ctrl_pat
+    if _VERSION == "Lua 5.1" then
+    	-- use the cluttered pattern because lua 5.1 does not handle \0 in a pattern correctly
+    	f_str_ctrl_pat = '[^\32-\255]'
+    else
+    	f_str_ctrl_pat = '[\0-\31]'
+    end
+
+    local _ENV = nil
+
+
+    local function newdecoder()
+    	local json, pos, nullv, arraylen, rec_depth
+
+    	-- `f` is the temporary for dispatcher[c] and
+    	-- the dummy for the first return value of `find`
+    	local dispatcher, f
+
+    	--[[
+    		Helper
+    	--]]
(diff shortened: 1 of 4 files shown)


rpatters1 (commented):
B-but locrian has no sol!

(I have no issues with this code. I'll let @Nick-Mazuk  comment first.)

jwink75:
> 

I asked Jason whether he could name me a single piece actually written in locrian off the top of his head... Nope! Still, it was easy enough to add!

rpatters1:
FWIW: I'm still seeing a ton of lint errors on the harp_pedal script. Mostly undefined globals, but also the use of `NULL` when you should be using `nil`. Not sure if you have fixed them.

The tie script looks great.

jwink75:
I thought I had fixed the null thing, but I have definitely been having issues with GitHub, so I’ll take another look. 

As for the globals… it’s there I way I can check the lint errors on GitHub?

rpatters1:
I use Visual Studio Code with the recommended settings.

jwink75:
Is there a way to see lint errors on ZeroBrane Studio?

jwink75:
Found those NULLs you mentioned. They are all replaced with nil now, but for what it’s worth that’s the way Jari had the AlertYesNo sample code :)

jwink75:
I believe I have VS Code set up correctly on my MacBook, and I have enabled Sync Settings so that when I install it on my Windows computer it should theoretically be set up the same. The only errors VS Code is showing on the harp script now are in the window_pos_x and y globals (two errors total), having to do with window position saving, which I don't think I'm getting right... But would like to go ahead and merge and fix later, especially as I turn my attention to other things like the bug fixes I just put in the staff_rename script.

@Nick-Mazuk, @rpatters1, OK to merge and close this PR?

jwink75:
Oops... I did not mean to commit that myself :P

rpatters1:
Hey @jwink75 I may have messed this pull request up, based on a comment I'm seeing right above this one. But I am now seeing some more serious lint errors on this script than undefined globals. For example, it is complaining that Line 150:

```lua
                new_pedals[i] = harpstrings[i]
```

is assigning a string to a number, which is certainly how it looks.

jwink75:
I will try to take a look sometime this weekend. Sent from my iPhoneOn Jan 28, 2023, at 7:14 AM, rpatters1 ***@***.***> wrote:﻿
Hey @jwink75 I may have messed this pull request up, based on a comment I'm seeing right above this one. But I am now seeing some more serious lint errors on this script than undefined globals. For example, it is complaining that Line 150:
                new_pedals[i] = harpstrings[i]
is assigning a string to a number, which is certainly how it looks.

rpatters1:
Never mind. I think the problem was that I was looking at code based on an old branch of the code. It's back to just the `window_pos_x/y` warnings.

BTW: these would be easy to fix. You probably just need to add at the top:

```lua
window_pos_x = nil
window_pos_y = nil
```

But an even better solution would be to convert the dialog to `FCXCustomLuaWindow` using the `mixin` library. You could delete a lot of code then.

rpatters1:
Hmm, when I open the harp_pedal_wizard script in VS Code it spits out more lint errors, including line 150 above that looks wrong. Anyway, if you could give it a glance that would be great. Lines 150 and 272 are the ones that look weird.

PR #352, opened 2022-08-26
Allow Items to Persist Outside eachentry

ThistleSifter:
So I had a number of interesting (if not rather crazy ideas) about how to go about this in ways which wouldn't involve copying the whole `eachentry` function, but in the end they all boiled down to using `debug` methods to do some trickery. And the `debug` library is not known for (or designed to be) fast.

So I settled on this solution, which just involved adding two mixins (`FCNoteEntryCell` and `FCNoteEntry`) and adding a copy of the `eachentry` function to the `mixin` library. The only modifications to `eachentry` were to change the calls to `finale.FCNoteEntryCell` to `mixin.FCMNoteEntryCell`.

For testing, I was using a modified version of the script posted here by @rpatters1 : https://github.com/finale-lua/lua-scripts/pull/324#issuecomment-1207261172

Unfortunately, I couldn't make Finale crash with the original script, so I'm not 100% certain, but the modified version below should work:
```lua

local extern
for entry in mixin.eachentry(finenv.Region()) do
    if not extern then
        extern = entry
    end
end

local lownote = extern:CalcLowestNote(nil)
```

I'm marking this a draft for now. If this works, I need to reorganise some things in the mixin library to tidy up the slight mess introduced by adding `eachentry` before I'm happy to have this merged.

### DIFF
--- a/src/library/mixin.lua
+++ b/src/library/mixin.lua
@@ -12,32 +12,32 @@ The Fluid Mixins library does the following:
 - As an added convenience, all methods that return zero values have a fluid interface enabled (aka method chaining)
 
 
-## finalemix Namespace
-To utilise the new namespace, simply include the library, which also gives access to he helper functions:
+## mixin Namespace
+To utilise the new namespace, simply include the library, which also gives access to the helper functions:
 ```lua
-local finalemix = require("library.mixin")
+local mixin = require("library.mixin")
 ```
 
-All defined mixins can be accessed through the `finalemix` namespace in the same way as the `finale` namespace. All constructors have the same signature as their `FC` originals.
+All defined mixins can be accessed through the `mixin` namespace in the same way as the `finale` namespace. All constructors have the same signature as their `FC` originals.
 
 ```lua
 local fcstr = finale.FCString()
 
 -- Base mixin-enabled FCString object
-local fcmstr = finalemix.FCMString()
+local fcmstr = mixin.FCMString()
 
 -- Customised mixin that extends FCMString
-local fcxstr = finalemix.FCXString()
+local fcxstr = mixin.FCXString()
 
 -- Customised mixin that extends FCXString. Still has the same constructor signature as FCString
-local fcxcstr = finalemix.FCXMyCustomString()
+local fcxcstr = mixin.FCXMyCustomString()
 ```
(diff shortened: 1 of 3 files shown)


rpatters1:
To crash Finale, try running the raw version on a region containing entries in several measure and staves.

ThistleSifter:
I tried that (selected whole document with 50+ staves, ~200 bars) and Finale was still standing.

rpatters1:
The problem is that the current script doesn't control when garbage collection runs, so it fails randomly. This will fail consistently:

```lua
local extern
for entry in eachentry(finenv.Region()) do
    if not extern then
        extern = entry
    end
    collectgarbage("collect") 
end

local lownote = extern:CalcLowestNote(nil)
```

ThistleSifter:
Nope, still doesn't crash. Just hangs for a short while, presumably because it's now gc'ing many thousands of times over.

rpatters1:
Try this:

```lua
local extern
for entry in eachentry(finenv.Region()) do
    if not extern then
        extern = entry
    end
    collectgarbage("collect") 
end

finenv.UI():AlertInfo("calling calc", "")
local lownote = extern:CalcLowestNote(nil)
```

This is the first one that fails for me without running it under a debugger.

ThistleSifter:
Still doesn't crash.

rpatters1:
Weird. Maybe it's a difference in how Macs and Win collect unused memory. But I have traced this code and the FCNoteEntryCell instances are calling their dtor functions as each cell is processed (with the explicit garbage colleciton). You could try debugging the code in ZeroBrane, placing a breakpoint right at the call to CalcLowestNote.

rpatters1:
If it's at state where I can try it, I will trace it and make sure it works. But it looks good to me so far.

ThistleSifter:
Yes, it's ready for testing.

rpatters1:
Yes it works. Much better than what I was thinking about because it only maintains references as needed.

rpatters1:
Another comment is that the "parent" `FCXNoteEntryCell` instance probably needs to be accessible so that entries can be saved if changed. Maybe this is already possible. I'm just raising the issue to be sure.

ThistleSifter:
Yes, that is already possible through `GetParent()`.

ThistleSifter:
This is ready to go now.

rpatters1 (approved):
lgtm

PR #353, opened 2022-08-27
Update rest_offsets.lua

cv-on-hub:
converted dialog to FCXCustomLua; added "Zero = Floating Rest" checkbox, text colouration, revised documentation

### DIFF
--- a/src/rest_offsets.lua
+++ b/src/rest_offsets.lua
@@ -2,18 +2,24 @@ function plugindef()
     finaleplugin.RequireSelection = true
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
     finaleplugin.AuthorURL = "http://carlvine.com/lua/"
-    finaleplugin.Version = "v1.36"
-    finaleplugin.Date = "2022/08/03"
+    finaleplugin.Version = "v1.38"
+    finaleplugin.Date = "2022/08/27"
     finaleplugin.Notes = [[
-    Several situations including cross-staff notation (rests should be centred between the staves) 
-    require adjusting the vertical position (offset) of rests. 
-    This script duplicates the action of Finale's inbuilt "Move rests..." plug-in but needs no mouse activity. 
-    It is also an easy way to reset rest offsets to zero in every layer, the default setting. 
-    (An offest of zero centres on the middle staff line.)
-]]
+        Several situations including cross-staff notation (rests should be centred between the staves) 
+        require adjusting the vertical position (offset) of rests. 
+        This script duplicates the action of Finale's inbuilt "Move rests..." plug-in but needs no mouse activity. 
+        It is also an easy way to reset rest offsets to zero in every layer, the default setting. 
+
+        By default an offest of ZERO centres rests on the middle staff line. 
+        If you instead want zero to set a FLOATING rest, to respect the `Floating Rests` setting 
+        under Document Options... -> Layer, then select the "Zero = Floating Rest" checkbox.
+    ]]
    return "Rest Offsets", "Rest Offsets", "Rest vertical offsets"
 end
 
+local mixin = require("library.mixin")
+
+-- ================= SCRIPT BEGINS =================================
 -- RetainLuaState retains one global:
 config = config or {}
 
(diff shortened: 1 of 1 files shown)


rpatters1 (changes requested):
I find the comment about "Zero = Floating Rest" to be a bit confusing. I think you should spell out exactly which option you mean in the Document Options->Layers screen and give an example of how it would work.

I'm also leery of using a lot of color on the screen. Remember that color is very personal, and many people are color-blind. I think color should be optional. In particular, reddish hue I find much harder to read that blue or black. (I don't know which color you chose, however. I just remember some earlier iterations where the color was red.) But I won't hold up merging because of color. I just need some clarification on the comment about "Zero = Floating Rest".

cv-on-hub:
Thanks for the advice Robert. Amended instructions and added a `?` info alert to the window. Had to revert to finale.FCCustomLuaWindow because couldn't get FCXCustomLuWindow to produce nested finenv.UI():AlertInfo dialog.

cv-on-hub:
Discovered aberrant tab spacing. Now fixed.

cv-on-hub:
Oops - one more indent. (Wish I could see how it would register on GitHub before posting!)

rpatters1:
@ThistleSifter Do you know why @cv-on-hub is having troubling invoking `AlertInfo` inside a button callback from `FCXCustomLuaWindow`?

ThistleSifter:
@cv-on-hub Could you post the code for the button handler when you tried it with `FCXCustomLuaWindow`?

cv-on-hub:
Certainly. The same as used with FCCustomLua (as in #[6754c21]):
```
local info_button = dialog:CreateButton(128, 75)
-- info_button: --- etc. setup
dialog:RegisterHandleControlEvent ( info_button,
    function() -- (control)
        local msg = "Newly created rests are \"floating\" and will avoid entries in other layers (if present) "
        .. "using the setting for \"Adjust Floating Rests by...\" in \"Document Options...\" -> \"Layers\". \n\n"
        .. "This script stops rests \"floating\", instead \"fixing\" them to a specific offset from the middle staff line. "
        .. "To return them to \"floating\", select the \"Zero = Floating Rest\" option and set the offset to zero."
        finenv.UI():AlertInfo(msg, "Rest Offsets Info")
    end
)
```

ThistleSifter:
Here's the fix:
```lua
dialog:CreateButton(128, 75):AddHandleCommand(function(self)
    -- Do some stuff here
end)
```
There are two parts to this:
With mixins, it's sort of intended for events to be handled at the control level. Apart from enabling better structured/tidier code, it also means that controls can have their own custom events (eg the `Change` event that I added to `FCMCtrlEdit`). I realise that it's not documented very well, my apologies. I need to do a big documentation update for mixins because so many things have changed or become clearer now they've started to be used.

The other part is that what I wrote to override `RegisterHandleControlEvent` obviously just doesn't work and I need to fix that. Again, my apologies.
&nbsp;
&nbsp;
@rpatters1 I'm not sure if anything can be done about this, but I noticed a while ago that if an error is triggered in a button handler that the error message is never displayed to the user. Is there anything that can be done about this? Here's a test script:
```lua
local button = dialog:CreateButton(20, 20)

dialog:RegisterHandleControlEvent(button, function()
    -- This error is never displayed, it just silently exits
    error("an error inside a handler")
end)

-- This error is always displayed
-- error("an error in normal flow")
dialog:ExecuteModal(nil)
```

rpatters1:
Odd you should mention errors inside callbacks. It has been on my todo list for a while, and I literally implemented the fix this evening. (It will be in the next version, 0.65.)

I would be curious to know what JW Lua does with these. RGP Lua 0.65 will terminate the lua session, immediately closing any dialogs, and display an error message box, unless you execute it in `pcall`. (I finally figured out how the exceptions get passed up the chain, by LuaBridge and then Lua to RGP Lua. It is actually fairly clever how the LuaBridge code manages the C++ exceptions and keeps the Lua code from having to know about C++ exceptions.)

cv-on-hub:
Many thanks @ThistleSifter - that works well. Always nicer using `mixin` windows! This script is back to using `mixin`

cv-on-hub:
dammit - more indent errors!

rpatters1 (approved):
lgtm

rpatters1:
@ThistleSifter I just checked, and JW Lua (like RGP Lua until now) only puts error messages in callbacks into the print output. It was a little more obvious with JW Lua if you were running in the JW Lua console. With RGP Lua, the only way you would see the errors is if you were running in the ZeroBrane debugger.

jwink75:
Just as an FYI, CJ is color blind and is a fantastic resource for letting folks know what works and what doesn’t for UI work

PR #355, opened 2022-08-30
Rgp changes for rgplua 0.64

rpatters1:
Documentation changes and bug fixes for RGP Lua v0.64 (which is now live).


rpatters1:
I'm going ahead and merging this because it is almost completely documentation with only a couple minor bugfix changes.

PR #357, opened 2022-08-30
Fixed bug in clef.clef_change

jwink75:


jwink75:
Ugh, I'm sorry I keep doing that... I'm trying to get it where it doesn't keep piling 'commits' on top of each other! The frustrating this I'm about 99% sure that even though I just merged this, the next time I make any changes all these commits will still be 'pending'

PR #358, opened 2022-09-05
Added 'dynamics' and 'metrics' libraries, dynamics_position script

jwink75:
- These contain functions for calculating various features of a region, and then for manipulating dynamics within that region.
- This focuses primarily on vertical placement, though there is also at least rudimentary support for getting rid of collisions horizontally.
- The 'score' library has also been updated with a list of all vocal instruments. This is the poor-man's version of another much larger list of ALL Finale instruments and various associated parameters including family... But this works as a jumping off point!
- Finally, there is a new script called 'dynamics_position' with various options, including:
    - Position dynamics automatically (vocal staves
    - Position dynamics below the staff
    - Position dynamics above the staff
    - Align dynamics
    - Nudge dynamics up
    - Nudge dynamics down
    
   Robert, I thought about what you said regarding not trying to treat the expressions 'equally'... But in reviewing CJ's code that's how he did it, and in running experiments with TG Tools Move/Align Dynamics, it looks like Tobias took the same approach (you can try it out yourself by changing the positioning on various dynamics... it doesn't change how TG Tools works in the slightest!). In fact, the alignment this script gives is identical to TG Tools.
   
  I know that you also have a script that essentially resets expressions to their default positions. Running that followed by running the function with the "align" mode would give the results you're looking for (I think!).

### DIFF
--- a/dist/clef_change.lua
+++ b/dist/clef_change.lua
@@ -232,6 +232,7 @@ __imports["library.clef"] = function()
 
         cell_frame_hold:ConnectCell(cell)
         if cell_frame_hold:Load() then
+
             if cell_frame_hold.IsClefList then
                 cell_clef = cell_frame_hold:CreateFirstCellClefChange().ClefIndex
             else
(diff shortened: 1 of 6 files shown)


src/library/metrics.lua
```diff
@@ -0,0 +1,754 @@
...
+    local dynamic_pos = 0
+    local system_scaling = metrics.calc_system_scaling(region:GetStartMeasure(), region:GetStartStaff())
+    local expr_above_baseline = finale.FCBaseline()
+    expr_above_baseline:LoadDefaultForMode(finale.BASELINEMODE_EXPRESSIONBELOW)
+    local baseline_pos = math.abs(expr_above_baseline:GetVerticalOffset())
+    local category_def = finale.FCCategoryDef()
+    category_def:Load(finale.DEFAULTCATID_DYNAMICS)
+    local staff_add_offset = -48 -- It's magic!
+    local entry_add_offset = -30 -- It's magic!
+
```
rpatters1, line 689: Magic numbers must be configurable.

src/library/dynamics.lua
```diff
@@ -0,0 +1,286 @@
...
+Sets the vertical position of hairpins to the input target position.
+ALSO adjusts horizontal position to avoid any analyzed expressions.
+
+@ region (FCMusicRegion) 
+@ target_pos (number) Target for hairpin vertical placement
+@ excl_zones (table) Each value is another table containing two values, zone.left and zone.right.
+]]
+function dynamics.hairpins_set_vertical_pos(region, target_pos, excl_zones)
+    local magic_vertical_offset = 10 -- "THERE'S NO SUCH THING AS MAGIC!" - Vernon Dudley
+    local marks = finale.FCSmartShapeMeasureMarks()
```
rpatters1, line 71: magic numbers must be configurable

src/library/metrics.lua
```diff
@@ -0,0 +1,754 @@
...
+Returns the y positions of the lowest and highest hairpin endpoints in the region. 
+This will work best when fed one system (FCStaffSystem) at a time.
+
+@ region (FCMusicRegion) The region to check.
+:(number) The lowest hairpin Y position in the input region.
+:(number) The highest hairpin Y position in the input region.
+]]
+function metrics.calc_lowest_highest_hairpin_pos(region)
+    local magic_vert_off = 10 -- More magic!!!
```
rpatters1, line 558: magic numbers must be configurable

rpatters1 (changes requested):
I don't see you using exp categories in a way that I disapprove of.

However, creating and freeing metrics every time you need information is *extremely* inefficient. Getting metrics can seriously slow down code. In every library routine where you create and free metrics, you should add an optional extra parameter so that the caller can provide the metrics and there is no creating or freeing. In my metrics library code you will find examples of this.

PR #359, opened 2022-09-06
Restore state for edit controls

ThistleSifter:
This PR adds full support for control state restoration in all mixin controls that are based on `FCCtrlEdit`.

I had to jump through a few hoops to make this all happen smoothly but the results should all be consistent with the native methods.


rpatters1 (approved):
This looks good to me. Are you ready for merging?

ThistleSifter:
Yes this is ready to go.

PR #361, opened 2022-09-10
Fix module state in bundled scripts

### LINKED ISSUE #360: [BUG] A new error in downloaded script using 'mixin' library
Downloaded `dist` version of `rest_offsets.lua` throws an error every SECOND time it is invoked (presumably because of RetainLuaState re-setting on the first failure).

As follows:
ERROR: rest_offsets.lua:7863:
Object is not mixin-enabled
lines 7862-7864:
```
local function catch_and_rethrow(tryfunczzz, levels, ...)
     return mixin_private.pcall_wrapper(levels, pcall(function(...) return 1, tryfunczzz(...) end, ...))
end
```

ThistleSifter:
Fixes #360 
This should fix the issue with modules not retaining their state in bundled scripts by changing `__imports` and `__import_results` to globals.

I have one question: since they're now global, do the two variable names need to be more unique or is it okay to leave them as they are?

Marking this as draft until I can work out why `pnpm` just keeps giving me errors.

### DIFF
--- a/.github/actions/bundle/dist/index.js
+++ b/.github/actions/bundle/dist/index.js
@@ -4647,8 +4647,8 @@ exports.resolveRequiredFile = exports.generateLuaRequire = void 0;
 const path_1 = __importDefault(__nccwpck_require__(1017));
 const generateLuaRequire = () => {
     return [
-        'local __imports = {}',
-        'local __import_results = {}',
+        '__imports = __imports or {}',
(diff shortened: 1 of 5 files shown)


ThistleSifter:
I got `pnpm` working. Hopefully everything has worked okay,

rpatters1 (approved):
lgtm

PR #363, opened 2022-09-15
Dynamics, Metrics Libraries

jwink75:
Dynamics vertical position functions are now working in Page and Scroll Views, including for Staff Sets.


jwink75:
Found some more BUGS

PR #364, opened 2022-09-15
Update rest_offsets.lua

cv-on-hub:
The `/dist` version of this script is still faulty, damaging the perceived reliability of the repository. This script works properly WITHOUT MIXIN until those bugs are fixed.

### DIFF
--- a/src/rest_offsets.lua
+++ b/src/rest_offsets.lua
@@ -2,8 +2,8 @@ function plugindef()
     finaleplugin.RequireSelection = true
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
     finaleplugin.AuthorURL = "http://carlvine.com/lua/"
-    finaleplugin.Version = "v1.40"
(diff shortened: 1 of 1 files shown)


ThistleSifter:
Check the `/dist` version again, the issue should hopefully be fixed now.

cv-on-hub:
Sorry, but both `rest_offsets.lua` and `layers_swap_selective.lua` both fail (freshly downloaded today). Oddly they have different problems. `offsets` seems to lose all values in the global config variable and won't retain window position. `layers` just doesn't work.

cv-on-hub:
Sorry - those errors the wrong way around. and in `rest_offsets`, the problem is at line 8583:
bad argument #1 to ‘abs’ (number expected, got nil)
This is in the main program, examining `config.layer` (integer) so that value isn’t being retained.

ThistleSifter:
Looking into this now.

cv-on-hub:
Script now functional. Closing this PR.

PR #366, opened 2022-09-16
Mixin bug fixes

ThistleSifter:
This should address the issues mentioned in https://github.com/finale-lua/lua-scripts/pull/364#issuecomment-1248890780
- Fixes a bug with `GetInteger` and `GetFloat` in `FCMCtrlEdit`
- Improves error reporting for bundled scripts

@rpatters1 Something I've noticed with a couple of methods now is that parameters that are marked as optional in the PDK are actually required when used in Lua. Is this intentional? Examples that come to mind are `FCString:GetFloat`,  `FCString:GetInteger` and `FCUI:AlertInfo`

### DIFF
--- a/src/library/mixin.lua
+++ b/src/library/mixin.lua
@@ -215,11 +215,12 @@ dialog:ExecuteModal(nil)
 local utils = require("library.utils")
 local library = require("library.general_library")
 
--- mixin_public: public methods and mixin constructors, stored separately to keep the mixin namespace read-only.
--- mixin_private: private methods
--- mixin_classes: fully resolved FCM and FCX mixin class definitions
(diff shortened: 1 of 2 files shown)


rpatters1:
The PDK Framework is a C/C++ framework, so optional parameters are only optional in C/C++. There is no way (that I've found) to bridge them to Lua as optional and have them retain their default values. This may be a LuaBridge limitation. I haven't tried to convert the bridge code to some other option.

FWIW: Booleans are always optional in Lua even when they are *not* optional in the Framework. But omitted booleans always get translated to false, which has been a challenge at least one. (With one of the constructor functions that isn't coming to mind at the moment.)

rpatters1 (approved):

PR #367, opened 2022-09-20
Restore state for FCCtrlStatic mixins

ThistleSifter:
A few small changes in this PR:
- Restores `TextColor` value for `FCCtrlStatic` mixins (according to my testing this isn't needed but the safety net is there anyway)
- `FCMCtrlStatic.SetTextColor` triggers an immediate visual change (gets and sets the text again under the hood)
- Some indentation fixes and better parameter names in `FCXCtrlStatic`


ThistleSifter:
Also added a small change for a quick fix in the `mixin` library.

Nick-Mazuk (approved):

PR #372, opened 2022-10-17
Update hairpin_creator.lua

cv-on-hub:
In response to user request, suppressed hairpin creation in empty regions and fixed pre-existing hairpin erasure in final measure.


Nick-Mazuk (approved):
Looks good to me

PR #373, opened 2022-10-19
Create noteheads_crossed.lua

cv-on-hub:
I just needed to make this little script to convert all noteheads in the selection to crosses. Thought it might be useful to others.


rpatters1:
Erasing the `FCNoteheadMod` will erase percent changes and custom font changes as well. I would suggest that's not the right thing to do, but maybe there's a reason for it?

I would like to see this use the `notehead.change_shape` function in `library/notehead.lua`. You would need to add the new notehead type, with noteheads for every duration (similar to what is there now for "diamond".)

Also, does "cross" actually mean "x-shaped"? If so, I think we should use Finale's terminology.

cv-on-hub:
Thanks Robert - good points all. I erased `FCNoteheadMod` because the notehead.lua routine does that (along with several earlier notehead changing scripts). Should the library file be changed to also omit erasure? Omitting erasure doesn't seem to otherwise affect the operation. 

Should the library include a `notehead.change_shape_number` routine? This could be more useful than multiple or "switched" routines for the standard shapes.

Forgot about "x". Script name: notehead_change_x.lua (?)

rpatters1:
I'd say let's leave the erasure in for now. `notehead.lua` is currently designed as an if-elseif-else type switch. It would need to be refactored at some point. If you want to take that on, be my guest.

cv-on-hub:
I was thinking of giving `notehead.lua` a shot but JW_Change lists thirteen "standard" notehead possibilities, and all of these may have distinct parameters to be saved/modified in `notehead.config.txt` with almost-magic snake_style text IDs, making it a much bigger job than I'm interested in. 

Will come back with a revised version of this `notehead_change_x` script.

cv-on-hub:
Just realised that under the old configuration setup there was no test for `diamond` as a valid shape choice. 

If `diamond` shape creates diamond noteheads, shouldn't the config point to `config.diamond` and not to `config.diamond_open`? Then if you want `diamond_closed` shapes the config points correctly to `config.diamond_closed`.

And isn't the real distinction here between `diamond_harmonics` and `diamond_rhythmic`? `diamond_harmonics` should always have an open diamond notehead even on short notes.

JW suggests 3 types of diamond notehead:

- Filled Diamond
- Hollow Diamond (Harmonics)
- Hollow Diamond (Rhythmic Notation) - bigger

(and now I remember why I didn't want to tackle `notehead.lua` as a complete issue!)

jwink75:
Hey everyone… Carl in particular :) …

I have all sorts of notehead replacement functions in JetStream that originated with CJ. I’m swamped right now, but seeing as how I already have the characters mapped in both SMuFL and non-SMuFL I’d be happy to try to build this stuff out at a later date!

cv-on-hub:
That sounds great @jwink75. I'll consolidate what I have now for you to fill out later.

cv-on-hub:
The attached .musx file is handy for testing changed noteheads.
[notehead_test.musx.zip](https://github.com/finale-lua/lua-scripts/files/9845349/notehead_test.musx.zip)

cv-on-hub:
Sorry - got completely obsessed by this. Have now added multiple shape "families" to the library file plus the ability to specify a single numeric shape instead of the family name. Happy for others to add/change the glyph choices at will. Following is the new `noteheads_change.lua` script to access all of this including simple numeric entry.

cv-on-hub:
Adding a family of "wedge" noteheads to library and "change" script

cv-on-hub:
This strikes me as a neater, more consistent and versatile setup for `notehead` config:
```
config = {
    shape_name1 = {
        quarter = { glyph = xxx, size = 100, offset = 0 },
        half = { glyph = xxx, size = 100, offset = 0 },
        whole = { glyph = xxx, size = 100, offset = 0 },
        breve = { glyph = xxx, size = 100, offset = 0 },
    },
    shape_name2 = { ...
    },
}
```
Is this a sufficiently economical format within Lua? It would make the component assembly switch uniform and build in a nul test for acceptable shape name.
`config[shape_name][duration][data_type]`

rpatters1:
Lol, this is the refactoring I was talking about. Thank you for taking it on. :-)

cv-on-hub:
That wasn't as hard as I anticipated! Coming through now ... revised library file and activator script.

cv-on-hub:
Adding two more shapes to the family: `Strikethrough` and `Circled`

src/library/notehead.lua
```diff
@@ -1,64 +1,261 @@
...
+        breve = { glyph = 202 },
+    },
+    default = {
+        quarter = { glyph = 207 }
+    },
 }
 
--- Default to SMuFL characters for SMuFL font (without needing a config file)
+-- change to SMuFL characters for SMuFL font (without needing a config file)
 if library.is_font_smufl_font() then
```
Nick-Mazuk, line 108: Let's wrap this in a function. Perhaps define both the non-SMuFL and SMuFL configs in the top-level, then have a function that returns the correct one depending on `library.is_font_smufl_font()`.

As it stands, this can introduce a subtle bug. If a script changes the font then reads noteheads, the wrong config will be used. However, if we wrap it in a function, it'll allow us to choose the correct config precisely at the time we'll need it.

src/noteheads_change.lua
```diff
@@ -0,0 +1,139 @@
...
+function user_chooses_glyph()
+    local dlg = finale.FCCustomWindow()
+    local x, y = 200, 10
+    local y_diff = finenv.UI():IsOnMac() and 3 or 0 -- extra y-offset for Mac text box
+    local str = finale.FCString()
+    str.LuaString = plugindef()
+    dlg:SetTitle(str)
+
+    str.LuaString = "Enter required character (glyph) number:"
+    _ = dlg:CreateStatic(0, y)
```
Nick-Mazuk, line 100: Why is the name of this variable an underscore? Usually variables named as an underscore represent variables that are never going to be used, but this one definitely is used later.

Also, why is not not defined as a local variable? I don't find any other references to this variable in this PR that require it to be a global variable. Perhaps a static window needs to be a global variable? It's been a while since I've done Lua code so I could easily be forgetting something.

Nick-Mazuk (approved):

src/noteheads_change.lua
cv-on-hub: Sorry @Nick-Mazuk. I thought I'd seen `_` used this way elsewhere and assumed it was automatically local and discardable, like `$_` in Perl. Now fixed.

Nick-Mazuk: Yeah, some languages discard it, others don't. I believe Lua is a language that doesn't discard it.

cv-on-hub:
Added `round` noteheads, discovered some better matching SMuFL shapes for others and minor tweaks to both `notehead.lua` and `noteheads_change.lua` following.

Nick-Mazuk (approved):

rpatters1:
Is this PR complete? We seem to get additions every day. (:lol).

cv-on-hub:
> We seem to get additions every day. (:lol).

Well ... every OTHER day. No - that's it for now. I can't see any obvious remnant bugs and I'll leave it up to @jwink75 to verify the character mappings.

cv-on-hub:
I closed this PR but then discovered that neither `notehead.lua` nor `noteheads_change.lua` have been updated in the repo.

rpatters1 (approved):

PR #378, opened 2022-11-13
Update staff_explode_layers.lua

cv-on-hub:
Noticed that there was no test for (the unlikely) event of layer number > 4. Also now employs `library.layer`.

### DIFF
--- a/src/library/layer.lua
+++ b/src/library/layer.lua
@@ -102,4 +102,16 @@ function layer.swap(region, swap_a, swap_b)
     end
 end
 
+
+--[[
+% max_layers
+
+Return the maximum number of layers available in the current document.
+
+: (number) maximum number of available layers
+]]
+function layer.max_layers()
+    return finale.FCLayerPrefs.GetMaxLayers and finale.FCLayerPrefs.GetMaxLayers() or 4
+end
+
 return layer
--- a/src/staff_explode_layers.lua
+++ b/src/staff_explode_layers.lua
@@ -1,10 +1,10 @@
 function plugindef()
     finaleplugin.RequireSelection = true
     finaleplugin.Author = "Carl Vine"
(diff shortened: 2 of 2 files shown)


rpatters1:
Please use `FCLayerPrefs::GetMaxLayers()` instead of hard-coding `4`. Since `GetMaxLayers` was only added in 0.62, perhaps there should be a function added to the layer library that returns:

```lua
return FCLayerPrefs.GetMaxLayers and FCLayerPrefs.GetMaxLayers() or 4
```

cv-on-hub:
> Please use `FCLayerPrefs::GetMaxLayers()` instead of hard-coding `4`

Ah - interesting! If min version is 0.62 then isn't a specific MaxLayers() function in `layer.lua` redundant? I also see that `tie.lua`uses `finale.FCLayerPrefs.GetMaxLayers()` without testing for RGP version. Is that a problem?

cv-on-hub:
Ah - interesting! Here follows a revised version of `layer.lua` adding your suggested `max_layers()` routine, along with `staff_explode_layers.lua` updated to use it.

rpatters1:
If your min version is 0.62 then a special func isn't strictly necessary. But it will be useful for other scripts.

`tie.lua` uses `FCTieMod` which also requires (at least) 0.62, so any code calling `tie.lua` must have the minimum version set anyway.

rpatters1:
Instead of passing in max_layers, could you not just get it one in a local variable at the top? That or else call the function directly each time (my preference). It's not like the function itself has any performance hit.

cv-on-hub:
Good to know about `tie.lua` - all news to me! I think I've done what you suggest in `staff_explode_layers.lua` following. I assume the revised `layer.lua` is OK.

rpatters1 (approved):
lgtm

Issue #389, opened 2022-12-30
Floating EVPU in dialog:CreateMeasurementEdit


cv-on-hub:
I guess this is mostly addressed to @ThistleSifter.

I'm working on a script using your CreateMeasurementUnitPopup in a FCXCustomLuaWindow, and the EVPU values in MeasurementEdits always revert to integers. How can I force EVPUs to be floats? (Line widths are either EFIX or decimal precision).

ThistleSifter:
`FCXMeasurementEdit` has three methods that change the underlying format used when converting between measurement units:
- `SetTypeMeasurementInteger()` - Whole EVPUs (this is the default)
- `SetTypeMeasurement()` - EVPUs with decimal
- `SetTypeMeasurementEfix()` - EFIXs with decimal

It sounds like the one you're after is `SetTypeMeasurement()`

cv-on-hub:
Ah - many thanks @ThistleSifter - that does the trick! Is it usual to get significant rounding errors when swapping Units? 1.75 EVPU, for instance, drops to 1.72 when swapping to points and back. (Script attached if you're interested). 
[gracenote_slash.txt](https://github.com/finale-lua/lua-scripts/files/10326310/gracenote_slash.txt)

ThistleSifter:
Hmm, I'll have to look into that further. In theory it shouldn't, at least not with the example figure you gave.

cv-on-hub:
Without any user changes, these values in EVPU:
```
measurement_unit = 1
upstem_artic_y_offset = -32
downstem_artic_y_offset = 8
upstem_artic_x_offset = 4
downstem_line_width = 2.25
```
get stored in the User prefs file as this in cm:
```
measurement_unit = 3
upstem_artic_y_offset = -32.003746032715
downstem_artic_y_offset = 8.0003700256348
upstem_artic_x_offset = 3.9996180534363
downstem_line_width = 2.2498559951782
```
Converting back to EVPU shows the rounding errors in the prefs file:
```
upstem_artic_y_offset = -32
downstem_artic_y_offset = 8
upstem_artic_x_offset = 3.9900000095367
downstem_line_width = 2.2400000095367
```

ThistleSifter:
I was looking into this just now and I noticed that Finale's inbuilt dialogs have the same issue.

For example, open _Document -> Document Options... ->Piano Braces and Brackets_ and play around with changing measurement units.

cv-on-hub:
[...]

PR #390, opened 2022-12-31
Create gracenote_slash.lua

cv-on-hub:
New script that duplicates Jari Williamsson's original JWGraceNoteSlash plug-in. **Happy New Year!**

### DIFF
--- a/src/gracenote_slash.lua
+++ b/src/gracenote_slash.lua
@@ -1,0 +1,174 @@
+function plugindef()
+    finaleplugin.RequireSelection = true
+    finaleplugin.Author = "Carl Vine"
+    finaleplugin.AuthorURL = "http://carlvine.com/lua/"
+    finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
+    finaleplugin.Version = "v0.56"
+    finaleplugin.Date = "2022/12/30"
+    finaleplugin.AdditionalMenuOptions = [[
+        Gracenote Slash Configuration...
+    ]]
+    finaleplugin.AdditionalUndoText = [[
+        Gracenote Slash Configuration
+    ]]
+    finaleplugin.AdditionalDescriptions = [[
+        Configure Gracenote Slash parameters
+    ]]
+    finaleplugin.AdditionalPrefixes = [[
+        slash_configure = true
+    ]]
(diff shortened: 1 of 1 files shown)


ThistleSifter:
Could you move the call to `SetTypeMeasurement` so that it comes before `SetMeasurement`? ie `SetTypeMeasurement():SetMeasurement(config[name])`

I've been brainstorming some solutions for #389 and if I end up going with my current idea, the order will make a difference.

(Incidentally, I would have written them in the order I suggested above regardless because `SetTypeMeasurement` is a setup method, and I think in general it's better form (and safer) to compete setup before feeding in actual data.)

cv-on-hub:
> Could you move the call to `SetTypeMeasurement`

Sure - easy to do. Doesn't (for now, naturally) alter the rounding error problem!

rpatters1:
This looks okay to me, but should we not hold off until the rounding problem is fixed?

cv-on-hub:
> hold off until the rounding problem is fixed?

Agreed.

cv-on-hub:
As Robert pointed out in Issue #394 the majority of parameter measurements should be integer EVPUs, so in this script only the EFIX value for LineWidth demands decimal EVPUs. In this revision I've returned the other parameters to `TypeMeasurementInteger`. Not sure if the discrepancy in unit accuracy needs visual clarification for the user.

ThistleSifter:
Finale doesn't offer any visual distinction between the different types in its dialogs. Not that I think that's good UI (it's confusing when entered values are rounded in seemingly random ways), but if you were to just leave it, at least it would be consistent.

If you were to distinguish them it would probably be with a text label of some sort, although I'm not sure what a clean solution would look like without confusing the user further ("What on earth is an EFIX??? I just want slashes on my grace notes, why do I need to know all this extra stuff?").

cv-on-hub:
Thanks for the response @ThistleSifter - yes the UI is muddled. **Smart Shape Options...** for instance has integers and decimals alongside each other with the actual stored (integer) value only confirmed when the dialog reopens. In which case I'll stick to the new version with `LineWidth` decimal and the rest integer, just like papa Finale.

rpatters1:
Is this PR ready to go?

cv-on-hub:
> Is this PR ready to go?

I believe so!

PR #391, opened 2023-01-07
Update delete_selective.lua

cv-on-hub:
Added selective deletion of Lyrics and improved identification of Dynamics.

### DIFF
--- a/src/delete_selective.lua
+++ b/src/delete_selective.lua
@@ -3,14 +3,15 @@ function plugindef()
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "http://carlvine.com/lua/"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "0.56"
-    finaleplugin.Date = "2022/08/22"
+    finaleplugin.Version = "0.57"
+    finaleplugin.Date = "2023/01/07"
     finaleplugin.MinJWLuaVersion = 0.62
 	finaleplugin.AdditionalMenuOptions = [[
         Delete Dynamics
         Delete Expressions (Not Dynamics)
         Delete Expressions (Measure-Attached)
         Delete Articulations
+        Delete Lyrics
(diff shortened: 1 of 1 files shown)


src/delete_selective.lua
```diff
@@ -68,14 +72,27 @@ function plugindef()
...
 end
 
 delete_type = delete_type or "expression_all"
 
+function expression_is_dynamic(exp)
+    if not exp:IsShape() and exp.Visible and exp.StaffGroupID == 0 then
+        local cd = finale.FCCategoryDef()
+        local text_def = exp:CreateTextExpressionDef()
+        if text_def and cd:Load(text_def.CategoryID) then
+            if text_def.CategoryID == finale.DEFAULTCATID_DYNAMICS or string.find(cd:CreateName().LuaString, "Dynamic") then
```
rpatters1, line 88: Use `cd.Type` instead of `text_def.CategoryID`. This calculation happens several times throughout the repository. I'm thinking it should be a library function in the expression library. A Lua-version-independent approach would be:

```lua
if text_def.CategoryID == finale.DEFAULTCATID_DYNAMICS then
   return true
end
if cd.Type == finale.DEFAULTCATID_DYNAMICS then
   return true
end
if string.find(cd:CreateName().LuaString, "Dynamic") then
   return true
end
return false
```

However, this script cannot run a version of FinaleLua that lacks `cd.Type`, so the first check is not needed here.

rpatters1 (changes requested):

src/delete_selective.lua
cv-on-hub: > Use `cd.Type` instead of `text_def.CategoryID`.

Many thanks Robert. Correction follows.

rpatters1: In the modified version, `cd` is not loaded. I think the definition has a `Create...` method to do this.

cv-on-hub: Oops. Thanks for the eagle eyes.

rpatters1 (approved):

PR #392, opened 2023-01-07
Update notehead.lua

cv-on-hub:
Suggesting change half-note diamond head to a larger size SMuFL character. Not sure how others feel about this.

Gould states all harmonic diamond heads should be the same size regardless of duration (indicate duration above the staff). But there's a really nice larger diamond in SMuFL to differentiate from the quarter. I'd already given whole and breve notes noticeably different characters.

### DIFF
--- a/src/library/notehead.lua
+++ b/src/library/notehead.lua
@@ -115,7 +115,7 @@ if library.is_font_smufl_font() then
     config = {
         diamond = {
             quarter = { glyph = 0xe0e1, size = 110 },
-            half  = { glyph = 0xe0e1, size = 110 },
+            half  = { glyph = 0xe0da, size = 110 }, -- or "0xe0e1" to match quarter notehead
             whole = { glyph = 0xe0d8, size = 110 },
             breve = { glyph = 0xe0d7, size = 110 },
         },


rpatters1:
My preference is that by default it should follow Gould's recommendation. That is, all note values less than a whole note use the same white diamond notehead. If you want a different notehead for half noteheads (which personally I very decidedly do not), use a config file. That's what it is for.

In fact, I'm not sure if the diamond noteheads should be different for whole noteheads either. I can't find an example in Gould of one, other than p. 417 which seems like it is something different. (Longs, breves, etc., I don't care about). But since it's different already, I don't advocate changing it.

cv-on-hub:
Elaine would be pleased with your faith! 

There's a slight problem with config files for `notehead.lua` since it uses nested table values like `diamond.half.glyph  = 0xe0da`. AFAICT `configuration.lua` can't parse nested table values.

rpatters1:
"Behind Bars" is one of 3 books that should be found on the desk of any composer or arranger. It superseded Ted Ross "The Art of Music Engraving and Processing" when it came out, which was no mean feat. (As far as I can tell, it mostly incorporates rather than contradicting the useful information in Ross.) It is a distillation of decades of learned wisdom from the music engraving industry, and its recommendations should never be discarded lightly. Ross and Gould seem to be the only books (at least in English) that address engraving minutia at that level of detail. If you know of others, do tell. Finale relied heavily on Ross in its initial development, and I believe in later versions they have switched to Gould as a reference.

FWIW: the other 2 books are Andrew Stiller "Handbook of Instrumentation" and Kurt Stone "Music Notation in the Twentieth Century". (Both of these could stand to have an update, similar to Gould's update of Ross, but I am not aware of any newer books that are better than they are for what they are.) Gould is for engraving recommendations, Stone is for recommendations on how to notate non-standard performance practice, and Stiller is for information about how instruments work. 

I just noticed that defect in config about the nested tables yesterday. I'm working on a pull request with some bug fixes, and I'll include support for dot-nested tables on the input side of config tables. (Probably won't be adding them to the output side.)

rpatters1:
I have implemented a change in `configuration.lua` that will allow either of these:

```lua
diamond.quarter.glyph = 226
```

or

```lua
diamond.quarter = { glyph = 0xe0e2, size = 100 }
```

Look for it in a bug fix PR coming soon.

rpatters1:
I am closing this PR in favor of #393.

PR #393, opened 2023-01-08
Rgp/bug fixes

rpatters1:
Addresses the following issues:

- standalone hairpins script now supports reduced staff sizes
- configuration files now support parameter names with dot-syntax for nested tables
- configuration files now support any Lua syntax for table-valued parameter values, including nested tables.

Potentially we could support any Lua syntax for any parameter value, but as of now this PR is restricted to table-valued parameters. This is limit error risk.

### DIFF
--- a/src/pitch_transform_harmonics_fourth.lua
+++ b/src/pitch_transform_harmonics_fourth.lua
@@ -12,6 +12,7 @@ end
 
 local articulation = require("library.articulation")
 local transposition = require("library.transposition")
+require('mobdebug').start()
 local notehead = require("library.notehead")
 local note_entry = require("library.note_entry")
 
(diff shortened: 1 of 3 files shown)


src/pitch_transform_harmonics_fourth.lua
```diff
@@ -12,6 +12,7 @@ end
 
 local articulation = require("library.articulation")
 local transposition = require("library.transposition")
+require('mobdebug').start()
```
Nick-Mazuk, line 15: Do we need to include this?

Nick-Mazuk (approved):

cv-on-hub:
My compliments on the concision of the recursion in `process_table()`. Couldn't work out how to process it myself!

src/pitch_transform_harmonics_fourth.lua
rpatters1: no, I'll remove it. Thanks

PR #394, opened 2023-01-10
Make FCXCtrlMeasurementEdit value persist through measurement unit changes

### LINKED ISSUE #389: Floating EVPU in dialog:CreateMeasurementEdit

ThistleSifter:
This should fix #389 and in the process, provide more consistent behaviour than Finale's own dialogs.

Note that setting the type does now affect the stored value, so the type should be set before setting the value.
For example:
```lua
-- Default type is MeasurementInteger
local edit = dialog:CreateMeasurementEdit(0, 0)

-- Because type is MeasurementInteger, value will be rounded
edit:SetMeasurement(12.5)

edit:GetMeasurement() -- Returns 13

-- Now change the type
edit:SetTypeMeasurement()

edit:GetMeasurement() -- Still returns 13

edit:SetMeasurement(12.5)

edit:GetMeasurement() -- Now it returns 12.5
```

The only question I have is whether in the example above, the value after the first call to `SetMeasurement`  should be 12 or 13.  ie should it be rounded to the nearest whole or rounded down. I thought the nearest whole seemed more appropriate.

Here's a little script for testing output values:
```lua

dialog:CreateMeasurementEdit(0, 0, "edit"):SetTypeMeasurement()--:SetMeasurement(12.4)
dialog:CreateButton(0, 40):SetText("Get Value"):AddHandleCommand(function(self)
    finenv.UI():AlertInfo(self:GetParent():GetControl("edit"):GetMeasurement(), "Measurement Edit Value")
end)
dialog:CreateMeasurementUnitPopup(0, 80)
dialog:ExecuteModal()
```

@rpatters1 Do you think it's worth trying to resolve the floating point inaccuracy issues from the `GetMeasurement*` methods in `FCXCtrlMeasurementEdit`?
Here's what I mean:
```lua
edit:SetMeasurement(12.4)
edit:GetMeasurement() -- returns 12.4

-- But if the user was to type "12.4" in the box...
edit:GetMeasurement() -- returns 12.39999961853
```

### DIFF
--- a/src/mixin/FCXCtrlMeasurementEdit.lua
+++ b/src/mixin/FCXCtrlMeasurementEdit.lua
@@ -5,13 +5,16 @@ $module FCXCtrlMeasurementEdit
 
 *Extends `FCMCtrlEdit`*
 
+_Note that the type should be set **before** setting any values._
+
 Summary of modifications:
-- Parent window must be an instance of `FCXCustomLuaWindow`
-- Displayed measurement unit will be automatically updated with the parent window
-- Measurement edits can be set to one of three types which correspond to the `GetMeasurement*`, `SetMeasurement*` and *GetRangeMeasurement*` methods. The type affects which methods are used for changing measurement units, for events, and for interacting with an `FCXCtrlUpDown` control.
+- Parent window must be an instance of `FCXCustomLuaWindow`.
+- Displayed measurement unit will be automatically updated with the parent window.
+- Measurement edits can be set to one of four types which correspond to the `GetMeasurement*`, `SetMeasurement*` and *GetRangeMeasurement*` methods. The type affects which methods are used for changing measurement units, for events, and for interacting with an `FCXCtrlUpDown` control.
 - All measurement get and set methods no longer accept a measurement unit as this is taken from the parent window.
+- Added measures to prevent underlying value from changing when the measurement unit is changed.
 - `Change` event has been overridden to pass a measurement.
-- Added hooks for restoring control state
+- Added hooks into control state restoration
 ]] --
 local mixin = require("library.mixin")
 local mixin_helper = require("library.mixin_helper")
@@ -22,6 +25,36 @@ local props = {MixinParent = "FCMCtrlEdit"}
 
 local trigger_change
 local each_last_change
+
+-- Converts a measurement value from one type (Measurement, MeasurementInteger, MeasurementEfix, Measurement10000th) to another
+local function convert_type(value, from, to)
+    -- Sanitise all integer types
+    if from ~= "Measurement" then
+        value = utils.round(value)
+    end
+
+    if from == to then
+        return value
+    end
+
+    if from == "MeasurementEfix" then
+        value = value / 64
+    elseif from == "Measurement10000th" then
+        value = value / 10000
+    end
+
+    if to == "MeasurementEfix" then
+        value = value * 64
+    elseif to == "Measurement10000th" then
+        value = value * 10000
+    end
+
+    if to == "Measurement" then
+        return value
+    end
+
+    return utils.round(value)
+end
 
 --[[
 % Init
@@ -37,6 +70,8 @@ function props:Init()
     private[self] = private[self] or {
         Type = "MeasurementInteger",
         LastMeasurementUnit = parent:GetMeasurementUnit(),
+        LastText = mixin.FCMCtrlEdit.GetText(self),
+        Value = mixin.FCMCtrlEdit.GetMeasurementInteger(self, parent:GetMeasurementUnit()),
     }
 end
 
@@ -45,7 +80,6 @@ end
 
 **[Fluid] [Override]**
 Ensures that the overridden `Change` event is triggered.
-
 
 @ self (FCXCtrlMeasurementEdit)
 @ str (FCString|string|number)
@@ -85,6 +119,19 @@ for method, valid_types in pairs({
 end
 
 --[[
+% GetType
+
+Returns the measurement edit's type. The result can also be appended to `"Get"`, `"GetRange"`, or `"Set"` to use type-specific methods.
+The default type is `"MeasurementInteger"`.
+
+@ self (FCXCtrlMeasurementEdit)
+: (string) `"Measurement"`, `"MeasurementInteger"`, `"MeasurementEfix"`, or `"Measurement10000th"`
+]]
+function props:GetType()
+    return private[self].Type
+end
+
+--[[
 % GetMeasurement
 
 **[Override]**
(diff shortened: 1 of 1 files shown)


rpatters1:
In my opinion, everything should tie back to integer EVPU or integer EFIX, because internally those are almost all Finale cares about for distance measurement. (It's mostly EVPU, but the EFIX fields are documented. EFIX = 1/60 EVPU.) There may be an oddball exception here and there, but I don't recall any now if so.

Percents are a different matter. As are EDU beat positions. But the same principle applies to them as well. (But are they outside the scope of this PR?)

If setting the value to 12.4 and setting it to 12.39999961853 results in the same EVPU or EFIX value then, no, it isn't worth addressing.

The same issue applies to rounding. The same EVPU or EFIX value should be the result. In fact, I am a little concerned about "providing more consistent behavior than Finale", because I wonder if the same EVPU/EFIX value should always produce the same measurement value in both our dialogs and Finale's. But as long as our dialog's more consistent values produce the same EVPU values internally, it's probably okay that they look better than Finale's.

ThistleSifter:
My comment about "more consistent behaviour" was referring to what happens when you change measurement units. If you type something in one of Finale's dialogs that supports EVPUs with decimals, the more times you change measurement unit, the more the value drifts away from what you originally typed. That's what this PR addresses.

The other stuff with improving rounding when setting programmatically was a side effect of the way I handled the above issue. The reason they "look better" is just because of differences in how C++ and Lua handle floats.

It doesn't have an effect on integer EVPUs/EFIXs but cleaning it up might make someone's config file look nicer, for example.

rpatters1:
I'm fine with anything as long as Finale receives the correct EVPU value. Is this ready to go?

ThistleSifter:
After pondering your comments, I think I'll do a little more testing and research first.

ThistleSifter:
@rpatters1 I found an issue. `FCString.GetMeasurement` and `FCCtrlEdit.GetMeasurement` produce different results. Here's a little test script:

```lua
local str = finale.FCString()

local static = dialog:CreateStatic(0, 3)
static:SetWidth(200)
str.LuaString = "Enter a value in EVPUs:"
static:SetText(str)

local edit = dialog:CreateEdit(120, 0)

local btn = dialog:CreateButton(0, 40)
btn:SetWidth(200)
str.LuaString = "Check Value"
btn:SetText(str)

dialog:RegisterHandleControlEvent(btn, function()
    edit:GetText(str)
    finenv.UI():AlertInfo("FCString: " .. str:GetMeasurement(finale.MEASUREMENTUNIT_EVPUS) .. "\n" .. "FCCtrlEdit: " .. edit:GetMeasurement(finale.MEASUREMENTUNIT_EVPUS), "Measurement Values")
end)

dialog:ExecuteModal(nil)
```

Try entering something like `12.4` into the text box and see what comes up. Can this be fixed (and for `GetMeasurementEfix` as well)?

ThistleSifter:
Looking into the bug further, it appears that `FCCtrlEdit.GetMeasurement` only returns the correct value while the window is showing. If it is called before the window has been shown, or after it has been closed (eg like the old way of programming modal dialogs), then it returns the same as `FCString.GetMeasurement` in the test script above.

rpatters1:
Here is the code for `FCCtrlEdit.GetMeasurement`:

```c++
double FCCtrlEdit::GetMeasurement(_ENUMCODE(MEASUREMENTUNITS) _LUACODE(twobyte) measurementunit)
{
    if (!WindowExists()) return nonwindow_store.text.value.GetMeasurement(measurementunit);

    double dTemp;
#if OPERATING_SYSTEM == MAC_OS
    /* Mac - Cocoa */
    _FCWindowController_cocoa* pCocoaLink = _GetCocoaLink();
    if (!pCocoaLink) return 0;
    FCString string;
    pCocoaLink->GetDlgItemText(GetControlID(), &string);
    dTemp = string.GetMeasurement(measurementunit);
#else
    /* Windows */
    FX_GetDlgItemUser(GetWindowHandle(), GetControlID(), measurementunit, &dTemp);
#endif
    return dTemp;
}
```

Based on this, there should be no discrepancy on Mac, only on Win. `FX_GetDlgItemUser` is a PDK call, but it can't be called if there is no dialog box. In short, I don't think a C++ fix is worth the effort. But if it can be fixed in the mixin I'm all for it.

ThistleSifter:
Hmm, what about `FCString.GetMeasurement`? Is a C++ fix possible for that?

ThistleSifter:
Does that mean that on Mac, typing in "12.4" gives `12.4` or `12.3999...`?

rpatters1:
On Mac it gives `12.39999...`.

A C++ fix is dicey because it is a `float`. A Lua fix could include a precision and round to that precision. Also, a C++ is dependent on the next RGP Lua version and when that might be released. (Not before the next Finale release, whenever that might be, unless there is a major problem.)

ThistleSifter:
Okay, I'll look into it. On the surface it looks like the whole process of converting between measurement units might need to be rewritten in Lua for this to work but I'll see what I come up with.

rpatters1:
The problem appears to be that the C++ code is using `float` instead of `double`. So precision is getting lost. I think I will fix the C++ code, but like I said, it may be a while before it is released.

rpatters1:
P.S. Perhaps a heroic Lua effort is not needed.

rpatters1:
Is this ready to go? Given that such an easy fix is available in C++ (which I have implemented for the next release), perhaps a massive further effort is not needed in Lua?

ThistleSifter:
I've got a fix up and running (I thought it would be good to have a fix for all versions of JW/RGPLua), I just need to finish writing comments and adding overrides for the other methods. The actual fix was quite easy to write, I just had difficulty finding time to spend testing it.

ThistleSifter:
This should be good to go now, although I would be grateful if @cv-on-hub could give this a whirl.

cv-on-hub:
LGTM. The only thing I note is that EVPUs are returned with max 2 decimal points. This is reasonable and a suitable level of accuracy for working with EFIXs, and otherwise the conversions seem more accurate than in native Finale.

ThistleSifter:
That's odd, I'll look into it.

ThistleSifter:
Was this using the test script from above or something else? What values were you entering / which units?

cv-on-hub:
I was using my gracenote_slash script: https://github.com/finale-lua/lua-scripts/blob/master/src/gracenote_slash.lua
which sets `line_width` to `CreateMeasurementEdit:SetTypeMeasurement()`
It always returns 2 decimal EVPU points on every conversion, unlike your test, and these are the values being saved to the user prefs file.

cv-on-hub:
PS:
0.645 pt -> 2.58 EVPU -> 0.64 pt
0.66999 pt -> 2.67 EVPU -> 0.67 pt
2.6999 EVPU -> 0.67 pt -> 2.69 EVPU

The rounding errors are reasonable, but every decimal returns with 2-point accuracy.

cv-on-hub:
PPS: 2.6999 EVPU is stored in the prefs file with 4-points, but returns in the dialog as 2-points after configuration.get_user_settings()

cv-on-hub:
Ah ... it's a display issue. The value is retained with multiple decimal points but only displays 2 points at `dialog.CreateMeasurementEdit(...):SetTypeMeasurement():SetMeasurement(n)`

ThistleSifter:
Ah, it looks like I will need to add `SetMeasurement` to `FCMString` after all. I've got it written up but I haven't tested it. I'll see if I can do it tonight or tomorrow.

Thanks for testing it.

rpatters1:
Is this ready to be merged? It looks good to me.

ThistleSifter:
Yes it's good to go.

PR #397, opened 2023-01-11
Split Layers

rpatters1:
This script is similar to Carl Vine's Explode to Layers script, but instead of exploding it splits chords into 2 layers based on a split point in the chord. In the process I made some modifications to libary functions as well:

- `layer.copy()` now optionally clones articulations to the target layer
- `note_entry.make_rest()` is a new function that calls `FCNoteEntry::MakeRest` but does all the necessary cleanup first. I have not changed any existing code to use it because as far as I can tell it isn't needed in the context in which the existing code is calling `FCNoteEntry::MakeRest`.

### DIFF
--- a/src/library/layer.lua
+++ b/src/library/layer.lua
@@ -11,8 +11,9 @@ Duplicates the notes from the source layer to the destination. The source layer 
 @ region (FCMusicRegion) the region to be copied
 @ source_layer (number) the number (1-4) of the layer to duplicate
 @ destination_layer (number) the number (1-4) of the layer to be copied to
+@ [clone_articulations] (boolean) if true, clone articulations (default is false)
 ]]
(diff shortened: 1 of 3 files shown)


cv-on-hub:
layer.copy(... clone_articulations) is a great idea. Otherwise, although instructions in the script are easy to follow, I find the nomenclature in the dialog confusing: "Split After (From Top)". Instead perhaps: "Split at (# notes from the top)". Rest of the script is of course fine.

rpatters1:
I made a change along the line you suggested. Let me know if it helps.

cv-on-hub:
Yes that's good. Just became aware that `layer.copy()` doesn't retain notehead changes - possibly a change, including many other potential NoteMods, for another day! (Pandora's box opened by `articulations`).

PR #398, opened 2023-01-16
Autoposition Rolled Chords issue

rpatters1:
fix issue that erroneously hid articulations if they weren't on the top selected staff


rpatters1:
This are minor fixes, so I'm going to go ahead an merge them before the repo gets too far ahead.

PR #401, opened 2023-01-17
Update expression.lua

cv-on-hub:
I think it was Robert's suggestion to incorporate `expression.is_dynamic()` into the `expression` library


rpatters1:
This looks good as a library function. Should we not also convert all the homegrown versions of it in the repo to use it?

cv-on-hub:
> convert all the homegrown versions of it in the repo

Only two of my scripts use the function which I've amended (locally) to use the new library function.

rpatters1:
Ok, I will update the others.

PR #402, opened 2023-01-18
Fix rounding issues in FCCtrlEdit and FCString measurement methods

ThistleSifter:
This PR fixes the measurement rounding issues that are partly causing the problems described in #389.

Here are the changes made:
- Rewrote `GetMeasurement` method from scratch in `FCMString` to solve the rounding problem. I also emulated the handling of overrides from `FCCtrlEdit` on Windows, which is more comprehensive than what `FCString` does (eg handling Picas overrides correctly).
- Changed the rounding of `minimum` in `GetMeasurementInteger` so that if decimal values are passed, they are handled correctly instead of being discarded (if I want an integer between `12.5` and `20` and I give it `11`, I should get `13` in return, not `12`).
- `GetMeasurementEfix` now returns whole EFIXes instead of decimals, rounded correctly (which is really what it should've done in the first place, just like `GetMeasurementInteger` does with EVPUs). My reason for doing this is that if a value like `63.75` is returned and then used on an EFIX field the decimals will be discarded, meaning that the actual value saved will be `63` instead of `64`. By returning an already rounded value, the script author doesn't need to think about handling this.
- Added `*Measurement10000th` methods for getting and setting values in 10,000ths of an EVPU, again mainly so that correct rounding can be handled automatically. This means that there are now methods for returning correct values in all 3 measurement types used in Finale (integer EVPU, integer EFIX and 10,000th EVPU)
- Modified `FCMCtrlEdit` so that it uses the new `FCMString` methods

I haven't touched the `Set*` methods. I might be able look into it next week, but at the moment finding time at the computer for testing is a bit tight and there's only so much I can write on my phone without being able to test.

Anyway, here is a test script. Since `FCMCtrlEdit` just calls the `FCMString` methods internally, the values should be the same but I put it in there just to make sure:
```lua
dialog:CreateStatic(0, 3):SetText("Enter a measurement:"):SetWidth(120)
dialog:CreateEdit(110, 0, "edit")
dialog:CreateMeasurementUnitPopup(190, 0)

local function create_handler(method)
    return function(self)
        local unit = self:GetParent():GetMeasurementUnit()
        local edit = self:GetParent():GetControl("edit")
        local str = mixin.FCMString()
        edit:GetText(str)

        local msg = ""
        if edit[method .. "_"] then
            msg = msg .."FCCtrlEdit: " .. edit[method .. "_"](edit, unit) .. "\n"
        end
        if str[method .. "_"] then
            msg = msg .. "FCString: " .. str[method .. "_"](str, unit) .. "\n"
        end
        if str[method] then
            msg = msg .. "FCMString: " .. str[method](str, unit) .. "\n"
        end
        if edit[method] then
            msg = msg .."FCMCtrlEdit: " .. edit[method](edit, unit) .. "\n"
        end

        finenv.UI():AlertInfo(msg, method .. " Results")
    end
end

dialog:CreateButton(0, 30):SetText("GetMeasurement"):SetWidth(90):AddHandleCommand(create_handler("GetMeasurement"))
dialog:CreateButton(95, 30):SetText("GetMeasurementInteger"):SetWidth(120):AddHandleCommand(create_handler("GetMeasurementInteger"))
dialog:CreateButton(220, 30):SetText("GetMeasurementEfix"):SetWidth(110):AddHandleCommand(create_handler("GetMeasurementEfix"))
dialog:CreateButton(335, 30):SetText("GetMeasurement10000th"):SetWidth(130):AddHandleCommand(create_handler("GetMeasurement10000th"))

dialog:ExecuteModal()
```
I intentionally made it regular text box, not a MeasurementEdit, so changing the unit doesn't change the display value, it just changes what unit the text box value is treated as when you click the buttons. Play around with it, including with overrides. For a list of overrides, check the code or look here: https://usermanuals.finalemusic.com/Finale2012Win/Content/Finale/OptionsOMeasurement_Units.htm There are hours of fun to be had!

@rpatters1 @cv-on-hub

### DIFF
--- a/src/library/utils.lua
+++ b/src/library/utils.lua
@@ -63,13 +63,16 @@ end
 --[[
 % round
 
-Rounds a number to the nearest whole integer.
+Rounds a number to the nearest integer or the specified number of decimal places.
 
 @ num (number)
+@ [places] (number) If specified, the number of decimal places to round to. If omitted or 0, will round to the nearest integer.
 : (number)
 ]]
-function utils.round(num)
-    return math.floor(num + 0.5)
+function utils.round(value, places)
+    places = places or 0
+    local multiplier = 10^places
+    return math.floor(value * multiplier + 0.5) / multiplier
 end
 
 --[[ 
@@ -144,6 +147,18 @@ Clamps a number between two values.
 ]]
 function utils.clamp(num, minimum, maximum)
     return math.min(math.max(num, minimum), maximum)
+end
+
+--[[
+% ltrim
+
+Removes whitespace from the start of a string.
+
+@ str (string)
+: (string)
+]]
+function utils.ltrim(str)
+    return string.match(str, "^%s*(.*)")
 end
 
 return utils
--- a/src/mixin/FCMCtrlEdit.lua
+++ b/src/mixin/FCMCtrlEdit.lua
@@ -6,6 +6,7 @@ $module FCMCtrlEdit
 Summary of modifications:
 - Added `Change` custom control event.
 - Added hooks for restoring control state
+- `GetMeasurement*` and `SetMeasurement*` methods have been overridden to use the `FCMString` versions of those methods under the hood. For more details on any changes, see the documentation for `FCMString`.
 ]] --
 local mixin = require("library.mixin")
 local mixin_helper = require("library.mixin_helper")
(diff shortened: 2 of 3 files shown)


rpatters1:
This looks good to me. I would like to hear from @cv-on-hub before merging it.

cv-on-hub:
LGTM. In fact LVGTM. Amazing to discover after all these years that `GetMeasurement10000th` is a thing.

ThistleSifter:
You didn't miss anything, in that they actually don't exist in the standard classes. But I thought that they should, so I added them to the mixins.

PR #403, opened 2023-01-20
Add `SetMeasurement` to `FCMString`

ThistleSifter:
Fixes an issue in `FCString` and `FCCtrlEdit` on Mac with displayed numbers being truncated (not even rounded...) at 2 decimal places. Since `FCMCtrlEdit` and `FCXCtrlMeasurementEdit` both use `FCMString` internally, this PR will fix the issues there too.

As with `GetMeasurement`, this once again emulates the behaviour that `FCCtrlEdit` has on Windows while the window is showing. All units display to 5 decimal places (for Picas, the whole number after the 'p' counts as the first decimal place).

For invalid measurement units, the Windows `FCCtrlEdit` behaviour is to just display the raw number, in other words as if it was EVPUs. The `FCString` behaviour was to set a blank string. I've gone with the Windows `FCCtrlEdit` behaviour but I'm happy to take suggestions.

I ran out of time this morning to write up a pretty test script. If you need one, I might be able to do it later today.

@rpatters1 @cv-on-hub

### DIFF
--- a/src/mixin/FCMString.lua
+++ b/src/mixin/FCMString.lua
@@ -4,8 +4,10 @@
 $module FCMString
 
 Summary of modifications:
-- Added `GetMeasurementInteger` and `SetMeasurementInteger` methods for parity with `FCCtrlEdit`
 - Fixed rounding bugs in `GetMeasurement` and adjusted override handling behaviour to match `FCCtrlEdit.GetMeasurement` on Windows
+- Fixed bug in `SetMeasurement` where all displayed numbers were truncated at 2 decimal places.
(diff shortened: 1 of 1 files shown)


cv-on-hub:
I don't know how many other contortions could be tested but this revision does everything that I want with strong accuracy to 5 decimal places. It also nicely trims the display of trailing zeroes.

rpatters1:
Let me know if I should merge. I am cool with it when you are.

ThistleSifter:
Go for it.

rpatters1:
It looks like this PR introduced some lint errors on `GetRangeMeasurement10000th`. Specifically, it looks as though `minimum` and `maximum` were omitted from the argument list of the function.

PR #404, opened 2023-01-24
Update cue_notes_create.lua

cv-on-hub:
Added the option of taking the source music from a layer other than 1.


rpatters1 (approved):
lgtm

PR #405, opened 2023-01-26
Add instructions for debugging with VS Code

asherber:
Instead of including the contents of `launch.json`, should the file live somewhere else (where?) and just be linked to here? It would make the docs shorter and less intimidating, especially with the need to include instructions to fix the manifest bug.

### DIFF
--- a/docs/rgp-lua/development-environment.md
+++ b/docs/rgp-lua/development-environment.md
@@ -3,9 +3,11 @@ RGP Lua Development Environment
 
 _RGP Lua_ is compatible with any external text editor or IDE. The only requirement is that the script you are developing has to have been configured in the [RGP Lua Configuration Window](/docs/rgp-lua/rgp-lua-configuration) and be visible in Finale's plugin menu. This can be done before you have actually added any code to the script.
 
-Two common development environments for Lua are [Visual Studio Code](https://code.visualstudio.com/) and [ZeroBrane Studio](https://studio.zerobrane.com/). Both have their advantages and drawbacks, and both were useful in the development of _RGP Lua_. Each has extensive documentation pages that need not be replicated here. These instructions focus on setting up debugging in ZeroBrane Studio because it is very easy to set up and use.
-
 One of the effects of the “Enable Debugging” option in _RGP Lua_ is to pre-embed the [`luasocket`](https://aiq0.github.io/luasocket/index.html) library in the Lua machine before calling the script. Many of the solutions for debugging embedded Lua over the years have used this library for communicating between the host program and the IDE.
+
+Two common development environments for Lua are [Visual Studio Code](https://code.visualstudio.com/) and [ZeroBrane Studio](https://studio.zerobrane.com/). Both have their advantages and drawbacks, and both were useful in the development of _RGP Lua_. Each has extensive documentation pages that need not be replicated here. Instructions for setting up a development environment in each application appear below.
+
+## Using ZeroBrane Studio
 
 The debugging solution presented here uses the script `mobdebug.lua` within ZeroBrane Studio.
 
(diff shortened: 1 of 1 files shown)


asherber:
Actually, I'd like to try tomorrow with one other debug extension. If I can get it to work, then I can remove the section about the manifest issue.

rpatters1:
I think the section on the manifest issue is exactly what needs to be there. If I had found that when I was trying to get it working it would have saved me a lot of time.

Documentation should be written in the manner of "explain it to me like I'm 5 years old". A link to a full `launch.json` is fine, but it would be very helpful to have the relevant code block for the debugger inline. It's, what, 10-15 lines of code?

asherber:
This morning I tried two other debugger extensions: Lua Debug (actboy168, has the most downloads) and Lua (Tencent, second most downloads). I couldn't get the first one working; it seems to use a different model for remote debugging that didn't play nicely with Finale. 

The second one is easy to use -- it even installs VS Code commands to download the debugger file and add the debugger code to your script! The downsides are that it doesn't seem to have a way to redirect print output to VS Code (although it does have a `printToVSCode` method that didn't seem to work) and it fills up the debug console with Chinese.

image

On the whole, I think I'm going to stick with my initial choice of debugger, even though it's a little more work to set it up. If you disagree, I can rewrite my instructions for the Tencent debugger. Otherwise, this PR is ready for review and merging.

rpatters1:
This looks great. Could add a comment about print redirect into the debugger? If it works out of the box, then a simple comment that no additional action is needed for it. I'm just trying to cover all the bases that the docs for ZeroBrane cover.

asherber:
I've got that -- look at my point 5.

Does it make sense to include mobdebug and vscode-debuggee in the repo, in a subfolder of src? That's one less step for everyone.

asherber:
@rpatters1 You may have seen, but this seems to have broken the page on the site. The rest of the site is fine, but if you click on this page, it's blank -- and then all other pages are blank.

View source shows a hydrate error:

`{name:"ReferenceError",message:"workspaceRoot is not defined",stack:"ReferenceError: workspaceRoot is not defined"},`

I'm not sure what framework is running your site, but it looks like it's trying to interpret the text in launch.json.

rpatters1:
I tried removing the json but that didn't fix it. I have reverted #405 for now, until we hear from @Nick-Mazuk about how to proceed.

rpatters1:
Thanks again @asherber for posting this. I'm definitely going to get this working on my system when I get a minute.

rpatters1:
P.S. I also like the idea of including modebug and vscode-debuggee in the repo. I would put them in a separate folder from src. We'll also need to include their license files from their source repos, which means each prolly needs to be in a separate sub-folder. Something like:

```
tools/modebug/
tools/vscode-debugee/
```
Lunajson is included in src because it is directly required by a production script. (The script that updates settings from a SMuFL json file.)

ThistleSifter:
Svelte (which powers the website) treats curly braces with something in between in a special way (ie `{workspaceRoot}`). So the options are either to escape the curly braces here in the markdown file (not preferred in my opinion because it won't make sense) or make a change in whatever markdown to HTML converter is used.

asherber:
Bummer. So how about I take that JSON out of there and instead link to a gist? I'm guessing I can't just link to a JSON file elsewhere on this site, because svelte will have the same issues with the braces.

asherber:
Actually, if I'm reading [https://svelte.dev/repl/2290fc67e7c0479881a266abb34e0f4d?version=3.20.1](https://svelte.dev/repl/2290fc67e7c0479881a266abb34e0f4d?version=3.20.1) and [https://stackoverflow.com/a/61089468/226781](https://stackoverflow.com/a/61089468/226781) correctly, I should be able to use &lbrace; and &rbrace; in the Markdown file. Does that sound right? I don't know anything about Svelte. 

What's interesting is that Svelte didn't complain about the braces in the diff file snippet.

Nick-Mazuk:
@ThistleSifter created a PR in the docs site that should fix it by escaping the curly braces:

https://github.com/finale-lua/jw-lua-scripts-docs/pull/23

I'm still looking at things in more detail to know for sure, but it seems like that PR fixes things.

FYI, sorry for just now seeing this. My job had layoffs this week and it was pandemonium trying to figure out who all was laid off and how to move forward without them.

Nick-Mazuk:
Well that PR removes the crashing, but it doesn't seem to format things correctly…

Screenshot 2023-01-28 at 10 40 43 AM

Looking into this some more.

Nick-Mazuk:
Ok, I have a fix.

Screenshot 2023-01-28 at 10 55 55 AM

https://jw-lua-scripts-docs-aaxu5tztx-finale-lua.vercel.app/docs/rgp-lua/development-environment

For those curious, the issue was the `${workspaceRoot}` in the file. Specifically the `${`. This wasn't a Svelte-related bug. But rather a JS related one.

In the website builder, the Markdown files for all the docs are converted to a JS template literal (a type of string). The `${}` lets you interpolate JS expressions into the string. So when it saw `${workspaceRoot}`, it started looking for the `workspaceRoot` variable.

The fix was just to escape the `$` to ensure JS doesn't interpolate in the string.

I deployed the fix to the site, so you should be able to update the documentation to include the original code snippets.

ThistleSifter:
Thanks for that. I did go digging but I must've missed where they were converted to template literals.

Nick-Mazuk:
That's understandable. The actual implementation is kind of crude. At some point I should redo the docs generation as more people are contributing now. And I'm kind of a single point of failure for that.

Anyways, probably the easiest thing to do is to see the "after" from the markdown processing. The Markdown file goes into a template literal, then that goes into a Markdown component that parses the template literal then converts it to HTML. The build system this 2-year-old version of Svelte uses doesn't support importing the Markdown files directly, which is why it's converted to a template literal.

https://github.com/finale-lua/jw-lua-scripts-docs/blob/main/src/routes/docs/rgp-lua/development-environment.svelte

ThistleSifter:
Ahhh... I was looking at this on my phone and I must've blinked or something and missed the backticks around `MARKDOWN_PLACEHOLDER` in the template.

Let me know if there's anything I can do to help out with updating the build process.

PR #409, opened 2023-01-28
Fix braces in VS Code instructions

asherber:
I know this doesn't look right in the GitHub viewer, but I think by the time Svelte is done with it, it should work. Can you merge this and see how it looks on the site? You can always undo the merge if needed.


asherber:
@rpatters1 Well, it didn't break, but it also isn't displaying properly. I guess I don't know how to make Svelte happy. Maybe the diff and launch.json need to just be images?

rpatters1:
I think the diff and json as images would be the best bet for now, at least until @Nick-Mazuk chimes in. There is probably some additional work needed on the homegrown [documentation processor](https://github.com/finale-lua/jw-lua-scripts-docs) to support it.

Put the images in the assets subfolder with links out of the markdown as in the rgp-lua-configuration markdown.

PR #411, opened 2023-01-28
Use images for VS Code instructions

asherber:


rpatters1 (approved):
let's give it a go

rpatters1:
The images are low res on a Mac Retina display, which is a little disconcerting. But I guess it will do for now, at least until Nick reappears.

asherber:
They look fine on my (WIndows) monitor, but I agree it's not the best solution. Interestingly, there's no problem with the braces in the code snippet. I think it might only be an issue when a left brace is followed by a character (which is why the diff snippet was actually fine as well).

So I think there's two things that it would be nice to resolve:

1.  How to enter the JSON text with braces so it displays correctly and doesn't break Svelte
2.  How to implement better syntax highlighting. When the diff snippet was text, it was displaying with no highlighting at all, which made it more difficult to interpret. In that sense, I think the image is an improvement. 

Anyway, I'm happy to revise again when we have more information.

rpatters1:
The big downside to using images is that I can't copy the text out that is needed to insert into the manifest file to fix the Lua Debugger issue. Dagnabbit.

asherber:
True, but even with text you'd be copying the + signs as well.

Issue #420, opened 2023-01-30
Suggestion: Improved text for Rest Offsets


asherber:
I thought there was a bug in how this plugin was positioning rests, but it turned out that I was just confused about the design.
 
Although the plugin is called Rest Offsets, the descriptive text in the dialog says `4 = top staff line, 0 = middle staff line, -4 = bottom staff line`. This gives the impression that the plugin will perform absolute positioning of rests. (Which, actually, would be nice functionality to have. When I use the Finale Move Rests plugin, I'm always having to count and figure out how many steps to move the rests from where they are to where I want them; it would be nice to be able to say "Put the rests here".)

In fact, the plugin does apply an offset, so that entering `4` will only put the rest on the top line of the staff if it started out on the middle line. But it applies that offset not to the rest's current position, but rather to its default position. Thus if I apply this to a whole rest and specify -6, it will wind up on the bottom line regardless of where it is currently. (Maybe I moved it somewhere previously and now need to move it again.)

From a functionality perspective, this is fine with me. But the name of the plugin and the descriptive text could be improved to better indicate what's going on. A better name might be "Rest Offsets From Default", and the text could indicate that the entered value will be applied against the rest's default value rather than its current value. 

cc: @cv-on-hub

cv-on-hub:
Fair comment @asherber, and renaming the script "Rests Vertical Position" makes sense. But it was never a massive step to assume that "offset" was relative to the default (middle) position! `JWChange` makes the distinction by offering "relative" or "absolute" values. (I trust that you also read the explanation of floating rests hiding behind the **?** button).

It's a messy business working out precise current rest positions and I'll leave that up to Finale's inbuilt **Move Rests...** plugin.  For me if you want the option of moving to an offset relative to CURRENT position then use that one. 

I've changed the documentation and the name of the script (and menu item).

asherber:
@cv-on-hub I think I wasn't clear, and maybe I've made things worse. "Offset" is a fine name, and accurately describes what the plugin does. It's just that the text in the dialog make it seem like it's doing absolute positioning, but it's not.

You say "relative to the default (middle) position", which would also make sense with the dialog text, but that's also not what it does. For a whole note rest, for example, it's relative the the default position on the 4th line, not on the 3rd (middle) line. So +2 to get to the top line, -6 to the bottom line -- not +4 to the top, -4 to the bottom, as the dialog indicates.

[...]

PR #422, opened 2023-01-30
move rest_offsets.lua to rests_vertical_position.lua

cv-on-hub:
Also modified documentation and minor code adjustments.


cv-on-hub:
Response to Issue #420

PR #426, opened 2023-02-02
Update cue_notes_create.lua

cv-on-hub:
Updated all dialogs to use mixin; implemented choice of multiple destination staves with set/preset options.

### DIFF
--- a/src/cue_notes_create.lua
+++ b/src/cue_notes_create.lua
@@ -3,18 +3,19 @@ function plugindef()
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "http://carlvine.com/lua/"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "v0.76"
-    finaleplugin.Date = "2023/02/02"
+    finaleplugin.Version = "v0.77"
+    finaleplugin.Date = "2023/02/03"
     finaleplugin.Notes = [[
         This script is keyboard-centred requiring minimal mouse action. 
-        It takes music from a nominated layer in the selected staff and creates a "Cue" version on another staff. 
+        It takes music from a nominated layer in the selected staff and 
+        creates a "Cue" version on one or more other staves. 
         The cue copy is reduced in size and muted, and can duplicate nominated markings from the original. 
         It is shifted to the chosen layer with a whole-note rest placed in the original layer.
 
-        Your prerences are saved in the User prefs folder after each script run. 
+        Your preferences are preserved between each script run. 
         This script requires an expression category called "Cue Names". 
         Under RGPLua (v0.58+) the category is created automatically if needed. 
-        Under JWLua, before running the script you must create an Expression Category called 
+        Using JWLua, before running the script you must create an Expression Category called 
         "Cue Names" containing at least one text expression.
         ]]
     return "Cue Notes Create...", "Cue Notes Create", "Copy as cue notes to another staff"
@@ -32,7 +33,7 @@ local config = { -- retained and over-written by the user's "settings" file
     source_layer        =   1,     -- layer the cue comes from
     cuenote_layer       =   3,     -- layer the cue ends up
     rest_layer          =   1,     -- layer for default wholenote rest
-    freeze_up_down      =   0,     -- "0" for no freezing, "1" for up, "2" for down
+    freeze_up_down      =   0,     -- "0" for no stem freezing, "1" for up, "2" for down
     -- if creating a new "Cue Names" category ...
     cue_category_name   =   "Cue Names",
     cue_font_smaller    =   1, -- how many points smaller than the standard technique expression
@@ -84,9 +85,7 @@ function new_cue_name(source_staff)
     dialog:CreateOkButton()
     dialog:CreateCancelButton()
     local ok = (dialog:ExecuteModal(nil) == finale.EXECMODAL_OK)
-    local str = finale.FCString()
-    the_name:GetText(str)
-    return ok, str.LuaString
+    return ok, the_name:GetText()
 end
 
 function choose_name_index(name_list)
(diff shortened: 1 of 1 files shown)


src/cue_notes_create.lua
```diff
@@ -3,15 +3,15 @@ function plugindef()
...
+    finaleplugin.Version = "v0.76"
+    finaleplugin.Date = "2023/02/02"
     finaleplugin.Notes = [[
         This script is keyboard-centred requiring minimal mouse action. 
         It takes music from a nominated layer in the selected staff and creates a "Cue" version on another staff. 
         The cue copy is reduced in size and muted, and can duplicate nominated markings from the original. 
         It is shifted to the chosen layer with a whole-note rest placed in the original layer.
 
-        Your choices are saved in the User preferences folder after each script execution. 
+        Your prerences are saved in the User prefs folder after each script run.
```
ThistleSifter, line 14: prerences -> preferences

src/cue_notes_create.lua
```diff
@@ -71,52 +75,32 @@ function region_contains_notes(region, layer_number)
...
-    local staff = finale.FCStaff()
+    local staff = finale.FCStaff() -- copy the source Staff Name
     staff:Load(source_staff)
-    the_name:SetText(staff:CreateDisplayFullNameString())
-
+    local the_name = dialog:CreateEdit(0, 40):SetWidth(200):SetText(staff:CreateDisplayFullNameString())
     dialog:CreateOkButton()
     dialog:CreateCancelButton()
     local ok = (dialog:ExecuteModal(nil) == finale.EXECMODAL_OK)
+    local str = finale.FCString()
```
ThistleSifter, line 87: I think lines 87-89 could be shortened since `FCMControl.GetText` returns a string:
```lua
return ok, the_name:GetText()
```

cv-on-hub:
That’s some really fine proof-reading! Thank you @ThistleSifter. Very handy to discover that about FCControl:GetText(). A few other minor tweaks in the following version.

ThistleSifter:
A couple of notes, not necessarily things you have to change:
Line 422: I was wondering if there was a reason for using `string.find` as opposed to `==` comparison?

```lua
function assign_expression_to_staff(staff_number, measure_number, measure_position, expression_id)
    local new_expression = finale.FCExpression()
    new_expression:SetStaff(staff_number)
    new_expression:SetVisible(true)
    new_expression:SetMeasurePos(measure_position)
    new_expression:SetScaleWithEntry(false) -- could also (possibly) be true!  
    new_expression:SetPartAssignment(true)
    new_expression:SetScoreAssignment(true)
    new_expression:SetID(expression_id)
    new_expression:SaveNewToCell( finale.FCCell(measure_number, staff_number) )
end
```

I wouldn't mind taking a deeper look at this, but I'm not sure when I'll be able to. So whoever merges this, don't wait for me.

ThistleSifter:
Bah, I accidentally pressed the `comment` button on my phone.

The other thing I was going to mention, just as an example, was the function at line 389. Since you've already included the `mixin` library, you can do the following at no extra charge:
```lua
function assign_expression_to_staff(staff_number, measure_number, measure_position, expression_id)
    mixin.FCMExpression()
        :SetStaff(staff_number)
        :SetVisible(true)
        :SetMeasurePos(measure_position)
        :SetScaleWithEntry(false) -- could also (possibly) be true!  
        :SetPartAssignment(true)
        :SetScoreAssignment(true)
        :SetID(expression_id)
        :SaveNewToCell( finale.FCCell(measure_number, staff_number) )
end
```
You can do that in all sorts of places, if you like.

cv-on-hub:
Wow! I didn’t know that `FCMExpression` existed, and can’t see it reflected in (current) repo info. Part of your larger documentation project I guess. As you say, with `mixin` loaded you’d be mad not to use the extensions. Hard to know right now what other classes have expanded `mixin` versions, so I look forward to full revelation. `FCMTextExpressionDef` **does** seem to exist (for instance) but doesn’t behave the same way as the FC version.

As for string matching at line 422, I was originally going to be more lenient matching category titles. But when I introduced a “fixed” config name for the new cat it didn’t seem so important. Equality is good.

FCMExpression is now so compact that the code fragment doesn’t deserve a stand-alone function. I appreciate  changes aren’t demanded but would still like to offer a more mixin-compressed script.

cv-on-hub:
PS: just re-read your PR #417 and the `mixin` biz is beginning to make sense. Beyond my ken and pay-grade, but I can make some headway through trial & error.

ThistleSifter:
I guess the tl;dr version is:
```lua
-- Replace this
finale.FCString()

-- With this
mixin.FCMString()

-- To get a Finale object with a fluid interface that can be modified
-- If mixins exist in the `mixin` folder (either for that class or a parent class), they will be loaded and the methods will be added / existing ones overridden
```
Feel free to shoot me an email anytime if you have any questions or you want something explained. The whole point of writing this stuff is so that it can be used.

My only concern with using `string.find` was that it will match if the search string exists anywhere in the subject, not necessarily at the start. And it will also match if there is more afterwards. For example:
```lua
string.find("Cue Names (Old)", "Cue Names") -- this is a match

string.find("Old Cue Names", "Cue Names") -- this is also a match
```

rpatters1:
Carl,

The 'mixin' library itself knows how to make a mixin version of any FC class. So there doesn't have to be a separate class script to get the basic mixin behavior, such as fluid interface. (Which is what was suggested above for `FCMExpression`.)

cv-on-hub:
Thanks @rpatters1. I think I'm getting the hang of it now and @ThistleSifter has pointed me in the direction of creating new mixins. Also explaining why `FCMTextExpressionDef` throws an error at `exp_def:Load()`.

rpatters1:
>Also explaining why FCMTextExpressionDef throws an error at exp_def:Load().

So is this PR ready to go?

cv-on-hub:
> So is this PR ready to go?

Thanks @rpatters1, yes it is. Might add another mod or two after a new `Mixin` gets added.

PR #432, opened 2023-02-07
add support for independent time sigs to meter_change.lua

rpatters1:
This PR adds support for independent time signatures to the meter change script. If you select a staff with independent time signatures, only that staff is modified. You can select a whole stack and all of them will be modified, independent or not.

The PR also removes spurious calls to `FCTimeSignature:Save` in meter_set_numeric. This future-proofs it against changes that are coming in RGP Lua 0.66.

### DIFF
--- a/src/meter_change.lua
+++ b/src/meter_change.lua
@@ -1,10 +1,10 @@
 function plugindef()
     finaleplugin.Author = "Nick Mazuk"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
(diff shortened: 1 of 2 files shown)


cv-on-hub:
So it seems that `FCTimeSignature:Save` is redundant! `src/meter_set_numeric.lua` LGTM. I hope you didn't want me to make it handle independent TS's!

rpatters1:
> I hope you didn't want me to make it handle independent TS's!

I leave that up to you.

rpatters1:
In the absence of further commentary, I'm merging this.

PR #433, opened 2023-02-07
Update mixin_helper and new file

cv-on-hub:
Add functions to mixin_helper: to_fcstring() and boolean_to_error(). New FCMTextExpressionDef.lua mixin file.

### DIFF
--- a/src/library/mixin_helper.lua
+++ b/src/library/mixin_helper.lua
@@ -1,5 +1,6 @@
 --  Author: Edward Koltun
---  Date: April 3, 2022
+--  Date: 2023/02/07
+--  version 0.2
 --[[
 $module Mixin Helper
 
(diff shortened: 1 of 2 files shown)


rpatters1:
I will to let @ThistleSifter comment on this. It looks good to me.

One thing I noticed is that after one of the recent merges, `mixin.lua` is throwing a couple of lint errors.

ThistleSifter (approved):
Just give this a test and it works well.

ThistleSifter:
@rpatters1 Was there anything in particular that you thought was concerning in the lint errors? The ones I saw were harmless (like type errors in a loosely typed language...)

rpatters1:
Type errors in a loosely typed language are exactly the kind of lint errors most worth addressing. Generally speaking, I think we should be coding in such a way as to avoid them, even if it means adding some extra lines of code here and there. Type errors are a potential minefield.

ThistleSifter:
So I dug into this further. The underlying issue is that we're relying on the linter to infer types (variables, function arguments, function returns) automatically. It's limited in what it can do in this regard, it doesn't always guess correctly and there are also certain things it just won't infer automatically. That's where annotations come in, which we're currently not using (the docblocks aren't recognised). The documentation for annotations is here: https://github.com/LuaLS/lua-language-server/wiki/Annotations

Whether it's worth adding annotations, I don't know. More research would be needed, in particular whether they can be reliably used to declare Finale object types and then mixin types as well.

rpatters1:
As of the latest commits, the only lint error that is appearing in the mixin code is an undefined global variable error.

ThistleSifter:
Oh? Hmm... I was seeing all sorts of things popping up. Maybe my config got messed up somehow.

I've got another PR coming with some changes to `mixin.lua` so it should be fixed in that.

ThistleSifter:
Is there anything preventing this from being merged?

rpatters1:
Merging now.

PR #434, opened 2023-02-08
Add document_options_serialize.lua

asherber:
All comments and suggestions welcome! I've tested this on Win 25 and 27, but I don't have access to a Mac.

### DIFF
--- a/src/document_options_save_as_json.lua
+++ b/src/document_options_save_as_json.lua
@@ -5,7 +5,7 @@ function plugindef()
     finaleplugin.Author = "Aaron Sherber"
     finaleplugin.AuthorURL = "https://aaron.sherber.com"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "1.1.0"
+    finaleplugin.Version = "1.1.1"
     finaleplugin.Date = "2023-02-09"
     finaleplugin.CategoryTags = "Report"   
     finaleplugin.Id = "9c05a4c4-9508-4608-bb1b-2819cba96101" 
@@ -52,7 +52,7 @@ function do_save_as_dialog(document)
     local filter_text = "JSON files"
 
     local path_name = finale.FCString()
-    local file_name = finale.FCString()
+    local file_name = mixin.FCMString()
     local file_path = finale.FCString()
     document:GetPath(file_path)
     file_path:SplitToPathAndFile(path_name, file_name)
@@ -63,8 +63,9 @@ function do_save_as_dialog(document)
     if extension.Length > 0 then
         file_name:TruncateAt(file_name:FindLast("."..extension.LuaString))
     end
-    file_name:AppendLuaString(get_file_options_tag())
-    file_name:AppendLuaString(text_extension)
+    file_name:AppendLuaString(" settings")
+            :AppendLuaString(get_file_options_tag())
+            :AppendLuaString(text_extension)
     local save_dialog = mixin.FCMFileSaveAsDialog(finenv.UI())
             :SetWindowTitle(fcstr("Save "..full_file_name.." As"))
             :AddFilter(fcstr("*"..text_extension), fcstr(filter_text))
--- a/src/document_options_save_as_json.lua
+++ b/src/document_options_save_as_json.lua
@@ -5,8 +5,8 @@ function plugindef()
     finaleplugin.Author = "Aaron Sherber"
     finaleplugin.AuthorURL = "https://aaron.sherber.com"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "1.1.1"
-    finaleplugin.Date = "2023-02-09"
+    finaleplugin.Version = "1.1.2"
+    finaleplugin.Date = "2023-02-10"
     finaleplugin.CategoryTags = "Report"   
     finaleplugin.Id = "9c05a4c4-9508-4608-bb1b-2819cba96101" 
     finaleplugin.Notes = [[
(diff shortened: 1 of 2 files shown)


asherber:
Should I add the word "settings" to the file name?

ThistleSifter:
I think YAML needs to go in the script and file name, since it's not the only serialization format out there.

Do you have plans for an unserialise YAML script?

I haven't tested this out but hopefully I can do it tomorrow.

asherber:
On the topic of readability, do you think it's worth translating all non-obvious measurements from their stored values? I'm thinking of EFIX values that I could turn into EVPU, or percentage values that Finale stores as a large integer. I wish there were a way of doing this through reflection, but it's probably not too many to set up a table of values that need to be transformed.

ThistleSifter:
Sorry about the confusion, I meant the script's `.lua` filename and display name. But having said that...

What about adding a dialog for script options? Then you can have options for things like:
- Export format (YAML or JSON)
- Export raw or normalized values (EFIX->EVPUs, 10,000th EVPU -> EVPU, Number -> percentage, etc)
- If normalized, which measurement unit
- Allow/disallow partial import (or just use an alert asking for user confirmation before continuing)

For normalization, I would suggest using measurement unit overrides, along with the `%` sign for percentages.

If you end up adding import functionality (which I would very much like to see), then supporting overrides would be very helpful for users who like to use different measurement units for different settings.

Since YAML supports comments, you could add inline comments with non-obvious value types. Eg
```yaml
# Raw
SlurTipWidth: 12345 #EVPU/10000

# Normalized
SlurTipWidth: 1.2345e #EVPU/10000

# Normalized into inches
SlurTipWidth: 0.00429i #EVPU/10000
```

In the case of a malformed file, you could either display an error message or show a Yes/No confirmation alert with the caveat that if the user proceeds, results are not guaranteed.

asherber:
Well, I don't want the perfect to be the enemy of the good. I like your suggestions, but from what I can see it would be a huge amount of work to allow for the kind of flexibility you envision -- basically tagging every single numeric property to indicate what the native unit is. I think it's more straightforward to assume that we're dealing in EVPU and just normalize any measurement with a different value (and normalize percentages). 

So I suggest:

*   Import to be a future possibility
*   Switch serialization format to JSON (for better import support), don't offer a choice
*   Basic normalization of numeric values
*   Clearer script filename and display name

If that's an acceptable goal, I'll change this PR to draft and get to work.

asherber:
Thanks for pointing out the mixin functions -- I admit I haven't really dived into those yet. I can see where those would be useful, especially if I added a function that basically does the reverse of `convert_to_EVPUs`. And as I think about it, you're right that it's not as big a task as I first envisioned.

But I'd like to get this to a place where a usable version can be released and then added on to later, so I'm going to proceed with mods as I outlined above.

If anyone else has any other suggestions for improvement, please let me know!

Nick-Mazuk:
Just wanted to say I've been looking for something like this for years! It's been on my backlog of scripts to write for quite a while now, so thanks a ton for putting something like this together!

I'm not going to have time to review the code or give implementation pointers (sorry), but on the feature side one thing that would be great is to be able to import several YAML files where the settings from the later ones override the settings from the first ones. That way you could effectively compose settings together. For instance, you could have the base settings for your house style. Then when you do jazz, marching band, classical, recording sessions, etc., you can just override the fewest numbers of settings necessary to tweak your house style for the genre.

ThistleSifter:
[serialize.txt](https://github.com/finale-lua/lua-scripts/files/10688527/serialize.txt)
@asherber Here's a quick example of normalizing output with a user selected unit that I wrote out on my phone from the existing code (hopefully I haven't mistyped something and sorry if the spacing in the dialog is wonky!). I just did it for the classes without a custom loader and I only added specific types from FCPianoBracePrefs (it assumes every other number value is an EVPU) but it should be easy to extrapolate.

rpatters1:
What is the use case we are solving here? I thought it was to save document settings in a way that makes them easily imported into another file. If that's the case, the internal format seems like it would hardly matter. The code could be repetitively stupid and ought not to have to know anything about units.

Same thing for error checking on input. I don't see a lot of value in it if the target use case is to reimport file originally exported with the tool.

asherber:
@ThistleSifter Thanks, impressive phone coding! That's the sort of thing I was picturing. 

@rpatters1 You're describing the original use case. However, even in just diffing files, if I see some enormous value, it's not going to make sense to me, and I'm not going to know what to enter in Finale to change that setting. And @Nick-Mazuk is describing a system of importing things from multiple files, some of which may be hand-edited or constructed. So I can see the value in serializing these values in a way that makes sense.

asherber:
Okay, I have switched to JSON, renamed the script file and changed the script name, and implemented basic normalization of values. All physical measurements are now converted to EVPU, with the exception of a couple of values that Finale always displays in spaces regardless of your preferred units. I have also normalized percentages so that they display the way they do in Finale; the raw values sometimes store the decimal value of the percentage, sometimes a multiple of the decimal value.

Nick-Mazuk:
Overall, I'm fine with JSON. Just one consideration, though: JSON doesn't allow for comments. This means we won't be able to add comments to guide non-devs into how to modify the preferences.

If we're optimizing for non-devs, I still think JSON is the better choice. However, I think it's at least worth noting this con.

asherber:
I know it's bad form to push changes to a PR under review, but I don't think anyone has had time to actively review this yet, and I just pushed one change to get rid of an internal mapping table.

rpatters1:
Lol, bad form or not, it's very common in this repo. I'm waiting for @Nick-Mazuk for the final ok. I'm fine with it.

src/document_options_save_as_json.lua
```diff
@@ -0,0 +1,709 @@
...
+    document:GetPath(file_path)
+    file_path:SplitToPathAndFile(path_name, file_name)
+    local full_file_name = file_name.LuaString
+    local extension = mixin.FCMString()
+                            :SetLuaString(file_name.LuaString)
+                            :ExtractFileExtension()
+    if extension.Length > 0 then
+        file_name:TruncateAt(file_name:FindLast("."..extension.LuaString))
+    end
+    file_name:AppendLuaString(" settings")
```
Nick-Mazuk, line 66: @asherber You're on Windows, right? I presume you've tested this?

Just want to make sure since I've seen issues with using spaces in file names, but that was a long time ago or on operating systems Finale doesn't support. Spaces in filenames definitely works on macOS, so just want to ensure we've also tested this on Windows.

src/document_options_save_as_json.lua
```diff
@@ -0,0 +1,709 @@
...
+
+function get_last_key(t)
+    local result
+    for k, _ in pairsbykeys(t) do result = k end
+    return result
+end
+
+-- Simple implementation because we know our inputs are simple
+-- Can't use a json library because we need to keep keys in order
+function get_as_json(t, indent)
```
Nick-Mazuk, line 634: This would be a great function to add to our function library. Not necessary for this PR, but a good future candidate for one.

https://www.finalelua.com/docs/library

src/document_options_save_as_json.lua
```diff
@@ -0,0 +1,709 @@
...
+    local documents = finale.FCDocuments()
+    documents:LoadAll()
+    local document = documents:FindCurrent()
+
+    local file_to_write = do_save_as_dialog(document)
+    if not file_to_write then
+        return
+    end
+
+    local file = io.open(file_to_write, "w")
```
Nick-Mazuk, line 696: Likewise, we're starting to duplicate writing to file behavior across multiple scripts. Would be good to abstract this out at some point.

Nick-Mazuk (commented):
Overall, LGTM. Feel free to merge without changes.

Looking through the code, I don't see deserialization. Was that intentional, or did I miss something? It's totally fine to create that in a followup PR. If so, we may want to distribute them using Additional Menu Options. If that's the case, we'll probably want to change the filename of this script to support both saving to and loading from JSON.

https://www.finalelua.com/docs/rgp-lua/finaleplugin-properties#additional-menu-options

Nick-Mazuk (approved):

src/document_options_save_as_json.lua
asherber: I'm on Windows and have tested, no issues with this.

src/document_options_save_as_json.lua
asherber: We'd need to be a little careful. The repo already includes lunajson, which I think is used by one of Robert's scripts, and it can do full serialization and deserialization -- but it doesn't do formatted serialization and can't keep the keys in order (needed in order to be able to diff these files). This little function does formatting and ordering, but it basically assumes that the values are all strings, numbers, bools, or tables. If it's desirable to adopt this into the library, then we should make sure it can handle any values that the library might reasonably throw at it (thinking of arrays, e.g., which are definitely not supported by this function).

src/document_options_save_as_json.lua
asherber: I was thinking of that -- I basically stole the File Save and write to file from Robert's `document_save_as_text` :smile:

asherber:
@Nick-Mazuk Deserialization will come later -- it's got issues of its own, and I thought there was value in getting this part out into the world as is.

PR #435, opened 2023-02-08
Update notehead.lua

cv-on-hub:
Added layer filtering to the change_shape() method.

But on reflection, would layer filtering be more useful at FCNoteEntry level, and not FCNote? In which case create two new methods accepting those inputs? Thoughts?

### DIFF
--- a/src/library/notehead.lua
+++ b/src/library/notehead.lua
@@ -216,7 +216,7 @@ Layer number is optional.
 ]]
 function notehead.change_shape(note, shape, layer)
 
-    if layer and type(layer) == "number" and note.Entry.LayerNumber ~= layer then
+    if layer and type(layer) == "number" and layer > 0 and note.Entry.LayerNumber ~= layer then
         return -- note on the wrong layer
     end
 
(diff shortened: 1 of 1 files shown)


rpatters1:
How are you imagining this will be called? If it inside an `eachentrysaved` loop, you can filter the layer in `eachentrysaved`.

cv-on-hub:
> you can filter the layer in `eachentrysaved`

That was my quandary. Are there other reasonable instances where `note` or `entry` might objects arise for which the layer isn't predetermined? And separately, are `notehead.change_shape_entry()` and `notehead.change_shape_region()` useful additions to the library?

rpatters1:
The is no reason to add library functions for the sake of adding library functions. There needs to be a clear and present problem it is solving. These functions you propose would not be added to `notehead`. The entry one would be added to `note_entry`. We could discuss where to add a region based one.

My question is specific. Why are we adding a layer filter at all at any level, when layer filters already exist in `eachentrysaved`? Is that not enough for what you are doing?

cv-on-hub:
> Is that not enough for what you are doing?

`eachentrysaved` is fine for what I need. I was asking whether some other circumstance I haven't imagined could sensibly use layer filtering.

PR #436, opened 2023-02-09
Update notehead.lua - default size correction

cv-on-hub:
setting `default` character now resets any previously applied size and position changes. (I thought these would have been cleared by the preceding `notehead_mod:EraseAt()` but they aren't.


rpatters1 (approved):
lgtm

PR #437, opened 2023-02-09
Create noteheads_change_by_layer.lua

cv-on-hub:
New script with the functionality of `noteheads_change.lua` but with layer filtering, a single menu item and a single confirmation dialog. I now use this always instead of the old script.

### DIFF
--- a/src/noteheads_change_by_layer.lua
+++ b/src/noteheads_change_by_layer.lua
@@ -3,8 +3,8 @@ function plugindef()
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "http://carlvine.com/lua/"
     finaleplugin.Copyright = "https://creativecommons.org/licenses/by/4.0/"
-    finaleplugin.Version = "v0.11"
-    finaleplugin.Date = "2023/02/09"
+    finaleplugin.Version = "v0.12"
+    finaleplugin.Date = "2023/02/10"
     finaleplugin.MinJWLuaVersion = 0.62
     finaleplugin.Notes = [[
         Change notehead shapes on a specific layer of the current selection to one of these shapes:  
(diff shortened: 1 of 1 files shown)


rpatters1:
Is there a reason you are using `FCMCustomLuaWindow` instead of `FCXCustomLuaWindow`? The `X` version has (among other things) automatic remembering of window position. (Though I discovered it isn't working in the current master branch, but we will fix that.)

cv-on-hub:
> `FCMCustomLuaWindow` instead of `FCXCustomLuaWindow`?

I noted that @ThistleSifter was migrating some of FCX to FCM and wondered what implications would arise using the "lower" version of `LuaWindow`. Happy to use either - just testing.

rpatters1:
Main I'm interested in if the script needs to have its own window position handling. I think it's built into `FCX...`.

cv-on-hub:
Updated my local library & mixin files to latest `/src/` versions and tried FCX - doesn't keep window positions in modal. The `private[window].HasBeenShown` conditional within `local restore_position()` suggests it's unique to modeless operation. All of the positioning routines (that I can find) are in the FCM file, not FCX.

Obviously it __could__ be implemented for modal, just a few more steps to add.

cv-on-hub:
PS: also might still be tied to `RetainLuaState` which I'm not using here.

ThistleSifter:
Stick to `FCX` for now. I'll replace them in a future PR when I remove the `FCXCustomLuaWindow` mixin.

Automatic restoring of window position only works with `RetainLuaState`, since you need a previous position to restore from. For something like this, where it's just being loaded from a config file each time, it will need to be manually handled. At least until automatic populating of the window from a config file is implemented ... maybe ...

cv-on-hub:
No problem. Happy to keep it in-house - only an extra 6 lines of code. FCX version follows.

rpatters1 (approved):
lgtm

PR #440, opened 2023-02-12
Update document_options_save_as_json

asherber:
I've added the grid/guides snap-to items, which I overlooked, and reorganized the SmartShapes section so it better matches the separate dialogs the settings appear in. Also some internal improvements to the serialization code.


rpatters1 (approved):
lgtm

PR #445, opened 2023-02-14
choose first part where the selected staff is visible

rpatters1:
This PR refines `UI_switch_to_selected_part.lua` as follows:

Previously it selected the first part where the staff in the part. This PR changes it to select the first part where the staff is visible in the part on the selected system.

What this addresses:

Many times a linked part will include one or more staves from other parts that are cue staves. It is often that case that the cue staves are mostly force-hidden and only appear in a very restricted passage. This change allows the script to ignore cue staves as long as they are hidden.


rpatters1:
In the absence of comment, I going ahead and merging this. If y'all want it backed out I'll do so.

PR #447, opened 2023-02-19
Update vscode docs for different extension

asherber:
After growing a bit frustrated with the VS Code debug extension I had been using, I went for another round of testing with other extensions and found one that is easier to set up and (on the whole) use.


rpatters1 (approved):
lgtm

PR #450, opened 2023-02-24
Update cue_notes_create.lua

cv-on-hub:
Added stem freezing "away_from_middle" and extra menu item "Cue Notes Flip Frozen". (additional coding by @asherber)

### DIFF
--- a/src/cue_notes_create.lua
+++ b/src/cue_notes_create.lua
@@ -73,12 +73,12 @@ function show_error(error_code)
     local errors = {
         only_one_staff = "Please select just one staff\n as the source for the new cue",
         empty_region = "Please select a region\nwith some notes in it!",
-        no_notes_in_source_layer = "The music selected contains\nno notes in layer " .. config.source_layer,
+        no_notes_in_source_layer = "The selected music contains\nno notes in layer " .. config.source_layer,
         first_make_expression_category = "You must first create a new Text Expression Category called \""..config.cue_category_name.."\" containing at least one entry",
-        no_cue_notes = "The region selected contains \nno cue notes in layer " .. config.cuenote_layer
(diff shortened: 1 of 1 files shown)


src/cue_notes_create.lua
```diff
@@ -54,9 +75,10 @@ function show_error(error_code)
         empty_region = "Please select a region\nwith some notes in it!",
         no_notes_in_source_layer = "The music selected contains\nno notes in layer " .. config.source_layer,
         first_make_expression_category = "You must first create a new Text Expression Category called \""..config.cue_category_name.."\" containing at least one entry",
+        no_cue_notes = "The region selected contains \nno cue notes in layer " .. config.cuenote_layer
     }
     local msg = errors[error_code] or "Unknown error condition"
-    finenv.UI():AlertNeutral("script: " .. plugindef(), msg)
+    finenv.UI():AlertInfo("script: " .. plugindef(), msg)
```
asherber, line 81: @cv-on-hub Shouldn't the `msg` param come first?

cv-on-hub: The **second** string parameter (the optional "Title") is bigger and bolder. I wan't the "message" to be bigger and bolder so place it second. In this usage, at least, I'm only adding the name of the script as a courtesy. Not so important.

asherber: It's up to you, of course, but on Windows the dialog title is neither bigger nor bolder. It also doesn't support line breaks, so you get text like "regionwith". And from a usability perspective, it seems unusual to look at the title bar of a dialog to get the important information.

cv-on-hub: Thanks for letting me know. That's really ugly. Will remove the script name and move the message to the place where Windows can handle it.

asherber: Thanks, lgtm

rpatters1 (approved):
lgtm

PR #451, opened 2023-02-25
Add import settings from JSON

asherber:
- Added the ability to import (complete or partial) JSON settings files
- Renamed script file 

More detailed help file at https://url.sherber.com/finalelua/options-as-json

As @Nick-Mazuk requested, the script can handle multiple imports (though sequentially, not through a single operation), so you can apply one complete settings file for your house style and then apply one or more partial files with minimal changes for a client/project/genre/etc.

Comments and suggestions welcome.


asherber:
As a side note, since the autopublish scripts still appear to not be cleaning out the dist folder correctly, it might be worth adding a commit that cleans that folder out by hand and then letting the autopublish repopulate. (I'm thinking  that since I renamed my script, the old file might stick around in the dist folder otherwise.)

Also, on [https://www.finalelua.com/scripts](https://www.finalelua.com/scripts) I see this:



I think this is because `system_move.lua` gives `AdditionalMenuOptions` as a string delimited with quotes, and the parser is expecting a multiline string.

asherber:
Folks, is there anything else needed for this PR?

Nick-Mazuk (approved):
LGTM

asherber:
@rpatters1 Can you please merge this? Or is something else needed?

rpatters1:
My earlier commit seems to have broken the documentation generator. Not sure if that will affect this.

asherber:
https://github.com/finale-lua/lua-scripts/actions/runs/4376204383/jobs/7657970173 has:

image

asherber:
That script doesn't seem to be in the repo -- your latest PR didn't add it.

rpatters1:
https://github.com/finale-lua/lua-scripts/blob/master/src/finale_lua_menu_organizer.lua

rpatters1:
Oh, now I get it. The bundler should ignore that, because it is embedded in RGP Lua. Do we have a way to make the bundler ignore it?

asherber:
Probably a question for @Nick-Mazuk. I can take a look at the bundler in a bit, see if I can suss it out.

asherber:
If I'm reading things correctly, I think there might be 2 ways of tricking the bundler so it skips your require.

```lua
local osutils = require(("luaosutils"))    -- double parens
local osutils = require([[luaosutils]])    -- string literal
local osutils = require( "luaosutils" )    -- spaces inside parens
```

The bundler is looking specifically for a single or double quoted string, optionally surrounded by (single) parens. Any of the three things above looks to be valid Lua and does not trigger the regex that the bundler uses. I think I would vote for the third as being easiest on the eyes.

asherber:
And I would also add a comment on that line indicating what you're doing and why.

asherber:
Longer term, can `finaleplugin` export a constant with the value `luaosutils`? Then scripts can do `require(finaleplugin.LuaOsUtils)` which also won't trigger the bundler because there are no quotes. 

Actually, that's another solution for now:
```lua
local osutils_name = "luaosutils"
local osutils = require(osutils_name)
```

rpatters1:
That did the trick. I was trying to add a list of ignore-values to `helper.ts` but this solution is fine, at least for now.

rpatters1:
Btw: this new script (combined with just-released RGP Lua 0.66) replaces JW Lua Menu plugin.

rpatters1:
Also, the markdown table didn't work. ;_;

asherber:
It would be nice if the doc generator supported a little more Markdown. It also didn't recognize the URL in my Notes that leads to a gist with more info.

PR #452, opened 2023-02-27
Add document_rename_pdfs.lua

asherber:
See `finaleplugin.Notes` for details. Basically emulating Sibelius tokens for PDF renaming, plus cleaning up certain Finale filename artifacts.

Before -- PDFs printed from Fin 26 on Win 10 to older Adobe driver

After

### DIFF
--- a/src/document_rename_pdfs.lua
+++ b/src/document_rename_pdfs.lua
@@ -1,0 +1,203 @@
+function plugindef()
+    finaleplugin.RequireScore = false
+    finaleplugin.RequireSelection = false
+    finaleplugin.RequireDocument = true
+    finaleplugin.Author = "Aaron Sherber"
+    finaleplugin.AuthorURL = "https://aaron.sherber.com"
+    finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
+    finaleplugin.Version = "1.0.1"
+    finaleplugin.Date = "2023-02-26"
+    finaleplugin.Id = "d9282b18-12ed-488a-b0e2-011a1ba7d5b4" 
+    finaleplugin.RevisionNotes = [[
+        v1.0.1      First public release
+    ]]
+    finaleplugin.Notes = [[
+        The main goal of this script is to emulate the tokens that Sibelius allows when exporting PDFs.
+        Using this script, you can change the names of your PDFs, after they have been created,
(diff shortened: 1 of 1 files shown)


asherber:
Side note: since the user settings files require us to pass in the name of the script, it would be nice if this were easily available to us. Can we add a function to the library to extract this from `finenv.RunningLuaFilePath`? I'd be happy to submit a PR.

rpatters1:
There is a function to split path and filename in `FCString`. Not saying there shouldn't be a library function, though. But it should probably use the FCString function.

asherber:
That's fine, although by the time you pull the extension off, it's less code to do the whole thing in Lua rather than going into and out of \`FCString\`.

rpatters1:
Now that I'm home and can look at the script I see what you are talking about. In fact I already wrote this library function:

`library.calc_script_name` in `general_library.lua`.

asherber:
Excellent, thanks! I'll update this PR.

rpatters1 (approved):
lgtm

rpatters1:
Does `os.rename` open a console window in Windows?

asherber:
Nope

rpatters1:
I merged and then immediately thought of the issue that the parts may not be in score order. And even if you sort the parts in score order in the UI, I'm not sure Finale's internal part numbers are in score order or even sequential. Would it not be more reliable to base it on score staves? (I realize that complicates matters with multistaff parts and parts with duplicate staves, but it seems like a manageable problem.)

If you are willing to to limit the script to RGP Lua, you can use the built-in [`eachstaff`](https://github.com/finale-lua/lua-source/blob/master/built-in-functions/rgplua_built_in_functions.lua) function. Or you can replicate it in your script for JW Lua support.

asherber:
In my testing, `each(parts)` has iterated in the order that parts appear in the Manage Parts dialog, sequentially, no gaps. Even if I delete one part and then add a new one ("Part 6") and change the order, the part numbers wind up being literally the current order in that dialog.

You're right, this might not be score order, but in most cases it will be. IMO, trying to base it on staff order could get very complicated, because there are a variety of ways that staves and parts can be related. 

The only actual mention of "score order" is in the intro notes for the script, and I could change this to be more neutral if you think it's misleading. Everything else refers to "part number".

rpatters1:
Perhaps simply changing the intro to say part number order is fine.

PR #456, opened 2023-03-09
Rgp changes for rgplua 0.66

rpatters1:
This is mostly documentation changes plus a new script, so I'm going to merge this immediately to keep the docs in sync with RGP Lua release.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,8 @@ dist/**/personal*
 
 .vscode/settings.json
 */**/.vscode
+
+# ignore finale_lua_menus.txt because it can be different for every user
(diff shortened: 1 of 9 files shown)


ThistleSifter:
Wasn't sure where to write this, but for consistency with common practice, I think `utils.lrtrim` should be renamed to `utils.trim`.

See [JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim), [PHP](https://www.php.net/manual/en/function.trim.php), [Rust](https://doc.rust-lang.org/std/primitive.str.html#method.trim), [Java](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#trim\(\)), etc

PR #467, opened 2023-03-11
fix alert messages etc.

cv-on-hub:
Use UI::AlertInfo() instead of UI::AlertNeutral() with "User Error" as 2nd parameter to look better on Win. Convert all dialogs to mixin; other minor code optimisations.


rpatters1 (approved):
lgtm

PR #470, opened 2023-03-11
Rgp fix menu organizer comment

rpatters1:
Here goes another try. (Trying to get the menu organizer comments to generate readable html.)


Nick-Mazuk:
If it helps, here's where the notes are parsed on the website: https://github.com/finale-lua/jw-lua-scripts-docs/blob/main/src/lib/components/script-notes/script-notes.svelte

PR #476, opened 2023-03-16
Add allotments to json export/import

asherber:
Allotments require RGPLua v0.66, but all the code is backwards compatible -- if run under an older version, it will just ignore allotments.


rpatters1 (approved):
LGTM. If the plugins doesn't include a class, it will be nil, so you could technically avoid using pcall (in `instantiate_prefs`), but this is obviously safer.

asherber:
Thanks, I realized this when I was testing with the previous RGPLua version, but I thought I'd just leave it as is to be safe.

PR #478, opened 2023-03-16
Add RequireSelection = true

asherber:
Also change one script to use [[ ]] instead of " " for AdditionalMenuItems.

cc: @Nick-Mazuk


Nick-Mazuk:
Thanks!

PR #481, opened 2023-03-19
Create measure_span.lua

cv-on-hub:
Divide single measures or join measure pairs by changing time signatures

### DIFF
--- a/src/measure_span.lua
+++ b/src/measure_span.lua
@@ -3,7 +3,7 @@ function plugindef()
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "http://carlvine.com/lua/"
     finaleplugin.Copyright = "https://creativecommons.org/licenses/by/4.0/"
-    finaleplugin.Version = "v0.17"
+    finaleplugin.Version = "v0.18"
     finaleplugin.Date = "2023/03/19"
     finaleplugin.AdditionalMenuOptions = [[
         Measure Span Join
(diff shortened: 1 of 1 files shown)


rpatters1 (changes requested):
These comments are mostly from reading the notes, and I would like to see clarification on at least the first two of them added to the notes.

- What happens if a measure is not completely filled? How is it joined/split?
- If I and joining [2/4][3/8], the option I'm going to most likely want is [2/4 + 3/8] *shown as* [7/8]. Which option gives me this?
- It's a shame that composite meters are not eligible for splitting. This seems like it could be quite useful. But if you wish, we can save that as a future enhancement.

cv-on-hub:
* incomplete measures are filled with rests before join/divide
* if you want to add "display only" meters do so after merging - this is too particular to make any assumptions automatically
* I'm looking at composite meters now and think it will be easy enough to implement, with a few limitations shortly
* will adjust documentation to match shortly

rpatters1:
About "display-only", I think a pretty safe bet is however many of the largest common denominator. (In the case of 2/4 + 3/8, 7 8th notes.) That should be an option, and anything more complicated than that should (as you say) be a manual change afterward.

This is super important. Raw 7/8 and 5/8 are useless because they do not beam correctly. I would be willing to bet that nearly all 7/8 and 5/8 meters are composites with display-only.

cv-on-hub:
Certainly that's the way that I always use  7/8 and 5/8. This would take a few days to implement, so leave it for now if you like and I'll add it along with composite splitting. (PS: I started working on this quite idly before realising that the approach is really useful, and in some ways an improvement on JW's versions).

rpatters1:
Okay, I will merge now and let you submit another for composite support.

PR #482, opened 2023-03-25
Options as JSON: Resync categories

asherber:
@rpatters1 I took a slightly different tack from what you suggested, for a few reasons.

* I used "resync" in the method names instead of "set", because I think it better expresses what's going on. The expression definition already says "I want to use category fonts/positioning", and the method just makes sure that it actually is.
* I didn't include an optional param for the category ID, because the only use case is making sure that the expression agrees with the category it's already assigned to.
* I made just a single method to work on an expression rather than two, because I can't see a case where an expression uses both category fonts and positioning and you only want to make sure that one of them matches the category.
* I included a second method `resync_expressions_for_category()`, because while it's a simple loop, I expect that this will be the primary use case: programmatically make changes to a category and then propagate those changes to all the expressions in the category.

If you're thinking of use cases that I haven't considered, I'm happy to revisit.

The other change in the PR, wrapping json decode in `pcall`, just lets me display clear error messages when the user tries to import a defective json file, getting rid of the reference to the line number in the Lua code file.

### DIFF
--- a/src/document_options_as_json.lua
+++ b/src/document_options_as_json.lua
@@ -5,8 +5,8 @@ function plugindef()
     finaleplugin.Author = "Aaron Sherber"
     finaleplugin.AuthorURL = "https://aaron.sherber.com"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "2.1.1"
-    finaleplugin.Date = "2023-03-16"
+    finaleplugin.Version = "2.1.2"
+    finaleplugin.Date = "2023-03-25"
     finaleplugin.CategoryTags = "Report"   
     finaleplugin.Id = "9c05a4c4-9508-4608-bb1b-2819cba96101" 
     finaleplugin.AdditionalMenuOptions = [[
@@ -16,6 +16,7 @@ function plugindef()
         action = "import"
     ]]
     finaleplugin.RevisionNotes = [[
+        v2.1.2      Resync expression definitions
         v2.1.1      Add music spacing allotments (requires RGPLua v0.66)
         v2.0.1      Add ability to import
         v1.2.1      Add Grid/Guide snap-tos; better organization of SmartShapes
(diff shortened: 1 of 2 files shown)


rpatters1:
The only prior use case in the repo is creating a new expression with a specific category Id. But I can also imagine it being used to move an expression to a new category id. Those were the use cases I was imagining for the option catid param.

asherber:
I think those use cases already work with existing functions, without having to resync like this. But I'll run some tests.

asherber:
As you noted, some scripts in the repo already create a new expression with a specific category ID (like [cue\_notes\_create](https://github.com/finale-lua/lua-scripts/blob/d5502e7320d4320c47b17d46e15744b5ddf50eb7/src/cue_notes_create.lua#L148)), and those work fine. \*\*\*

For moving to a new category, `AssignToCategory()` followed by `Save()` appears to update position info but not font info. However, if you do `expression.resync_to_category()` after `AssignToCategory()`, it all comes out okay -- once the expression knows the new category ID, the `resync` function can do its thing.

\*\*\* I can think of a different kind of function, maybe `insert_category_fonts()`, that could assist, though. Currently, the script does:

```lua
str.LuaString = "^fontTxt"
    .. tfi:CreateEnigmaString(finale.FCString()).LuaString
    .. exp_name
```

But it would be nice if you could just do `"^fontTxt" .. exp_name` and then use a method to insert the other bits of the Enigma string from the category. That way, the script itself isn't responsible for creating the text font info and inserting it as above. I think this is out of scope here, but I'll keep it in mind.

rpatters1:
Honestly, that code snippet with `fontTxt` you found confuses me. It doesn't seem like it should work. Perhaps @cv-on-hub  can shed some light on it.

asherber:
The second line returns something like `(Times New Roman,4096)^size(14)^nfx(1)`, and `exp_name` is the text for the cue name. So if your cue name is `Ob.`, this gives `^fontTxt(Times New Roman,4096)^size(14)^nfx(1)Ob.` That seems to be just like all the other default expression definitions. 

My suggestion for future development was that it would be nice to just set this to `^fontTxt(Ob.)` or something like that, and then have a method that grab the text font info (and music and number font info) from the current category and do a `gsub` to get the full Enigma string.

cv-on-hub:
> Perhaps @cv-on-hub can shed some light on it.

Sorry - beyond my ken.

asherber:
It's the same approach that Jacob Winkler used in `region_multimeasure_rest_tacet.lua` -- maybe Carl borrowed it from there. 

Robert, what seems odd to you about it?

rpatters1:
I just assumed the PDK Framework function added the font tag, but apparently not. It's fine.

cv-on-hub:
> maybe Carl borrowed it from there.

My code came from a CJGarcia script (somewhere).

PR #486, opened 2023-04-05
Update mixins

ThistleSifter:
This PR mainly includes doc fixes, increased use of `mixin_helper` functions and migrating to the new mixin format.

I also fixed some bugs and made a couple of DRY improvements.

### DIFF
--- a/src/mixin/FCMControl.lua
+++ b/src/mixin/FCMControl.lua
@@ -3,20 +3,21 @@
 --[[
 $module FCMControl
 
-Summary of modifications:
-- Setters that accept `FCString` now also accept Lua `string` and `number`.
-- In getters with an `FCString` parameter, the parameter is now optional and a Lua `string` is returned. 
+## Summary of Modifications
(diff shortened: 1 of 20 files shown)


rpatters1:
It would be nice to develop a set of tests, especially for library functions. I have a massive set I run on RGP Lua itself. (It's in the finale-lua org, forked originally from JW's personal github account.)

I'm thinking it might be especially useful for mixins, due to their interconnectedness. Even better if it could be done without running Finale and have it happen as part of the merge process. But I'd be happy it they could be run in Finale like the RGP Lua tests.

Anyway I'll go ahead and merge these and we'll see if anything breaks. Heads up to @asherber, since he uses the scripts out of the dist folder.

ThistleSifter:
Yes I agree that a testing suite would be great. I've got a few more planned updates that I still need to submit (I don't like having too many half-finished things on the go) and then I'll look at testing afterwards.

PR #489, opened 2023-04-07
Bundle modules with `package.preload`

### LINKED ISSUE #488: `require` in bundled scripts
I've been looking into a better solution for the usage of `require` in bundled scripts and I've come up with the following plan:
- Add all bundled modules as package loaders.
- Remove the overridden `require` function and just use the standard one.

This should mean that the loading of any other modules (eg `luasocket`) will be unaffected by the bundling process.

There are two ways of directly including package loaders.
1) `package.preload`
```lua
package.preload["library.utils"] = package.preload["library.utils"] or function()
-- module contents...
end
```

2) `package.searchers`
```lua
table.insert(package.loaders, function(package_name)
-- return module wrapped in function from here
end)
```

By default, `require` looks for a package loader in the following order (according to `package.searchers`):
1) `package.preload`
2) `package.path`
3) `package.cpath`
 
That order can be changed by reordering the table or inserting a function earlier in the `package.searchers` table for loading bundled modules.

Let me know if you have any thoughts as to which option is better.

(Edited because I initially misread the manual)

Cc @rpatters1

ThistleSifter:
Fixes #488 

Removes the override for `require` and adds all bundled modules to `package.preload`.

@rpatters1

I'm also wondering if a similar approach could work for `luasocket` so that `require` wouldn't need to be overridden in the `lua-source` repo?


rpatters1:
I believe the `require` in lua-source is addressing a different issue, which is the inflexibility of the folder layout that `socket` expectsm for finding its lua components. Anyway, it isn't broken at the moment, so I am in no mood to fix it. But once we have verified this PR (and once I finish some other projects) I might have a look at it.

As part of this PR, please delete all the scripts in `dist`. That will force them to be rebuilt when it is merged.

Thanks. This is a big improvement.

ThistleSifter:
`dist` has been reset.

rpatters1:
i just merged the other PRs, and now there is a conflicting file in `dist`. If you could delete it, I'll do the merge.

ThistleSifter:
Done.

rpatters1:
@asherber I will be interested to hear if this breaks anything. Merging now.

ThistleSifter:
Ah crap, I forgot to run `pnpm build`, so the compiled bundler hasn't changed. I'll do it in the morning.

PR #491, opened 2023-04-07
Replace 2 old MIDI scripts with one

cv-on-hub:
Removing 2 old MIDI scripts, combined into new (better) midi_note_values script


rpatters1:
For future reference, the way you delete files in `dist` is:

- open the `dist` folder on your local copy of the repo.
- delete the files that need deleting.
- check in the repo into your working branch.

PR #501, opened 2023-04-12
Add Lua 5.4 compatibility library

### LINKED ISSUE #493: Has something changed in `mixin` with SetTypeMeasurementInteger? [BUG]
Just noticed in the latest incarnation of `mixin` that dialog.CreateMeasurementEdit() edit boxes display a decimal point even if they are specified as 
`edit:SetTypeMeasurementInteger()`
and
`edit:SetMeasurementInteger(value)`

Unsure if this behaviour has been present before the latest amendments.
@ThistleSifter

ThistleSifter:
Fixes #493 

This should address the Lua 5.4 integer issues. More functions can be added to the `lua_compatibility` library as needed.

I also added `float` and `integer` as types for argument type assertion.

If anyone has a better name than `whole_float_to_integer` I'm happy to change it.

Here's a quick test script:
```lua
dualog:CreateMeasurementEdit(10, 10):SetMeasurementInteger(12)
dialog:CreateMeasurementUnitPopup(10, 50)
dialog:ExecuteModal()

```

### DIFF
--- a/src/library/lua_compatibility.lua
+++ b/src/library/lua_compatibility.lua
@@ -1,0 +1,34 @@
+--  Author: Edward Koltun
+--  Date: April 9, 2023
+--[[
+$module Lua Compatibility
+
+This library assists in providing compatibility across Lua versions by polyfilling standard library functions in older Lua versions.
+
+The following functions are polyfilled:
+```
+Function         | Lua Versions Polyfilled
+-------------------------------------------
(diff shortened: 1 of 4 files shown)


rpatters1 (changes requested):
There's a typo in one of the comments: `br` should be `be`.

How about `integer_if_whole` for the function name?

PR #514, opened 2023-04-30
Mixin Update

ThistleSifter:
Originally I had just intended to add properties but then it became apparent that other changes needed be made in order for that to work so this PR has ended up becoming rather large. I would like to be able to call the `mixin.lua` library stable after this PR, so I've tried to get as many potential issues resolved as I can. Updated documentation will come in a follow-up PR.

## Summary of changes

**Updated Format**
I've slightly modified the mixin format again for improved extensibility and future-proofing. It now looks like this:
```lua
-- All fields are optional except for Parent in FCX mixins
local class = {
    Parent = "FCMParentClass", -- FCX only
    Init = function(self) ... end,
    Methods = {...},
    StaticMethods = {...},
    Properties = {...},
    Disabled = {...},
}
...
...
return class
```

**`class.Methods`**
Methods have moved from what was previously a catch-all 'public' table for pretty much any property/method, to a dedicated table for methods. All fields must be functions.

**`class.StaticMethods`**
Static methods are now defined in a `StaticMethods` table. Static methods cannot be accessed through an object, only via the `mixin` namespace.

**`class.Properties`**
Mixins can now have properties with defined getters and setters. Here's a brief example:
```lua
class.Properties = {
    Foo = {
        Get = function(self) return private[self].Foo end,
        Set = function(self, value) private[self].Foo = value end,
    },
}
```
Although not shown in the above example, the main advantages of having get/set functions is to provide control over writability and being able to validate any assigned values.

Properties are inherited and can be overridden in the same way as methods.

Properties can also be accessed statically in the same way as methods if needed:
```lua
mixin.FCMCustomLuaWindow.SomeCustomProperty.Set(dialog)
```

**Name Clashes**
A method or a static method or a property cannot share the same name, including clashes that occur via inheritance. Clashing names will result in an error during mixin loading.

**`class.Disabled`**
The mixin library itself now supports disabling properties or methods directly, without needing to rely on helper functions. Disabled properties/methods simply return `nil` and cannot be overwritten (attempting to set a disabled property/method will result in an error).
```lua
-- FCMCustomLuaWindow.lua
class.Disabled = {"MethodOne", "MethodTwo", "PropertyOne"}
...

-- script.lua
local dialog = mixin.FCMCustomLuaWindow()
dialog.MethodOne = function() return true end -- Error: No writable member named 'MethodOne'
dialog.PropertyOne = true -- Error: No writable member named 'PropertyOne'
```

**Reflection**
I have added reflection properties to the `mixin` namespace in a similar vein to the `finale` namespace (`__class` for methods, `__static` for static methods, `__propget` for property getters and `__propset` for property setters). I also added an additional property, `__disabled`, for disabled methods & properties. I don't think there's much of a use for this right now but it should allow all sorts of testing to be done without needing to make any changes to `mixin.lua`.

**Error Checking**
Introduced more stringent validation during the loading of mixins and more helpful error messages.

**`personal_mixins`**
`personal_mixins` will now only be loaded if `finenv.TrustedMode ~= true`.

**Original Method**
Referencing an original `FC` method is now done with a double underscore instead of a single (ie `edit:SetText__(str)`) which is more conventional and visually clearer. Let me know if you see any issues with this.

## Issues to Resolve
I did originally have a longer list but I solved most of them as I went along.

**Checking name and type clashes with `FC` classes**
I started adding checks for clashes between mixins and their underlying `FC` classes (ie not overriding methods with properties and vice versa, parent mixin not having method that is introduced in child `FC`, etc) but it started to turn into quite a large undertaking so I took it out. Rather than adding bloat to `dist` scripts, I'm wondering if that would be better placed in a testing suite?

**Reflected Methods**
Should the original methods be returned via reflection or should they be be proxied (not to be fluid, just for automatic mixin enabling of any returned objects)? At the moment they are proxied and I can't think of any reasons for not doing it, but I thought I'd ask anyway in case I missed something.

**Constants**
Not a blocking issue, but I would like to implement custom constants (something like `mixin.MEASUREMENTSUFFIX_SHORT`) but I haven't worked out a good way of doing it at the global level (ie not just per mixin). Maybe just put them in their own file? Let me know if you have any ideas.

cc @rpatters1 @cv-on-hub

### DIFF
--- a/src/library/mixin.lua
+++ b/src/library/mixin.lua
@@ -292,18 +292,33 @@ local library = require("library.general_library")
 local mixin_public = {}
 -- Private methods
 local mixin_private = {}
--- Fully resolved FCM and FCX mixin class definitions
+-- FCM and FCX mixin class definitions
 local mixin_classes = {}
+-- Flattened class definitions for optimised runtime lookup
+local mixin_lookup = {}
 -- Weak table for mixin instance properties / methods
 local mixin_props = setmetatable({}, {__mode = "k"})
 
--- Reserved properties (cannot be set on an object)
+-- Reserved properties, all accessible statically (cannot be set on an object)
 local reserved_props = {
-    MixinReady = function(class) return true end,
-    MixinClass = function(class) return class end,
-    MixinParent = function(class) return mixin_classes[class].meta.Parent end,
-    MixinBase = function(class) return mixin_classes[class].meta.Base end,
-    Init = function(class) return mixin_classes[class].meta.Init end,
+    MixinReady = function(class_name) return true end,
+    MixinClass = function(class_name) return class_name end,
+    MixinParent = function(class_name) return mixin_classes[class_name].Parent end,
+    MixinBase = function(class_name) return mixin_classes[class_name].Base end,
+    Init = function(class_name) return mixin_classes[class_name].Init end,
+    __class = function(class_name) return mixin_private.create_method_reflection(class_name, "Methods") end,
+    __static = function(class_name) return mixin_private.create_method_reflection(class_name, "StaticMethods") end,
+    __propget = function(class_name) return mixin_private.create_property_reflection(class_name, "Get") end,
+    __propset = function(class_name) return mixin_private.create_property_reflection(class_name, "Set") end,
+    __disabled = function(class_name) return mixin_classes[class_name].Disabled and utils.copy_table(mixin_classes[class_name].Disabled) or {} end,
+}
+
+-- Reserved properties that are accessible from an object instance
+local instance_reserved_props = {
+    MixinReady = true,
+    MixinClass = true,
+    MixinParent = true,
+    MixinBase = true,
 }
 
 -- Create a new namespace for mixins
@@ -319,11 +334,23 @@ local mixin = setmetatable({}, {
         mixin_public[k] = setmetatable({}, {
             __newindex = function(tt, kk, vv) end,
             __index = function(tt, kk)
-                local val = reserved_props[kk] and utils.copy_table(reserved_props[kk](k)) or utils.copy_table(mixin_classes[k].public[kk])
-                if type(val) == "function" then
-                    val = mixin_private.create_fluid_proxy(val, kk)
+                local value
+
+                if mixin_lookup[k].Methods[kk] then
+                    value = mixin_private.create_fluid_proxy(mixin_lookup[k].Methods[kk])
+                elseif mixin_classes[k].StaticMethods and mixin_classes[k].StaticMethods[kk] then
+                    value = mixin_private.create_proxy(mixin_classes[k].StaticMethods[kk])
+                elseif mixin_lookup[k].Properties[kk] then
+                    -- Use non-fluid proxy
+                    value = {}
+                    for kkkk, vvvv in pairs(mixin_lookup[k].Properties[kk]) do
+                        value[kkkk] = mixin_private.create_proxy(vvvv)
+                    end
+                elseif reserved_props[kk] then
+                    value = reserved_props[kk](k) -- reserved_props handles calls to copy_table itself if needed
                 end
-                return val
+
+                return value
             end,
             __call = function(_, ...)
                 if mixin_private.is_fcm_class_name(k) then
@@ -337,7 +364,6 @@ local mixin = setmetatable({}, {
         return mixin_public[k]
     end
 })
-
 
 function mixin_private.is_fc_class_name(class_name)
     return type(class_name) == "string" and not mixin_private.is_fcm_class_name(class_name) and not mixin_private.is_fcx_class_name(class_name) and (class_name:match("^FC%u") or class_name:match("^__FC%u")) and true or false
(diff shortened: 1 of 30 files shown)


rpatters1:
I'm a bit puzzled by the "not loading personal_mixins if `finenv.TrustedMode ~= true`". If TrustedMode isn't true, mixins wont work at all, correct? The reason I ask is that maybe there is a misunderstanding about what `finenv.TrustedMode` does.

The "reflection" that is built into finale classes is not true reflection. It basically only reflects the names of methods and properties. (Though those can then be used to get the actual functions.) Which is all by way of saying I don't see a problem with proxies at the moment.

I'm all for keeping bloat down in `dist` where possible. I'm starting to notice performance hits with some of the bigger ones, at least on my slow Win machine. I think a test suite makes a lot of sense.

A mixin constants file is probably fine. A bigger question is can you make them non-writable like the LuaBridge ones are? If not, it's not a huge deal.

The further down this "trusted code" rabbit hole I've gone, the more I would like to see scripts not be trusted when they don't have to be. For example, I've added various functions for common things like launching a web page that are safe whereas otherwise a script would have to use an unsafe function like `os.execute` to accomplish it. That said, the hash checking scheme for this repo is working well, and I don't have any concerns about any of them. Still, if your refactoring is leading you away from the idea of modifying the metatables, do let us know.

cv-on-hub:
I'm not competent to comment on deeper issues but am running your new `library` & `mixin`  directories in my active folder looking for incompatibilities. Preliminary checks are clean. (It's also running RGP 0.67b4).

ThistleSifter:
"not loading personal_mixins if finenv.TrustedMode ~= true"
It's the other way around: if it is true, they won't be loaded. But that doesn't interfere with the mixins bundled with a script, just ones loaded from the user's local machine. If I've misunderstood something, I can just change it to not load `personal_mixins` if `package.preload["library.mixin"]` exists (ie it's bundled).

When you say you're noticing a performance hit, do you know where it's occurring or why? Is it just from the large file size of `dist` script or something else?

This lot of refactoring was about solidifying the way mixins are defined and it doesn't change the way that mixins work as far as metatables go. I don't really have a personal preference regarding metatables vs wrapper tables. Both ways have their pros and cons and their usage would look slightly different but either way can work. Do you have a preference or arguments to be made in either direction?

rpatters1:
As currently written, the `TrustedMode` boolean does not distinguish between a script that is trusted because of the user explicitly trusting it through the config UI and a script that's trusted because its hash url was verified from the repo. It may not be the right flag for deciding whether to include personal mixins. I'm open to changing it, perhaps to an enumerated type. That would make it more useful for this, I believe.

The performance problem is literally just the file size. I mostly work on Mac, but I have a super cheap Win machine that I use to build and test the Win versions of things. It is quite slow, and the load times for the `dist` versions that use mixins are noticably slower that the `src` versions. I also noticed an issue when RGP Lua is scanning the file for the `plugindef` function. It was only bad when RGP Lua has to do it for the all the scripts in the `dist` folder at once. Then it was slightly noticeable even on my Mac.

If we could dispense with modifying the metatables for mixins, then RGP Lua could simply never allow them to be modified, which as I've gotten further into this would be my preference. Another advantage would be that then none (or almost none) of our scripts would have to run in trusted mode. It would obviate the need for us to worry about when to exclude personal mixins. I would still retain the hash verification, because that's going to be useful for checking for new versions.

I hate to be a yo-yo on this topic, which is how I feel I am being. My views are evolving. No matter what, my goal is to support mixins whatever that entails. I am not sure we'll ever really use them in the chameleon like manner that was originally proposed (or that we even should), so perhaps the metatable approach is overkill.

ThistleSifter:
On the issue of file size, is there some other way of distributing mixins, eg bundled in a library file? The file size issue is just going to get worse and it's not going to be long before it becomes a block for adding new mixins. I do have another PR mostly done that has generic proxy functions which will save maybe 200 lines of code across the current lot of mixins but that's not really a solution.

I get the yo-yo thing, I've done that a few times since the issue with metatables came to light. From memory there were a couple of issues I ran into with wrapping tables, namely:
- How to handle "unwrapping" the object (see below)
- Is there a limit of one wrapper per object?
- What happens if something tries to "rewrap" an "unwrapped" object? (Somewhat tied to the issue of limit for wrappers, and especially messy if the existing wrapper has already been subclassed into an FCX)
- Some mixins completely override the underlying functionality (eg restoring control state). Any direct calls to an "unwrapped" object could potentially entirely break that new functionality. Maybe classes can specify if they permit or deny "unwrapping"?
- I haven't dug too deeply into forwarding non-overridden method calls to the underlying object so I don't know if there are hidden gotchas there

The 3rd and 4th ones were the main stumbling blocks for me.

rpatters1:
I just released 0.67. The documentation for finenv.TrustedMode should show you what to do. Let me know if you have questions.

ThistleSifter:
I've updated the TrustedMode checks. If you could take a look at line 444 in `mixin.lua` and see if that looks correct, that would be great.

I'm working on a replacement for the mixin library that uses wrapping tables. There's a lot to rewrite/update though so it will probably take at least a few weeks.

On a side note, is `require_embedded` still needed? I thought that removing the overridden `require` from the bundler would have obviated the need for it.

rpatters1:
`require_embedded` evades the parser that looks for `require` statements to convert into `package.loaded` statements. I believe it is still needed, but I would like to be wrong.

PR #515, opened 2023-04-30
Update note_entry.lua

cv-on-hub:
A good idea to include PerfMod deletions?

also ...

in `note_entry.rest_offset()` setting offset to zero is different from wanting a `FloatingRest`. Shouldn't there be an additional boolean input: `zero_is_floating`
???

### DIFF
--- a/src/library/note_entry.lua
+++ b/src/library/note_entry.lua
@@ -322,6 +322,13 @@ function note_entry.delete_note(note)
         finale.FCTieMod(finale.TIEMODTYPE_TIESTART):EraseAt(note)
         finale.FCTieMod(finale.TIEMODTYPE_TIEEND):EraseAt(note)
     end
+    if entry.PerformanceDataFlag then
+        local perf_mod = finale.FCPerformanceMod()
+        perf_mod:SetNoteEntry(entry)
+        if perf_mod:LoadAt(note) then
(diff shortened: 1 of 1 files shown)


cv-on-hub:
Actually, to limit breakage, the additional boolean input should be: `zero_not_floating `

rpatters1:
Personally I dislike boolean inputs and prefer to avoid them when possible. Doesn't the PDK have a call to make it floating? Does that need to be duplicated here?

cv-on-hub:
`entry:SetRestDisplacement()`automatically makes the rest movable, not floating. An explicit `entry:SetFloatingRest(true)` could reasonably be left up to the user.

rpatters1 (changes requested):
I would like to see this change replaced with a single line, similar to the rest of the function:

```lua
finale.FCPerformanceMod():EraseAt(note)
```

The `EraseAt` method takes care of all necessary checks. (Actually, we should not check the flag because it isn't reliable for removing cruft, but `EraseAt` takes care of it.)

rpatters1:
I'm a big believer in if it ain't broke don't fix it. Is the function not doing what it needs to now?

rpatters1:
I will go ahead and merge this. If `rest_offset` needs further attention it can be a separate PR.

PR #519, opened 2023-05-05
Update note_entry.lua

cv-on-hub:
`rest_offset()` is broke. You can’t currently set a rest to `zero` position if another layer is present - it will automatically “float” to the optional layer float position. “Zero” and “floating” aren’t equal.

I added this function to the library last August based on my `rest_offsets.lua` script (with algebraic help from @rpatters1). Subsequently determined that `zero_is_floating` is a significant problem and updated the script but not the library function. (The script can't use the library version in its current form). The boolean input is the only sensible way I can imagine to make the distinction. No scripts in the repo use this function yet.

### DIFF
--- a/src/library/note_entry.lua
+++ b/src/library/note_entry.lua
@@ -449,13 +449,14 @@ Confirms the entry is a rest then offsets it from the staff rest "center" positi
 
 @ entry (FCNoteEntry) the entry to process
 @ offset (number) offset in half spaces
+@ [zero_is_floating] (boolean) optional flag to make zero-offset rests "float"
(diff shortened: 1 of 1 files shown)


cv-on-hub:
PS: the other option is simply to omit any reference to `SetFloatingRest()` and leave that up to the caller.

rpatters1:
If nothing is calling the function, I'd say change it so your code can use it. I don't understand why `rest_offset` would ever set floating rest. If we need a function for that, it should be a separate function.

cv-on-hub:
Cool. Easy fix.

PR #523, opened 2023-05-14
Create deletion_chooser.lua

cv-on-hub:
A re-imagining of `delete_selective.lua`.

I wanted to add new items to `delete_selective.lua` but it grew up to 20 menu items and that's too many even for me. This is an experiment using a single-menu script to present a macro-palette dialog to choose the 20 items, with configurable "code" keys, mimicking Keyboard Maestro's palettes (cf: @Nick-Mazuk). Call the script, type the key, hit `Return`.

These are the items that I most commonly like to delete independently. Obviously the list is as long as a piece of string but this feels like a reasonable compromise - inclusivity vs prolixity.

### DIFF
--- a/src/deletion_chooser.lua
+++ b/src/deletion_chooser.lua
@@ -3,8 +3,8 @@ function plugindef()
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "http://carlvine.com/lua"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "0.71"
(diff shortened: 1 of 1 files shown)


cv-on-hub:
Interesting to note that while entry `ReverseStem` flags are present `FCPrimaryBeamMods` are at least partially masked and won't "delete".

rpatters1 (approved):
This looks good to me. It would be nice if the keystroke itself closed the window, but I imagine that would require a new RGP Lua version.

cv-on-hub:
> It would be nice if the keystroke itself closed the window

Looked as hard as I could for a solution with no luck.

PR #526, opened 2023-05-24
Update staff_explode.lua

cv-on-hub:
The whole `staff_explode.config.txt` inside `script_settings` thing always felt messy. This revision allows setting the `respacing` pref by holding down the shift key without increasing menu numbers

### DIFF
--- a/src/staff_explode.lua
+++ b/src/staff_explode.lua
@@ -3,8 +3,8 @@ function plugindef()
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "http://carlvine.com/lua/"
     finaleplugin.Copyright = "https://creativecommons.org/licenses/by/4.0/"
-    finaleplugin.Version = "v1.55"
-    finaleplugin.Date = "2023/04/30"
+    finaleplugin.Version = "v1.56"
+    finaleplugin.Date = "2023/05/24"
     finaleplugin.AdditionalMenuOptions = [[
         Staff Explode Pairs
         Staff Explode Pairs (Up)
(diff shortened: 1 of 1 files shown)


rpatters1:
The latest version of RGP Lua can access the Automatic Music Spacing preference in Finale. It seems like that should be the indicator rather than a one-off option here.

rpatters1:
Correction: the latest version of the plugin *with* Finale 27.3 can access the Automatic Music Spacing preference.

cv-on-hub:
> latest version _with_ Finale 27.3 can access Automatic Music Spacing

so _IF_ RGP0.67 _AND_ Fin27.3 then one can momentarily set FCGeneralPrefs.AutomaticMusicSpacing, do the work, then reset the user's original preference? Seems like all other conditions still need the warning message, and that users may want to save this behaviour uniquely for `staff_explode.lua`, still requiring `user_settings` file.

cv-on-hub:
Sorry - I've been thinking about this tangentially for too long. User choice at `Finale->Settings->Edit->Spacing` will enable/disable spacing regardless of the script settings. Will revise removing spacing options but with that advice at `plugin.Notes`

rpatters1:
Actually, what I was imagining was that you would base whether or not to do the spacing based on the user's choice in Automatic Music Spacing. That is, if it's set, respace after explode. If it's not set do not respace after explode. No need for a separate setting.

cv-on-hub:
If the User has chosen Automatic Music Spacing then it looks like it will happen anyway even if the script doesn't do so.

cv-on-hub:
(at least in Fin 27.3)

PR #528, opened 2023-05-28
Fix name of function in docs

asherber:
Incorrect function name given in https://www.finalelua.com/docs/mixin/FCMCtrlCheckbox

I assume the website docs will regenerate after this is merged, right?


rpatters1:
They should.

PR #529, opened 2023-05-29
Create macro_key_script_palettes.lua

cv-on-hub:
I'm not sure this is worth posting - intended primarily  as a proof of concept, and may be better waiting until (if ever) keystrokes can simulate an "OK" action in CustomWindows. More info at the top of the script.


rpatters1:
I'm not sure keystrokes will ever be able to simulate the Ok button. The easy part is notifying windows of keystrokes, but providing a close window function is tricky to do from within Lua.

cv-on-hub:
I suspected that would be the case. If this _KEY_ approach has any perceived usefulness there are a couple more mods I'd implement before launching.

PR #530, opened 2023-05-31
Create hotkey_script_palettes.lua

cv-on-hub:
So what the hey - this includes all the additions I’m going to make to this script and despite the redundant [Enter] keys I kinda like the way it works. Will post it in case it is useful to others or inspires some other approach. (And “hotkey” is a more informative title than “macro key” - thanks @asherber)


asherber:
I can see that a lot of work went into this; it's a neat idea, and I'm fascinated by some of the code. I've only had time to play with it a little, but here are some observations.

- When I think of palettes, I think of a modeless window that's in view until I dismiss it and that I can use whenever I want, but this script opens a modal window. This makes it a little less useful for exploring plugins (IMO), because if I'm going to wind up executing one that requires a region selection, I have to make the selection before I open the palette.
- The hotkeys only function within the palette if the list box has focus, but when the window opens, the box does not have focus; I have to hit tab first to focus the box. The same is true of the second window. (Maybe this is different on Mac?)
- On the code side, I'm curious about why you chose a custom delimited format for storing the configuration instead of just storing it as json, which I think would be easier to move back and forth between tables and text.

rpatters1:
I can answer the third bullet. Configuration must

- be simple to use
- use human-readable files
- have zero external dependencies

So I rolled a mini-lua-table parser. Even now with cjson embedded, I couldn't achieve the second bullet using json. (This is a flaw in cjson: no formatted option like dkjson. I've thought about patching it.) Zero external dependencies is a drop-dead requirement for this, but tbh it was done before I had added lunajson to the repo. (But I don't know if lunajson has a formatted option. I only know that dkjson does.)

I have no strong feeling about converting it to json, but it ain't broke.

asherber:
@rpatters1 I think we may be talking about different things. I think you're talking about the format of configuration files -- I'm talking about the specific serialized format that Carl is using to store his palettes.

```
local config = { 
    main = "Automations`A`1^Chords & Notes`C`1^Expressions & misc.`E`1^Intervals`W`1^Layers etc.`1^Measure Items`B`1^",
    ...
}
```

That delimited string could have been a json string instead.

rpatters1:
Oh, I see. Yes, and with cjson embedded it would not require any dependencies to parse it.

cv-on-hub:
1. This was envisaged as an end-point process, not a script "explorer". I thought the original LUA menu would suffice for that. I hate modeless windows - I want them all out of the way after use. (You can obviously change that behaviour).
2. Mac always gives focus to the first dialog. In the following version I've added a "KeyboardFocus" to the main ListBox. Hope that works for Win.
3. Didn't use JSON because I didn't know how or what it was. I discover now that it's wonderfully easy. Only thing is to use it with `configuration.lua` need to escape all the quotes. Can't work out how to do that within `cjson`, or should I just do escaping before storage?

asherber:
I don't think you would need to escape the quotes.

```lua
local configuration = require("library.configuration")
local library = require("library.general_library")
local cjson = require("cjson")

local t = {
    name = "Macro Palette 1",
    key = "A",
    last = 1
}

local config = {
    main = cjson.encode(t),
}
local script_name = library.calc_script_name()
configuration.save_user_settings(script_name, config)
```

That produces:

```
-- User settings for playground.lua

main = "{"last":1,"name":"Macro Palette 1","key":"A"}"
```

Similarly, you can use `cjson.decode(config.main)` to turn it back into a lua table. Seems to work for me.

cv-on-hub:
Hmmm. Didn't work for me but I'll try again. Following script uses `cjson` and has the extra keyboard focus.

asherber:
Even in lua code, I don't think you need to escape the quotes -- just use single quotes for the outer ones.

```lua
local config = {
    json = '{ "name": "foo", "last": 1 }'
}
```

It would be even nicer if you could use a multiline string:

```lua
local config = {
    json = [[
        { 
            "name": "foo", 
            "last": 1 
        }
    ]]
}
```
This seems to work fine for saving values, but not for reading them -- the configuration parser doesn't know how to handle the newlines.

cv-on-hub:
You're right again. New version without escaped quotes, and using the embedded `cjson`. (Documentation is a little scant).

cv-on-hub:
Thanks for that Robert. MinJWLua updated. I'm now employing `list_box:SetKeyboardFocus()` on the primary dialog.

cv-on-hub:
I'm working on a version of this that includes the option of adding other Finale menu items to the palettes. Will close this PR until the new version is ready.

rpatters1:
I haven't analyzed your code, so I'm not certain if this comment matters, but just in case here goes:

- Plugin menu items have unreliable command codes. The codes will change depending on which and how many plugins are loaded in Finale.
- Lua menu items have reliable menu text, so you can use menu text to find menu items for Lua scripts. I would consider using `FCLuaScriptItems` to get the menu text, so that the palette code is not dependent modifications to the menu text in the scripts over time. Another option would be to read the text straight out of the menus, using `luaosutils.menu`. But also be aware that items may have beem moved to custom menus, so `FCLuaScriptItems` might be easier. 
- Menu text for non-Lua scripts (including other plugins and base Finale functions) is *not reliable*. The text will change depending on locale. For non-plugin menu items, you can use the menu items' command codes.

rpatters1:
Just as an example, I move my RGP Plugins into their own menu called RGP (in the main menu). I also move all my RGP Lua scripts into a menu called Lua and I put all the JW plugins into a menu calle JW. See `finale_lua_menu_organizer.lua`.

cv-on-hub:
Yes - it was inspecting your `finale_lua_menu_organizer.lua` that led me down this path in the first place ... (!)

PR #533, opened 2023-06-02
Update staff_rename.lua

cv-on-hub:
These two little changes fix the error under `RGPLua 0.67`

att: @jwink75


rpatters1 (commented):
FWIW: I've gotten so annoyed by this little `tostring` incompatibility with older versions that I've have fixed it in future versions. That is, you will be able to assign a number directly to `LuaString` again in RGP Lua 0.68.

rpatters1:
This pull request is no longer compatible with the master branch. Please update it, and I'll merge it.

PR #539, opened 2023-06-19
Update measure_span.lua

cv-on-hub:
Just noticed that `IncludeInNumbering` was being set false (erroneously) for all newly created measures. Assorted other code optimisations.


rpatters1:
I've noticed frequently that the description in the PR doesn't begin to describe all the changes in the files. Just for documentation, I would prefer to see descriptions that list a summary of all the changes.

PR #541, opened 2023-06-27
Add measure_hide_empty

asherber:
@rpatters1 As it turns out, testing shows that Finale seems to reorganize its internal database of staff style assignments after each `Save()`, so that consecutive measures with the same staff style wind up under the same assignment. So if assign a style to m. 1 and then m. 3 and then m. 2, and then I do `assign:LoadAllForItem()`, I'll see just one assignment, to mm. 1-3. 

In the interest of efficient code, I opted to assign the Hide Staff style to consecutive blocks of measures in one go, rather than measure by measure, but I didn't bother with looking at adjacent ranges, because Finale will munge them all together for me.

### DIFF
--- a/src/measure_hide_empty.lua
+++ b/src/measure_hide_empty.lua
@@ -1,0 +1,136 @@
+function plugindef()
+    finaleplugin.RequireScore = true
+    finaleplugin.RequireSelection = true
+    finaleplugin.RequireDocument = true
+    finaleplugin.MinJWLuaVersion = 0.65
+    finaleplugin.Author = "Aaron Sherber"
+    finaleplugin.AuthorURL = "https://aaron.sherber.com"
+    finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
+    finaleplugin.Version = "1.0.1"
+    finaleplugin.Date = "2023-06-26"
+    finaleplugin.Id = "4aebe066-d648-4111-b8b3-22ac2420c37d"
+    finaleplugin.RevisionNotes = [[
+        v1.0.1      First public release
+                    Apply to to consecutive measures at once
+        v0.9.6      Fix bug with single staff style
+        v0.9.5      Simplify: Finale combines assignments for us
+        v0.9.4      Minimum dialog width
+        v0.9.3      Fix table index bug
+                    Use both adjacent assignments if available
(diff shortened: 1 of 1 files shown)


rpatters1:
Remind me what the use case for this script is.

I see a couple of minor issues, one of which may be on purpose:

- As it stands, the staff style is applied only to the current score/part in edit focus. This is a logical choice, but I just want to confirm that is your intention.
- Most "Hide Staff" styles can be applied to partial measures. Perhaps you should not use the mere existence of an assignment to a measure as a basis for excluding it. (But I could be persuaded.) 
- `FCStaffStyleAssign` automatically initializes the end measure position to `INT32_MAX` (i.e., the maximum possible measure duration position.) That means your code works as written for full measure assignment. That said, I am wondering if the script should do partial measure assignment on the first and last measures of the selected region, if they are partially selected.

I am ready to merge the script as-is. Let me know your preference.

asherber:
This was written to satisfy a request from the FB group, and then gussied up a little. Here's the [original post](https://www.facebook.com/groups/finalelua/permalink/1709098392894199/), and here's [an image](https://www.facebook.com/groups/finalelua/posts/1709098392894199/?comment_id=1709216832882355) of what he was after.

The issues you raise are valid; my current thoughts are:

- I think current score/part is probably correct; in fact, I think this sort of thing really only makes sense in the context of a score. If you've got a use case where it makes sense in parts as well, let me know and I'll incorporate it.

- Regarding your two points about partial measures, I think the intent of the request was just for full measures. In order to work on partial measures, I think I'd have to change from the current approach of looking for measures with no entries and include partial measures (just at start and end?) with rests.

I think I'm happy with what the script currently does. As I said, if you've got a good case for a related use, I'll take a look at extending it. 

If no changes are desired, I can also edit the Notes to make it clear that it works only on full measures, and only in the current score/part.

rpatters1:
I don't need any changes. I guess the use case is auto-cutaway score. Personally I think it looks like crap. (Not cutaway, per se, but I think it requires more judgement to decide what to cut away.) But I'm happy to merge as-is or with updated Notes.

asherber:
It's not a look I have use for either, but apparently someone does. I'll revise the notes in the morning.

asherber:
Okay, I've amended my commit.

PR #544, opened 2023-07-09
Create hotkey_script_palettes.lua

cv-on-hub:
A LUA version of the FinaleSuperUser Hotkey Palette paradigm. Includes changes as previously discussed.


rpatters1:
This line:

```lua
local cjson = require("cjson")
```
breaks the deployment build. You should use this code instead:

```lua
local utils = require("library.utils")
local cjson = utils.require_embedded("cjson")
```

PR #548, opened 2023-07-20
Update hotkey_script_palettes.lua

cv-on-hub:
Sorry about this - listing my `info` variable at the top of the script ruined the deployment build. More sober approach here should work.


rpatters1:
As you discovered, the `plugindef` function can't have (external) dependencies. This is clearly stated in the docs (going back to JW Lua days.) However, there are a lot of docs, so I don't fault you if you didn't see it.

cv-on-hub:
Yes - I missed that nicety. Also forgot that I can't have tab characters in the `Info` dialog window contents.

PR #571, opened 2023-08-21
Update measure_span.lua

cv-on-hub:
Important correction for smart_shape shifting to correctly handle note-attached shapes. Clarified duplicate expressions. Clarified documentation.


cv-on-hub:
> Have you considered using `FCNoteEntryCell::MoveEntry`

Great idea - thanks! If it really does shift note-shapes along with the notes then that would save a lot of redundant code! And possibly some cruft that my current version probably leaves here and there. The `moved` notes need to be placed in the second half of the target.

But I've just noticed that "doubling" the denominator of the lead measure followed by Rebar does a fair job of shifting notes and note-shapes as well. (Though not measure-shapes). Will investigate!

PR #578, opened 2023-08-24
Update measure_span.lua

cv-on-hub:
Useful breakthrough letting `region:RebarMusic()` do the heavy lifting. Added optional re-beaming after re-barring. Several other key corrections and improvements. Could remove much redundant code and overall behaviour greatly improved.


rpatters1:
Another function that may be useful is `FCNoteEntryLayer.ReduceEntries`. This is a coming attraction in 0.68. It combines tied notes and sequences of rests to match Finale's rules for how they should be presented in standard notation.

This function has been in the PDK all along. I'm not sure why Jari never hooked it up, but I'm hooking it up in 0.68. (Along with some other obscure PDK functions that were left out as well.)

cv-on-hub:
> This is a coming attraction in 0.68

Kewl.

PR #580, opened 2023-09-02
Update measure_span.lua

cv-on-hub:
Finally squashed a long-term gremlin of smartshapes failing to affix properly to newly inserted measures on `join` span. (I think). Minor code fixes.

### DIFF
--- a/src/measure_span.lua
+++ b/src/measure_span.lua
@@ -461,16 +461,15 @@ function shift_divided_measure_shapes(rgn, measure_num, new_duration)
     end
 end
 
-function shift_divided_measure_expressions(measure_num, old_duration, old_width, new_duration, new_width)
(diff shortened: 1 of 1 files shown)


rpatters1:
What's the status of this? you've added several changes since you created the PR. Is it set now?

cv-on-hub:
> What's the status of this?

Initially I needed to fix some serious inadequacies. Those are now stable but  I keep discovering expression types that don't shift correctly, so it might be best to leave it for a couple of days. I won't update again here until I've done everything I can. Hopefully the repo `update` system will be back working by then!

cv-on-hub:
Trial method six is the charm. Mid-measure tempo markings will now reposition with adequate stability. New version is ready for absorption.

cv-on-hub:
This script won't be updated again for the foreseeable future, and the sum of the submissions here correct substantial bugs in the version provided on `dist`.

Issue #586, opened 2023-09-19
[BUG] Mixin library gives error for `FCXCustomLuaWindow.RegisterHandleControlEvent`


rpatters1:
**Describe the bug**
Whenever I call `RegisterHandleControlEvent` on `FCXCustomLuaWindow` it gives an error: FCMCustomLuaWindow.lua:599: attempt to call a nil value (method 'RegisterHandleControlEvent_')

**To Reproduce**
Run this code:
```lua
local mixin = require("library.mixin")

local dlg = mixin.FCXCustomLuaWindow():SetTitle("Test Handler")
local ctrl = dlg:CreateEdit(10, 10)
dlg:RegisterHandleControlEvent(ctrl, function(control) print("got here") end)
dlg:CreateOkButtong()
dlg:ExecuteModal()
```

@ThistleSifter

ThistleSifter:
The call in line 599 should have a trailing double underscore instead of a single.

I can submit a fix but I need to fix my commit history first, hopefully in the next day or two.

rpatters1:
A couple of days is fine. I've got another issue preventing from using mixins for this dialog. I have added a new control type for the dev version of RGP Lua, and mixin is choking on it. I could share the error message, but I'm not sure it would do much good without releasing a copy of the working dev version of RGP Lua.

ThistleSifter:
I'm going to take a guess here and hope this solves the issue (even if it doesn't, it would still need to be done anyway).

Every new control type needs to have a method added to the mixin to trigger the setup and have it work properly. If you look in `FCMCustomWindow.lua` at line 310, there is a list of controls grouped by the number of parameters that the non-mixin `Create*` method accepts. Most just take `x` and `y`, so they fall under `[2]`. All you need to do is add the name of the control to the list as it appears in the `Create*` method name. And add a docblock for it to the list of docblocks above.

rpatters1:
I went ahead and fixed the issue with RegisterHandleControlEvent. I should be merging it soon.

rpatters1:
I added the new control name, as @ThistleSifter  suggested, and that fixed the problem with the new control. It will probably be necessary at some point to add a `FCM...lua` library for it as well, but for now it is working.

PR #587, opened 2023-09-20
OpenAI Hyphenation

rpatters1:
Provides automatic hyphenation of lyrics using OpenAI apis. (By default it uses GPT-3.5)


rpatters1:
closes #586

PR #590, opened 2023-09-23
Add option to use abbreviated staff name

asherber:
@cv-on-hub I would like to suggest this addition to Create Cue Notes. It gives the user the option of using the abbreviated staff name instead of the full staff name when creating a new cue name. There'a a checkbox on the "New Cue Name" dialog to toggle back and forth, and the user's choice is saved as a preference.

The default is not to abbreviate, to preserve current functionality.


cv-on-hub:
Thanks @asherber - a good idea. I had moved on from the old version 0.88 you started with. I thought that `mixin` didn't handle `dialog:RegisterHandleControlEvent()`, and that command certainly still aborts on my system.

I offer a revised version 0.92b (if I can work out how to add a new commit here).

asherber:
@rpatters1 Is that a Mac thing? The code runs fine on my Windows machine. 

@cv-on-hub I'm out tonight, will look at your code in the morning.

rpatters1:
I just merged a fix for that problem a day or two ago. If @cv-on-hub  merges the master branch on this branch, it should go away.

cv-on-hub:
> If @cv-on-hub merges the master branch on this branch, it should go away.

I did and it does.

asherber:
@cv-on-hub I looked at your 92b, and it's fine with me as far as working in the "abbreviate" checkbox. But I did notice that when the dialog first opens, the edit box is no longer focused -- even if I add an explicit `SetFocus()` It seems to have something to do with the Info button you added.

Adding the following code solves the issue:

```lua
dialog:RegisterInitWindow(function(self)
    the_name:SetFocus()
end)
```

cv-on-hub:
The loss of focus seems to be on the Win side, but that addition looks good to me. Do you want to submit that as an update to this PR?

asherber:
I took care of it. I also reverted to `dialog:RegisterHandleControlEvent()`, now that it's working for you, just to keep all 3 of the handler calls parallel.

cv-on-hub:
Great. Do you want to append `local` to `abbrev_name =` at Line 136?

rpatters1:
Somebody post a message when y'all are ready for this to be merged.

asherber:
Thanks, I missed that -- it's fixed now.

I also noticed that the text in the Info dialog had an extra space at the beginning of each line. I added a commit to fix this; it's not the most elegant code, but it gets the job done. (Might have been easier with real regular expressions.)

```lua
local formatted = finaleplugin.Notes
    :gsub("\n%s*\n%s*", "xxx")   -- replace all paragraph breaks with a placeholder
    :gsub("%s+(%w)", " %1")      -- replace all multiple whitespace chars
    :gsub("xxx", "\n\n")         -- put the paragraph breaks back
    :sub(2)                      -- there's still one space at the beginning
```

cv-on-hub:
I certainly don't see any space problems on the Mac side. Is it possibly a Win CRLF text encoding thing? It's disturbingly inelegant!

asherber:
It's not a Windows thing, because the scripts get checked out of the repo with Linux/Mac line endings. 

Here's what your replacement looks like:

image

Note the slight indentation at the start of each paragraph. Here's what it looks like with my version:

image

It's a small effect, so it's easy to miss. Your replacement basically condenses multple spaces to a single space. But the string starts with multiple spaces, so it will start with a single space after replacement. And each paragraph inside the string likewise starts with multiple spaces and so starts with a single space after replacement.

But I did find a neater way of doing this, and I updated the PR:

```lua
local formatted = finaleplugin.Notes
    :gsub(" %s+", " ")   -- replaces all multispaces
    :gsub("\n ", "\n")   -- removes the extra space at the start of internal paragraphs
    :sub(2)              -- removes the extra space at the start of the string
```

cv-on-hub:
Interesting ... it seems that the Mac side automatically suppresses spaces leading a paragraph. (i.e. no visible difference in the old and new `:gsub()` versions). But that's a much more elegant solution for Win. Please ask Robert to go ahead and merge when you post the `gsub` revision.

asherber:
@rpatters1 All yours!

PR #592, opened 2023-09-26
mixin fix to prevent button creep in macOS dialogs

### LINKED ISSUE #588: [BUG] InitWindow routine causes controls to wander with `FCXCustomLuaWindow`
**Describe the bug**
The macOS version of the PDK Framework cannot recreate the controls on a dialog in the same position. (The Cocoa framework requires idiotic transformations because it measures height coordinates from the bottom up instead of top down.) That means the controls wander around each time you open it with `RetainLuaState` set true. I'm not sure what the best fix is, but I wanted to document it. Some controls wander more than others, or wander differently than others.

**To Reproduce**
1. Execute the code below on a Mac.
2. Close the window and reopen it.
3. You will see that the edit text and label have moved down.

```lua

if not dlg then
    dlg = mixin.FCXCustomLuaWindow()
                :SetTitle("Test Control Wandering")
    local lyric_label = dlg:CreateStatic(10, 11)
                :SetWidth(40)
                :SetText("Lyric:")
    local popup = dlg:CreatePopup(45, 10)
                :SetWidth(70)
                :AddString("Verse")
                :AddString("Chorus")
                :AddString("Section")
    local lyric_num = dlg:CreateEdit(125, 9)
                :SetWidth(25)
                :SetInteger(1)
    dlg:CreateOkButton()
    dlg:RegisterInitWindow(function()
            dlg.OkButtonCanClose = true
        end)
end
dlg:RunModeless()
```

I realize @ThistleSifter can't test this, but I'm interested in ideas. I've tried to debug the Mac PDK Framework, but this particular code is nearly opaque to me.

rpatters1:
This PR fixes #588.

I am proposing this change to mixin control restore handling to place it before the window starts running. The PDK Framework has a bug on the macOS side that calling `FCControl::SetHeight` after the window is running causes `NSButton` controls to meander up the window each time the same instance of `FCCustomLuaWindow` is used open a window. Addressing the issue in `mixin` is the preferred solution because:

- It allows an immediate fix to all supported versions of Lua on Finale.
- This particular code in the PDK Framework is complicated and modifications are likely to introduce new issues. I don't think a fix in the Framework is worth the risk of breaking something else.

What say you, @ThistleSifter ?

### DIFF
--- a/src/mixin/FCMControl.lua
+++ b/src/mixin/FCMControl.lua
@@ -342,10 +342,10 @@ function methods:StoreState()
     private[self].Text = temp_str.LuaString
     private[self].Enable = self:GetEnable__()
     private[self].Visible = self:GetVisible__()
-    private[self].Left = self:GetLeft__()
-    private[self].Top = self:GetTop__()
     private[self].Height = self:GetHeight__()
     private[self].Width = self:GetWidth__()
+    private[self].Left = self:GetLeft__()
+    private[self].Top = self:GetTop__()
 end
 
 --[[
--- a/src/mixin/FCMControl.lua
+++ b/src/mixin/FCMControl.lua
@@ -362,10 +362,10 @@ Restores the control's stored state.
 function methods:RestoreState()
     self:SetEnable__(private[self].Enable)
     self:SetVisible__(private[self].Visible)
-    self:SetLeft__(private[self].Left)
-    self:SetTop__(private[self].Top)
     self:SetHeight__(private[self].Height)
     self:SetWidth__(private[self].Width)
+    self:SetLeft__(private[self].Left)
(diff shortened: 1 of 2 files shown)


ThistleSifter:
When I was testing the code when I first wrote it, if certain values were changed in between showings (as would now the case with the change here), they would be reverted when the window is next shown. The call to `RestoreState` is in `InitWindow` to ensure that any changes persist.

Since calling `SetTop` after `SetHeight` fixed the issue ( https://github.com/finale-lua/lua-scripts/issues/588#issuecomment-1735503966 ), what about moving lines 367-368 in `FCMControl.lua` 2 lines up instead?

rpatters1:
Good idea. What say you now?

ThistleSifter (approved):
Looks good.

PR #596, opened 2023-10-28
Create layer_actions.lua

cv-on-hub:
This script replaces four old (deprecated) ones in the repo: `layer_hide`, `layer_mute`, `stem_direction_by_layer` and `layer_clear_selective`


cv-on-hub:
P.S.: should the deprecated scripts be removed from the repo?

rpatters1:
If the new script fully replaces the functions in the old scripts, the only reason to keep them might be to support JW Lua. But if they require RGP Lua anyway, then for sure delete them.

PR #601, opened 2023-11-12
Create go_to_time.lua

cv-on-hub:
Navigate to a specific time in the current file using minutes (integers) and seconds (with decimals). To simplify distribution it doesn't use `mixin` dialog.


rpatters1:
Whether you use mixins or not, the `dist` version will be entirely self-contained. I'm not sure what your concern is.

cv-on-hub:
No concern. Just noting that it was created for a direct FB posting, so without `mixin`.

PR #613, opened 2023-12-08
Update deletion_chooser.lua

cv-on-hub:
Fixed ReverseStem flags for RGP0.68+ and important correction to delete_staff_styles. Sundry other small improvements, keystroke filtering and two key "commands".


rpatters1:
Instead of checking lua version, I would check if the property or property setter exists:

```Lua
if entry.ReverseUpStem ~= nil then
```

or

```Lua
if entry.SetReverseUpStem then
```

You only need to check one. If UpStem exists then obviously DownStem does as well.

rpatters1:
Or better yet, up the minimum version to 0.68 and dispense altogether with checking.

cv-on-hub:
> up the minimum version to 0.68 and dispense altogether with checking.

I love simpler solutions.

PR #614, opened 2023-12-10
Create rest_slider.lua

cv-on-hub:
This was designed specifically to help align rests midway between staves with cross-staff notes.


cv-on-hub:
Don't process this one just yet. There's a sneaky bug needs squashing.

PR #615, opened 2023-12-10
Create cue_notes_overlay.lua

cv-on-hub:
This was designed specifically to create cue notes above or below existing "played" material in the destination. It is now my preferred CUE NOTE creator.


rpatters1:
Instead of choosing the first empty layer, could you (at least optionally) allow a layer to be specified? I always use Layer 4 exclusively for cues. The layer is set as muted in Layer Prefs, and that takes care of playback.

cv-on-hub:
Notes: '_If the nominated "cue" layer already contains music the cue will copy to the first empty layer._'
This was just an expedient in case the user accidentally nominated a layer for cues that was already occupied.

rpatters1:
Oh, I see. Perhaps even this should be an option, though. If it were me, I would prefer an error message if I am trying to create a cue where there is already music in the designated layer.

PR #628, opened 2024-01-09
Create ostinato_maker.lua

cv-on-hub:
Copy the current selection (of any length) and paste it consecutively to the right a nominated number of times.

I’m not sure this warrants inclusion in the repo since much the same thing can be done with “Paste Multiple”, though this is simpler and more intuitive to use. It’s also a natural successor to and considerable improvement upon `region_replicate_music` which is already included.


rpatters1:
This script will be far more useful as a modeless dialog. It won't take much to change it into one, just adding Undo handling in the make_ostinato function. (And adding `finaleplugin.HandlesUndo = true`.) A nice to have might be to show the current selection, which means adding a timer to check when it changes.

Note that mixin `RunModeless` has the mod_key repeat (without opening dialog) functionality baked in, provided you activate `make_ostinato` from calling `RegisterHandleOkButtonPressed`. See `transpose_by_step.lua` for a relatively simple modeless dialog implementation.

PR #629, opened 2024-01-09
Create diatonic_doubler.lua

cv-on-hub:
Double notes and chords up or down by a chosen diatonic interval, with optional octaves.

Not sure if this deserves a spot in the repo. The action can be achieved with  any existing transposition options that include “Retain Original Pitch” … like `transpose_chromatic` for instance. But it’s also a natural progression from `pitch_entry_double_at_interval` simplifying the selection of diatonic intervals without having to manually alter script Prefix inputs.


cv-on-hub:
Revised version tests clean.

PR #635, opened 2024-01-11
Update transpose_by_step.lua

cv-on-hub:
So this feels insane. A correction and a suggestion.

1) I'm pretty sure that the CreateEdit position offset for Mac is vertical, not horizontal.

2) Amazed to discover that I couldn't find a single transposing script that filters by layer. So maybe this one?

### DIFF
--- a/src/transpose_by_step.lua
+++ b/src/transpose_by_step.lua
@@ -42,7 +42,7 @@ end
 local transposition = require("library.transposition")
 local mixin = require("library.mixin")
 
-function do_transpose_by_step(global_number_of_steps_edit)
+function do_transpose_by_step(global_number_of_steps_edit, global_layer_number)
     if finenv.Region():IsEmpty() then
         return
     end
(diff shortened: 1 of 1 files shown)


rpatters1:
All scripts filter by layer based on Show Active Layer Only. That is, the `eachentry` iterator serves up exactly those entries that are visible. I would rather not clutter up the transpose scripts with layer filters, because it just leads to errors and requires more keystrokes, and you can end up making changes that you can't see and thereby lose the chance to Undo them.

There intentionally is both a vertical and horizontal position offset for edit boxes. The horizontal offset is due to the blue rectangle that draws around the edit control when it has keyboard focus. It probably is a moot point in this dialog, but I see no reason to change it. "It it ain't broke,..." and all that.

cv-on-hub:
All understood. I'm one of those dinosaurs who keeps "Automatic Music Spacing" engaged so "Active Layer" creates havoc - the reason almost all my scripts filter by layer. TETO. (To each their own)!

PR #642, opened 2024-01-18
`FCMCtrlCheckBox` store check state

rpatters1:
`FCMCtrlCheckBox.lua` was not storing its `Check` state. This meant that the dialog boxes would not remember checkbox state unless the script explicitly called `SetCheck` before the dialog was created the first time.

This PR adds `StoreState` and `RestoreState` overrides to `FCMCtrlCheckBox.lua`. I have tested with both 2-state and 3-state checkboxes.

### DIFF
--- a/src/mixin/FCMCtrlCheckbox.lua
+++ b/src/mixin/FCMCtrlCheckbox.lua
@@ -11,9 +11,27 @@ local mixin_helper = require("library.mixin_helper")
 
 local class = {Methods = {}}
 local methods = class.Methods
+local private = setmetatable({}, {__mode = "k"})
 
 local trigger_check_change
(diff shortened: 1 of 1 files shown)


ThistleSifter:
You'll need to modify `SetCheck` so that it modifies the mixin storage when the window isn't showing. `GetCheck` will also need to be added so that it reads from the mixin storage when the window isn't showing. Check out https://github.com/finale-lua/lua-scripts/blob/7b79df9e6531a7debe49f858faa6f67bbee38b79/src/mixin/FCMControl.lua#L228 for an example of both.

The rest looks good, just from reading it through.

ThistleSifter (approved):
Yep, that looks good.

PR #646, opened 2024-01-20
lint fixes and build broken

rpatters1:
`page_singles_changer.lua` broke the build because of the raw call to `require` for the embedded `cjson` c-library. Recall that the build parses the source files for dependencies by looking for `require` statements. Therefore, in this repo we have to wrap those require statements for embedded utilities with a call to `util.require_embedded`.

While I was at it, I fixed a couple of lint errors and changed the code to use `library.calc_script_name` instead of hard-coding the script name.

Please test these changes. If you would rather hard-code the script name, that's fine. I just like to avoid hard-coding when possible.

### DIFF
--- a/docs/rgp-lua.md
+++ b/docs/rgp-lua.md
@@ -103,6 +103,13 @@ _RGP Lua_ (starting in version 0.67) pre-loads the lua-cjson 2.1.0 library, whic
 local cjson = require('cjson')
 ```
 
+If you are writing a script to be deployed at the [Finale Lua](https://finalelua.com) website, you must wrap the call to `require` as follows:
+
+```lua
+local utils = require('library.utils')
+local cjson = utils.require_embedded('cjson')
+```
+
 The json strings formatted by cjson are flat, containing no line feeds. If you wish to format them in human-readable format, you can use the built-in function [`prettyformatjson`](#prettyformatjson).
 
 More information on how to use the cjson library is available here:  
@@ -136,6 +143,13 @@ _RGP Lua_ (starting in version 0.68) pre-loads the luafilesystem library ('lfs')
 local lfs = require('lfs')
 ```
 
+If you are writing a script to be deployed at the [Finale Lua](https://finalelua.com) website, you must wrap the call to `require` as follows:
+
+```lua
+local utils = require('library.utils')
+local lfs = utils.require_embedded('lfs')
+```
+
 More information on how to use the lfs libray is available here:  
 [https://lunarmodules.github.io/luafilesystem/](https://lunarmodules.github.io/luafilesystem/)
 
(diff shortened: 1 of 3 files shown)


asherber:
@rpatters1 Maybe the docs should be updated regarding requiring embedded libs? On your site and the finalelua site it still says just to do `require('cjson')`. Maybe the explanation is that you can do this in your own personal scripts, but you should use `require_embedded` for any script submitted to the repo.

rpatters1:
As issue with the docs is that the RGP Lua docs are targeted at *any* RGP Lua user whereas the need to use `utils.require_embedded` is just an idiosyncrasy of this repo. I suppose this could be spelled out in the docs.

asherber:
Yes, that's what I was trying to get at in my comment.

cv-on-hub (changes requested):
That all looks terrific except for line 32 in `pitch_singles_changer.lua`.  "library" is misspelt as "libary". 

I had previously used `utils.require_embedded("cjson")` but changed to `require('cjson')` because of the documentation. Happy to revert. Thanks also for converting `globabl_info`  into a mixin dialog variable.

cv-on-hub:
PS: hadn't noticed `library.calc_script_name()` before. Will use it universally now instead of hard coding.

jwink75:
Can I get confirmation that going forward (and maybe even on older scripts) we are all supposed to use utils.require_embedded(“…..”) rather than require(“…..”)?I’ll also check out library.calc_script_name()  …Sent from my iPhoneOn Jan 20, 2024, at 1:15 PM, Carl Vine ***@***.***> wrote:﻿
@cv-on-hub requested changes on this pull request.

That all looks terrific except for line 32 in pitch_singles_changer.lua.  "library" is misspelt as "libary".
I had previously used utils.require_embedded("cjson") but changed to require('cjson') because of the documentation. Happy to revert. Thanks also for converting globabl_info  into a mixin dialog variable.

rpatters1:
You only use `utils.require_embedded` for C libraries embedded in RGP Lua. I believe there are three (possibly 4 if you count `socket`) that require this. (`cjson`, `lfs`, and `luaosutils`). I could conceivably embed more, but I've looked for widely used C libraries that have utility in Finale, and I am hard-pressed to find more. Maybe [LPeg](https://www.inf.puc-rio.br/~roberto/lpeg/)? (It's powerful but looks like it has a steep learning curve.)

All the regular lua libraries should continue to use `require`.

The reason we need `utils.require_embedded` is that the build parses the files for `require` statements and dumps the libraries into the `dist` version of the script, so that we can keep the distributed scripts free of dependencies.

jwink75:
OK, thanks, that makes sense.

From a usage standpoint, would I do something like:

local osutlities = utils.require_embedded("luaosutils”)

?

asherber:
@cv-on-hub I wanted to call out that the build process has been revised so that `require_embedded()` is no longer needed (in fact, is no longer available). Your current scripts have been adjusted to just use `require()`, and you should use this going forward. We've also updated the docs to reflect this.

cv-on-hub:
> call out that the build process has been revised

Thanks again @asherber, and for updating `hotkey_script_palettes.lua`. Another of my scripts uses `cjson` (`pitch_singles_changer.lua`) which I've updated on a new PR.

Issue #643, opened 2024-01-19
Default values for missing class methods


rpatters1:
**Is your feature request related to a problem? Please describe.**

I am working on localization of some scripts using new localization tools I am developing. Specifically, I would like to localize the transposition scripts as a pilot project for other scripts in the repo. But these scripts still work with JW Lua. I would like to find an elegant solution for skipping the new method calls in RGP Lua so that the scripts can continue to work with JW Lua.

**Describe the solution you'd like**

It occurred to me that `mixin` might be able to provide default implementations of missing methods. What I have in mind is that any method call could provide an optional additional parameter. If the additional parameter were present (non nil), the `mixin` library would check for the existence of the underlying RGP Lua method. If not found, it would return either return the default value passed in or (if the method is fluid) ignore the call entirely and simply return `self`. (If found it would call the method and behave as it does now.)

**Describe alternatives you've considered**

I suppose I could write one-off versions of the new methods that do the same thing. Or I could pepper the scripts with ugly if-statements.

@ThistleSifter

ThistleSifter:
I would suggest that a wrapping or proxy function

ThistleSifter:
As far as mixin solutions go, I think there are two reasonable options here.

1) Write the methods
Self explanatory...

2) Wrapping function & wrapping method
Add a function to the mixin namespace, like the example below. This could be used for anything (eg if a new function is added to RGPLua along the lines of `each`, etc)
```lua
local str = mixin.FCMString()
mixin.fallback_call("fallback_value", str.MethodName, self, arg_1, arg_2)
```
For DX/readability, a method can be added which proxies the function above. Eg
```lua
local str = mixin.FCMString()
str:__FALLBACK__("fallback_value", "MethodName", arg_1, arg_2)
```

If you want it to be fluid, pass `self` as the fallback value.

rpatters1:
After I posted this, it occurred to me that it would be possible to rewrite the methods in Lua. It's not like the OSs provide a lot of support for automatic control sizing, but (especially macOS) provides at least some.

I like the fallback call idea. I'm not crazy about the double-underscores (aesthetically) for the proxy. Do you think simply all caps would sufficient future-proofing?

rpatters1:
This was closed with #649.

PR #647, opened 2024-01-20
Implement fallback_call

rpatters1:
This PR implements a fallback call for `mixin` similar to the solution proposed in #643. I modified the proposed solution as follows:

- used `fallback_call` as the name of the proxy method as well as the mixin function. No PDK Framework class will ever use snake case for a property, method, or constant value, at least as long as I am on the project. That means any snake case method name is safe for use by mixin.
- changed the order of parameters for readability.

This solution has an added benefit that if a mixin version of a missing method is ever implemented, this function should pick it up.

An open question is

- Should `mixin.fallback_call` be private instead of public?


ThistleSifter:
It might be safe, but I really don't like the idea of mixing case styles in the *same* context, it's messy.

Perhaps it might be better to add a `__FCMBase` mixin and put it there instead?

rpatters1:
The `__FCMBase` idea is elegant. It isolates the change, leaving the mixin library itself unchanged. Excellent suggestion.

I also changed the name to use pascal case with a leading underscore. The PDK Library has some public class methods that start with underscore, but none of them are exposed to Lua. I don't plan to start, either.

src/mixin/__FCMBase.lua
```diff
@@ -0,0 +1,37 @@
...
+and get a default return value in that case.
+
+@ self (userdata) The class instance on which to call the method.
+@ method_name (string) The name of the method to return.
+@ fallback_value (any) The value that will be returned if the method does not exist. If this value is `nil`, the function returns `self`.
+@ additional_parameters (...) The additional parameters of the method.
+]]
+function methods:_FallbackCall(method_name, fallback_value, ...)
+    if not self[method_name] then
+        return fallback_value or self
```
ThistleSifter, line 31: This should explicitly check for `nil` so that it doesn't return `self` if `false` is passed, which could be a legitimate fallback value.

rpatters1:
Another excellent catch. I've fixed it, and I'm going to go ahead an merge it because I need to have it in place for the next PR I will be submitting soon. If there are further tweaks we can do it there.

PR #649, opened 2024-01-21
Localization

rpatters1:
This PR introduces a mechanism for localization of scripts. It consists of the following:

- `library.localization` provides the run-time localization service that can be used everywhere except the `plugindef` function.
- `library.localization_developer` provides an API for auto-translating strings using OpenAI APIs.
- `rgp-lua.md` describes a new feature where the user's OS-configured locale will now be passed to the `plugindef` function starting in RGP Lua 0.71. This allows `plugindef` to localize strings while remaining dependency-free. The change is fully backwards compatible with all previous versions of RGP Lua and with JW Lua, provided you check `locale` for nil.
- `auto_layout` is a demonstration script of auto-layout features coming in RGP Lua 0.71.
- `auto_layout_localizing_script` shows how to use `library.localization_developer` to translate strings.
- All of the transposition scripts have been localized in Spanish and German. While they require RGP Lua 0.71 to be fully functional, they still work in JW Lua and earlier versions of RGP Lua. One way this is accomplished is by using the new mixin `_FallbackCall` method to call any method that is new in 0.71.

In developing this approach, I have come to realize that LLM translation is amazing, but it is not reliable enough to put in front of a user without human inspection and revision. I speak enough Spanish and German to be comfortable that the translations for transposition are usable. I do not feel the same way about the French, Chinese, Arabic, and Farsi that I put into the `auto_layout` sample. (Although a Farsi-speaking friend of mine said they are pretty good.) I mainly included Farsi and Arabic in the sample to confirm that right-to-left text works reasonably well.

Unfortunately, `auto_layout` requires 0.71, but it is mainly a test of auto-layout features rather than localization per se.

I welcome any and all suggestions. I have specifically added @ThistleSifter and @asherber as reviewers, but I welcome comments from anyone. I haven't marked it as a draft, but if there is enough conversation about the best way to make it work, I will so mark it. Localization adds complexity but it makes for a much better UX. That said, I would like to make the DX as smooth as possible.


asherber:
That's a big PR! It'll take me some time to sift through it.

I agree with your take that machine translation needs human review -- especially when we're talking about isolated words and phrases that may mean something particular in a musical/Finale context that the AI is unaware of.

One question I have is about OS locale vs Finale language. According to [this page](https://www.finalemusic.com/products/international-dealers/), the only non-English languages that Finale comes in are French, German, Italian, Japanese, Dutch, Polish, and Swedish. I suspect that there are many users with a non-English locale who are running a version of Finale that doesn't match their locale. Will users want to see plugin dialogs in a language that doesn't match the rest of Finale? Is there any way to determine the language of the currently running Finale? (Probably not.)

rpatters1:
What a nice can of worms @asherber opened. I'll try to share what I know about it in a manner that is halfway coherent.

- The localizations for the six languages (other than English) mentioned on the website are each maintained by separate international affiliates who have partnership deals with MakeMusic.
- MakeMusic itself maintains the localizations for English and Spanish. (Spanish was introduced in 2020. I don't remember the exact version number, perhaps the first F27 version?)
- The six affiliate localizations appear originally to have created by directly modifying hard-coded strings in Finale source code. I was tangentially involved in the multiyear effort to modernize the UI code so that this was no longer necessary: a major development effort behind the scenes from ~Finale 2012 until the Spanish release in 2020.
- Each Finale version appears to contain only one localization, but all the plugins (including the MakeMusic-supplied plugins) contain all the localizations. You can explore them yourself. Use ResEdit on Windows or just open the plugin bundle on Mac.
- What you will discover is that the plugins are only partially translated, with some languages being more complete than others. The last time I looked, only Spanish was 100% translated, and I'm guessing it's because MM controls that version. That's one reason why I took the trouble to translate the Patterson Plugins to Spanish, and I'm planning to do the same for the RGP Lua dialogs (finally) this cycle.

As part of testing my plugins originally and now testing Lua, I took the following steps.

- Set up (in addition to US English) US Spanish, German, and Japanese on both macOS and Windows.
- Switching between them requires a reboot on macOS and a logout/login on Windows.
- Downloaded the Spanish Finale demos (both Mac and Win) from MakeMusic and the German Finale demos from Klemm.

As for how it all works, it matters which version of Finale you are running and which operating system.

- Windows appears to be the simplest. For the plugins you get the language you specified in the Settings app, irrespective of which Finale version you run. (I haven't tested pre-Win10, but this is how it works in WIn10+.)
- macOS is not so simple. The German version of Finale appears to override the OS locale setting and force the plugins to load the German localization. The Spanish version of Finale, on the other hand, defers to the OS locale setting. So I can run the Spanish version with my language set to English, and I'll get the English version of the plugins. The only way to see the Spanish version is to reboot into the Spanish version of the OS.

There are other differences. The Spanish version uses the built-in OS folder-name translation. That is, all the default folder paths are the same between English Finale and Spanish Finale. The merely appear with different translations in the Finder. The German version, otoh, has hard-coded German folder names. I attribute this to the German version predating modern localization techniques. It's possible that the other Finale languages use yet different approaches. I haven't tried to test them.

So now we get down to Aaron's key question: 

>Will users want to see plugin dialogs in a language that doesn't match the rest of Finale?

There probably is no single answer to what users *want*. But I can say with certainty that what they actually see now is a hodge-podge mix of languages. I have always taken it is as given that if you have set a particular locale as your primary default locale, you should present that locale to them if you have it. But ymmv.

rpatters1:
Something else I'd like us to consider is, should we build in a standard way to extend the number of localizations? For example, a user might prefer to see the dialogs in Vietnamese, but there is no Vietnamese built into the script. We could conceivably modify the `localize` function to attempt to load it. It seems like a lot of overhead for a rare situation, though, so there would have to be a way to try only once per execution of the plugin (Lua state).

ThistleSifter:
One thing that stands out to me is the use of the entire English phrase as the key for language strings in the examples, rather than a regular identifier. Although it might seem nice and readable, it's not good practice and it will introduce more difficulties in the long run, especially when the need arises to change the strings. Best to encourage good habits to start with.

Some changes to the mixin library could be made to make the code less verbose but I'll think on it some more before I make some suggestions.

ThistleSifter:
For dealing with localised strings:

`FCMControl.lua`
```lua
function methods:SetLocalizedText(key)
    mixin_helper.assert_argument_type(2, key, "string", "FCString")

    self:SetText(localization.localize(key))
end
```

`FCMUI.lua`
```lua
function methods:AlertLocalizedError(message_key, title_key)
    mixin_helper.assert_argument_type(2, message_key, "string", "FCString")
    mixin_helper.assert_argument_type(3, title_key, "string", "FCString")

    self:AlertError(localization.localize(message_key)), localization.localize(title_key))
end
```

etc, etc

rpatters1:
@ThistleSifter I thought about building this into `mixin` and I think it's a sound idea in principal. The main downside I see is that now every script that uses `mixin` will copy in the `localization` library into the `dist` version whether they use it or not. But maybe that's not a big deal given everything else that's copied in.

Taking that as given:

- Is there any reason to have separate `Localized` versions of the methods? (I'm not saying there isn't; I just want to think it through.) The current methods could instead localize automatically.
- Your reply made me realize I should be using [`CreateChildUI`](https://pdk.finalelua.com/class_____f_c_user_window.html#a6cefd9a5e0581d2501ce05a77289317e) when it is available, e.g.:

```lua
global_dialog:_FallbackCall("CreateChildUI", finenv.UI()):AlertError( --...
```

- I'm guessing that `CreateChildUI` called on an `FCXCustomLuaWindow` instance automatically returns `FCMUI`. But what about `finenv.UI()`?

ThistleSifter:
> The main downside I see is that now every script that uses `mixin` will copy in the `localization` library into the `dist` version whether they use it or not.

Yes,  the bundled scripts are big enough as it is. I guess it comes down to whether or not you want to enforce localisation as part of a coding style expectation in this repo. If all the scripts that use mixins also use the `localization` library, then there is no functional difference I suppose. But that's another can of worms.


> * Is there any reason to have separate `Localized` versions of the methods? (I'm not saying there isn't; I just want to think it through.) The current methods could instead localize automatically.

Yes, because then the `mixin` method would go beyond being an enhancement and break existing functionality. (Ie if I wanted to set plain text, for whatever reason, I can't do that any more except by calling `SetText__` with a `FCString` object). It would also break other `mixin` methods (eg the measurement setting methods) which use `SetText` internally.



> * I'm guessing that `CreateChildUI` called on an `FCXCustomLuaWindow` instance automatically returns `FCMUI`. But what about `finenv.UI()`?

`mixin.UI()` returns `FCMUI`.

rpatters1:
If anyone is tracking this branch, I'm going to rebase it on upstream/master when I get the chance.

rpatters1:
Okay, this version incorporates the suggestions from @ThistleSifter for `mixin` support of localization. (Please check it over for completeness/consistency/etc.)

I have moved the localization tables to a sub-directory called `localization`. These are automatically supported by the `localization` library. All we would need to do to add a language `lg` is add a `script_name_lg.lua` file to this directory. The base script would not need to be changed. (It will also support locales, e.g., `script_name_es_MX.lua` which could be a subset of `script_name_es.lua` for just those terms that are specific to Mexico.

What still remains is to modify the build script to embed these in the `dist` version of the scripts. But before I do that, I am looking for suggestions and comments. For example, to keep things tidy, should each script name have its own subdirectory inside the `localization` directory? Comments welcome.

ThistleSifter:
I think it's worth adding a function to `mixin_helper` to handle creating `*Localized` versions of methods automatically, to minimise vast amounts of copying and pasting for the rest of the text setting methods that aren't covered in this PR (eg `SetTitleLocalized` in `__FCMUserWindow`).

And for consistency, having the `*Localized` methods support `FCString` with a quick call to `mixin_helper.to_fcstring`.

For `FCMCtrlPopup`, I'd add an `AddStringLocalized` method and then have `AddStringsLocalized` call that.

Other than that, looks good.

One thing I was meaning to ask: is there a situation in which locale would change while a script is running or is it assumed to be fixed?

rpatters1:
You have my attention, @ThistleSifter. I like the way you are thinking. But I am not sure how to create the automatic `*Localized` function in `mixin_helper`.

I intentionally excluded `FCString` as inputs to the `*Localized` functions because the keys into the localization tables have to be Lua strings. But I am willing to be otherwise persuaded.

ThistleSifter:
Is there a way for me to add commits to this PR?

rpatters1:
You could add my personal fork as a remote on your clone of the repo. Then you could simply commit to the RGP-Localization branch. We'll need a little coordination if you are going to do it, especially if we need to do another rebase, but it works quite well.

BTW, to answer you question about changing locales, there is no end-user requirement I can think of to change locales in mid-stream. Windows requires logging out and macOS requires rebooting to change locales. For QA and testing, however, it is useful to be able to switch them quickly, but even for that I have not contemplated switching locales in a single Lua session.

ThistleSifter:
Hmm, maybe I'll just paste some snippets here. I'm still without a PC so I'm pretty much limited to what I can do through the web UI.

ThistleSifter:
This should hopefully do it.

Add to `src/library/utils.lua`:
```lua
--[[
% create_lookup_table

Creates a value lookup table from an existing table.

@ t (table)
: (table)
]]
function utils.create_lookup_table(t)
    local lookup = {}
    for _, v in pairs(t) do
        lookup[v] = true
    end
end
```

Add to `src/library/mixin_helper.lua`:
```lua
--[[
% create_localized_proxy

Creates a proxy method that takes localization keys instead of raw strings.

@ method_name (string)
@ class_name (string|nil) If `nil`, the resulting call will be on the `self` object. If a `string` is passed, it will be forwarded to a static call on that class in the `mixin` namespace.
@ only_localize_args (table|nil) If `nil`, all values passed to the method will be localized. If only certain arguments need localizing, pass a `table` of argument `number`s (note that `self` is argument #1).
: (function)
]]
function mixin_helper.create_localized_proxy(method_name, class_name, only_localize_args)
    local args_to_localize
    if only_localize_args == nil then
        args_to_localize = setmetatable({}, {__index = function() return true end})
    else
        args_to_localize = utils.create_lookup_table(only_localize_args)
    end

    return function(self, ...)
        local args = table.pack(...)

        for arg_num = i, args.n do
            if args_to_localize[i] then
                mixin_helper.assert_argument_type(arg_num, args[arg_num], "string", "FCString")
                args[arg_num] = localization.localize(mixin_helper.to_string(args[arg_num]))
            end
        end

        --Tail call. Errors will pass through to the correct level
        return (class_name and mixin[class_name] or self)[method_name](self, table.unpack(args, 1, args.n))
    end
end
```

Usage:
```lua
-- Will call mixin.FCMControl.SetText
methods.SetTextLocalized = mixin_helper.create_localized_proxy("SetText", "FCMControl")

-- Will call self:AlertError
-- All arguments automatically localized
methods.AlertErrorLocalized = mixin_helper.create_localized_proxy("AlertError")

-- To only localise certain arguments (eg obj:MyMethodLocalized("foo_key", 12, 6555, "bar_key")
methods.MyMethodLocalized = mixin_helper.create_localized_proxy("MyMethod", nil, [2, 5])
```

Let me know if something doesn't work.

rpatters1:
Would there be a way to automatically create `*Localized` methods in `__newindex`? Is this an idea worth pursuing if so?

ThistleSifter:
I don't think so, because either way you'd still need some way of registering which methods to create localised versions of, where the call needs to be forwarded to, and which arguments to localise.

ThistleSifter:
How would you prevent a `SetIntegerLocalized` method being created, for example?

rpatters1:
You wouldn't, but it would not do anything other than call `SetInteger`, since none of its arguments are strings.

rpatters1:
Or if it matters, we could throw an error out of the `mixin_helper` function if none of the arguments are localizable.

rpatters1:
I believe I have incorporated all of @ThistleSifter's suggestions. Have a look.

In addition, I implemented the following:

- Allow Lua tables (of strings, FCString, and/or numbers) to be sent to `AddStrings` and `AddStringsLocalized` on both `FCCtrlPopup` and `FCCtrlComboBox`.
- added a utility function to construct an array table from the keys of an input table.

rpatters1:
@asherber I have reached the stage where I have to modify the bundler to inject all the localizations into the `dist` version of the file. It looks like I will need to modify `bundle.ts`. The requirements are

- Search the `src/localization` directory for a directory that matches our file name minus the extension.
- bundle any `.lua` files found there so that this will work in the main script:

```lua
require ('localization.<script-name>.<found-file-minus-extension>')
```
Example (for `transpose_chromatic.lua`):

```lua
require ('localization.transpose_chromatic.es')
```

I have made changes to the bundler before, but it was risky, lol. I have zero experience with `.ts` and almost none with `.js`. I tested it by repeatedly committing to `master` until it worked. I'm hoping there's a better way to test.

Any pointers would be much appreciated.

asherber:
Would you like me to take a stab at it? Or were you looking forward to the challenge?

rpatters1:
I would be delighted if you take a stab at it. I have many challenges and surmount them only if necessary. (lol)

asherber:
Sure thing! 

So, to flesh out your example:

- There would be a folder `src/localization/transpose_chromatic`
- Inside this folder is a file `es.lua`
- We need to inject the line `require ('localization.transpose_chromatic.es')` into the source script, probably immediately after `plugindef()`
- The bundler needs to read this line and bundle in the file contents, the way it currently does with something like `require('library.utils')

Is that a good statement of the task?

rpatters1:
The localization library does the require statements, so there is no need to inject the require statements themselves. We just need to bundle in the localization scripts as preloads so that the require statements will work. I would say it is more similar to how it bundles in all the files in the `mixin` directory, even though the scripts do not contain explicit require statements for them.

rpatters1:
To continue: they should be bundled if they are there, even though they are *not* explicitly required in the scripts.

asherber:
Ah, okay, makes sense. Let me see what I can do.

asherber:
@rpatters1 Do these files look correct?

[dist.zip](https://github.com/finale-lua/lua-scripts/files/14181996/dist.zip)

rpatters1:
That looks great!

asherber:
Sent a pull request to your fork

rpatters1:
Okay, I believe I have everything ready. I just need comments from @ThistleSifter on the `mixin` changes (and any other comments). Then I will merge it.

src/library/mixin_helper.lua
```diff
@@ -578,4 +597,64 @@ function mixin_helper.boolean_to_error(object, method, ...)
...
+    for i = 1, select("#", ...) do
+        local v = select(i, ...)
+        mixin_helper.assert_argument_type(i + 1, v, "table", "string", "number", "FCString", "FCStrings")
+
+        if type(v) == "userdata" and v:ClassName() == "FCStrings" then
+            for str in each(v) do
+                method_func(self, str)
+            end
+        elseif type(v) == "table" then
+            mixin_helper.process_string_arguments(self, method_func, table.unpack(v))
```
ThistleSifter, line 652: I had intentionally omitted the ability to pass tables as I explained here: https://github.com/finale-lua/lua-scripts/pull/183#issuecomment-1134205699

src/library/mixin_helper.lua
```diff
@@ -561,6 +562,24 @@ function mixin_helper.to_fcstring(value, fcstr)
...
+@ [fcstr] (FCString) An optional `FCString` object to populate to skip creating a new object.
+: (FCString)
+]]
+
+function mixin_helper.to_string(value)
+    if mixin_helper.is_instance_of(value, "FCString") then
+        return value.LuaString
+    end
+
+    return tostring(value)
```
ThistleSifter, line 580: Huh, I thought I wrote this function once upon a time. I must not have submitted it. Anyway, a `nil` value should be cast to an empty string, so this line would need to be changed to:
```lua
return value == nil and "" or tostring(value)
```

src/library/mixin_helper.lua
```diff
@@ -561,6 +562,24 @@ function mixin_helper.to_fcstring(value, fcstr)
     return fcstr
 end
 
+--[[
+% to_string
+
+Casts a value to a Lua string. If the value is an `FCString`, it returns `LuaString`, otherwise it calls `tostring`.
+
+@ value (any)
+@ [fcstr] (FCString) An optional `FCString` object to populate to skip creating a new object.
```
ThistleSifter, line 571: This line in the docblock isn't relevant to this function.

src/mixin/FCMCtrlComboBox.lua
```diff
@@ -0,0 +1,90 @@
...
+
+@ self (FCMCtrlPopup)
+@ ... (table, FCStrings | FCString | string | number)
+]]
+function methods:AddStrings(...)
+    mixin_helper.process_string_arguments(self, mixin.FCMCtrlComboBox.AddString, ...)
+end
+
+--[[
+% AddStrings
```
ThistleSifter, line 77: `AddStringsLocalized`

src/library/mixin_helper.lua
```diff
@@ -561,6 +562,24 @@ function mixin_helper.to_fcstring(value, fcstr)
     return fcstr
```
ThistleSifter, line 562: In keeping with the comment below on `to_string`, the line above this one (I can't target it with a review comment) should read:
```lua
fcstr.LuaString = value == nil and "" or tostring(value)
```

samples/auto_layout.lua
```diff
@@ -0,0 +1,349 @@
...
+    ["Left Checkbox Option 2"] = "Left Checkbox Option 2",
+    ["Menu"] = "Menu",
+    ["Right Three-State Option"] = "Right Three-State Option",
+    ["Second Option"] = "Second Option",
+    ["Short "] = "Short ",
+    ["Test Autolayout With Localisation"] = "Test Autolayout With Localisation",
+    ["Third Option"] = "Third Option",
+    ["This is long menu text "] = "This is long menu text ",
+    ["This is long text choice "] = "This is long text choice ",
+    ["This is longer option text "] = "This is longer option text ",
```
ThistleSifter, line 38: I still submit that encouraging authors to use full language strings as keys is inviting a world of pain...

ThistleSifter (changes requested):

samples/auto_layout.lua
rpatters1: >is inviting a world of pain...

Perhaps. Full language strings are encouraged (not required) in macOS localization. The advantages are

- the code is more readable
- the code can be written and tested before any considerations of localization, beyond using the `*Localized` methods.
- the text injected into the control is guaranteed to be human readable in some language.
- the solution is agnostic as to which language the programmer chooses for the base language.

That said, I have encouraged using non-human-readable keys in the `plugindef`. I would (in particular) hope that if we ever have to localize `finaleplugin.Notes` it will be with some key other than the full text. There is nothing requiring full language keys, but then there has to be a fallback language. I would rather leave that bridge uncrossed at the moment.

samples/auto_layout.lua
```diff
@@ -0,0 +1,349 @@
...
+
+    local line_no = 0
+    local y_increment = 22
+    local label_edit_separ = 3
+    local center_padding = 20
+
+    -- left side
+    dlg:CreateStatic(0, line_no * y_increment, "option1-label")
+        :DoAutoResizeWidth()
+        :SetWidth(0)
```
asherber, line 202: I gather that these methods are coming later, but why can't `DoAutoResizeWidth()` implicitly call `SetWidth(0)`, so that the two don't have to always go together?

samples/auto_layout.lua
asherber: Why do these strings include the trailing space?

rpatters1:
Okay, I think this addresses all the issues. I went ahead and implemented all the `AddString...` methods for `FCCtrlListBox` as well.

samples/auto_layout.lua
rpatters1: In real windows we probably don't want to call `SetWidth(0)`. This is more for demonstration. One of the features of `DoAutoResizeWidth` is that it only widens, never narrows. There are many situations where you will want a minimum width, or will want to take the default the control offers as a minimum. (Command buttons are a good example.)

rpatters1:
>Why do these strings include the trailing space?

They get numbers appended to them. Again, it's just a demonstration. One thing about the demo was to test r-to-l languages, even though Finale does not support any. I was interested to see that (at least in some cases) the space is automatically moved around to the left, with the appended numeral to the left of it. Editors like VS Code behave really weirdly with r-to-l strings.  Try using the left/right arrow keys to step through them in the `auto_layout` script.

samples/auto_layout.lua
ThistleSifter: > Full language strings are encouraged (not required) in macOS localization

I find that very surprising. If you look at various open source code bases (or even docs of some proprietary software), you'll be hard pressed to find i18n approached like that. For example: https://github.com/lokalise/i18n-ally/tree/main/locales which also utilises a kind of namespacing for grouping its strings. Or you can try any forum software, eg https://xenforo.com/docs/xf2/phrases/ or https://github.com/SimpleMachines/SMF/tree/release-3.0/Languages I'd encourage you to check out others too.

&nbsp;
> * the code is more readable

Not necessarily, as it can also make it more verbose. If I'm flicking through something, I'd much rather see this
```lua
finenv.UI():AlertError(loc.localize("error_msg_transposition"), loc.localize("error_title"))
```
than this
```lua
finenv.UI():AlertError(
            loc.localize(
                "Finale is unable to represent some of the transposed pitches. These pitches were left unchanged."),
            loc.localize("Transposition Error")
        )
```
I also find it gets in the way of separation of concerns. But more in the next dot point.

&nbsp;
> * the code can be written and tested before any considerations of localization, beyond using the `*Localized` methods.

Generally, a piece of code is either written with localisation in mind or it isn't. The current approach attempts to straddle the line so you get code that doesn't have to be localised, but it could be and I don't think it works. It relies too much on the magic of fallbacks rather than having well written and well structured code from the start. In the process you also get language string keys that don't look like keys, and for which you don't get the benefit of autocomplete, and which are also harder to edit.

For example, let's say I want to change the wording of an error message. If the key is `user_error` then I just have to change the value of `loc.user_error` (and generate a new set of translations if needed). But if the key is `"Fatal user error, please step away from the keyboard as it will self destruct in 3 ... 2 ... 1 ..."` then I need to change the key as well, or else every time I see the message in the code, I'm going to think "hold on, I changed that". So now instead of just editing the string, I have to highlight the text and use find and replace (and hope no errors were introduced). If for whatever reason it was a minor edit (like capitalisation or punctuation) and regenerating the translations wasn't necessary, I'd have to do that anyway or at least remember to replace the keys in _every_ language file.

There's also the matter using the keys, especially if you need them in multiple places. If I'm typing and I want to use a language string, I have to go back and copy and paste the whole thing. Typing from memory isn't even reliable since all keys in Lua are case-sensitive. That's not easier than typing `user_error` or even just getting part way through and hitting enter is easier.

Code written with i18n in mind will always look different to code that isn't: the content of the language string doesn't matter any more, what matters is what it is, which can also include when/where it is used (eg `menu_item_1`, `menu_item_2`, `exit`, `unknown_error`, etc). And that's how keys are named, according to what it is, which isn't necessarily what it says. When you're reading the code, you can focus on the code itself. When you're using the strings, the keys are easy to type and remember (or autocomplete). Then if you want to edit the copy, just look at the language file and focus on that.

&nbsp;
> * the text injected into the control is guaranteed to be human readable in some language.
> 
> * the solution is agnostic as to which language the programmer chooses for the base language.

Both of these could be better handled by having a default language (eg `en`) for which a language file is required to exist (if it doesn't, throw an error). If the author wants to change this, they could do so by calling a function (eg `localization.set_default_locale('fr')` ).

&nbsp;
> I would rather leave that bridge uncrossed at the moment.

I18n is something that's much easier to handle if you spend a little time to plan it out and get it right from the start,  before it becomes unwieldy with vast amounts of language files and keys/strings.

ThistleSifter:
> They get numbers appended to them.

Some sort of token handling is a must for situations like these. Take a look at ICU MessageFormat, for example.

rpatters1:
I slept on it, and I agree with both of you:

- `DoAutoResizeWidth` will accept an optional minimum width.
- The `localization` library will look for a `Base.lua` fallback table. This is the way Apple handles it, and I think it works for us. It also remains agnostic as to which is the fallback language.
- I'll think about token support. It feels like scope creep, but there might be a simple way to do it.

rpatters1:
Apple added `Base` localization relatively recently. Before that, when localization was getting off the ground, they encouraged (or a least suggested) full-text keys. Of course, since Windows still uses numerical identifiers as keys, any cross platform project did not do this. That includes Finale and my plugin suite.

You can still see it (I believe, sort of) with the strings files that are created for Interface Builder nibs. But I don't think those make it into the final binaries that ship.

rpatters1:
Okay, this is ready for more review and suggestions. Per comments from @asherber and @ThistleSifter I have made the following revisions:

- `Base.lua` is now recommended as a fallback localization. This allows constant keys no matter what we want the messages to say. I believe that all comments, plugindef functions, samples, and utilities have been updated to recommend and use `Base` localization. If you find one I missed, let me know.
- The new `DoAutoSizeWidth` method in RGP Lua 0.71 will accept an optional minimum size argument. All scripts, samples, and utilities have been updated to use this feature as appropriate. (I am still using `SetWidth` on the transposition scripts because I want to leave them JW Lua compatible.)
- The strings with trailing spaces in `auto_layout.lua` have been modified to have format parameters. I remembered that the way I have always handled localized strings with format parameters is that the caller constructs the final string from the localized string resource using (in the case of `c` or `c++`) `sprintf` or one of its equivalents. For Lua I am using `string.format`, which is the Lua version of `sprintf`. If one of you has a better idea of how to do this, I am interested. But I don't want to muck up the `mixin` API too much with a feature that is not that often used.

Once I have signoff, I will merge it. I'm thinking I'll use squash and merge, because the long commit history is rather annoying.

ThistleSifter:
> * `Base.lua` is now recommended as a fallback localization.

I'm not sure that this is a suitable option for our purposes. Take `/src/localization/transpose_by_step/Base.lua` for example. What locale is this? Sure I can take a guess, but guessing shouldn't even be an option. Someone should be able to look at the names of the files in the directory and know exactly which languages and territories are provided.

Do we actually need a general fallback other than `en` or if that doesn't exist, then the first one alphabetically? Are we trying to address an issue (that of incomplete or partial translations) that should be addressed somewhere else? I'm not talking about fallbacks within the same language (eg `en` as fallback for `en_GB` or `en_AU`), that's a different scenario. 

&nbsp;
> * If one of you has a better idea of how to do this, I am interested.

What about changing the signature of `localize` to `function localization.localize(input_string, ...)` where `...` is a series of parameters to insert? Mixin `*Localized` methods could accept `table`s if substitutions are required. For example:
```lua
finenv.UI():AlertErrorLocalized(["input_error", "Some", "inserted text"], "error_title")
```

rpatters1:
Passing a table to `*Localized` seems like it falls right back into the same problem that passing a table to `AddStrings` does: you can't properly enforce the types. If `string.format` is going to be called, shouldn't the person using it be calling it?

If `Base` localization is good enough for Apple, it's good enough for me. There has to be a fallback, and forcing English to be the fallback is not appropriate in the current age.

asherber:
I don't have a strong feeling about fallback languages, and I find the discussion very interesting. I note that Finale itself is developed in English and then localized, and that keywords in Lua (along with almost every programming language) are in English; it doesn't seem unreasonable in the context of software for English to be the general fallback language, which I think is part of what @ThistleSifter is suggesting.

What if name of the fallback file specified its locale (e.g., Base_en.lua)?

Feel free to ignore if this isn't helpful; I admit that I haven't dealt with localization issues before.

rpatters1:
The language a language is developed in is irrelevant to how it should be localized. Finale is released with only a single localization and that localization is the language it's in. I have the Spanish and German versions on my computer, for example, and both of those (like the English version) have the one localization in `Base.lproj`.

But if @ThistleSifter is comfortable with `Base_en.lua`, that is an easy change to make.

asherber:
Re `Base_en.lua` what I was thinking is that instead of `Base.lua` being a recommended fallback (per @rpatters1's earlier post), the recommendation would be `Base_<locale>.lua`. The code to find the fallback file would pick the first one that matches this pattern (which I think was part of @ThistleSifter's suggestions).

rpatters1:
I suppose there may be a way to pattern match in `require` by mucking with the the path, but I'm not sure it could be made to work with the preloads in `dist`. Anyway, here's a simpler way:

- get rid of the `Base.lua` concept (and rename the current `Base.lua` files to `en.lua`.
- add an api to the `localization` library `set`/`get`/`fallback_locale`.
- this value will default to `en`.

That allows a script writer to specify a different fallback language if need be.

asherber:
Ah yes, I missed your reference to `require`. What you propose sounds reasonable to me, but I'll butt out because it's not really my area.

ThistleSifter:
> Lol, fwiw, Apple is definitely still recommending full text as keys.

Out in the open source world, where people have to work together, snake case (sometimes other, but snake is the most popular esp. in the web languages) keys have been used for referencing translatable strings for well over 15 years. Organisations that can dictate terms because they have a functional monopoly can have some interesting policies, to put it gently. Not always bad, but it's often worth looking a little wider to see what everyone else is doing.

&nbsp;
>     * add an api to the `localization` library `set`/`get`/`fallback_locale`.

Yes, I like this idea.

ThistleSifter:
> Passing a table to `*Localized` seems like it falls right back into the same problem that passing a table to `AddStrings` does: you can't properly enforce the types

Yes, good catch. I have been thinking about this for the past couple of days because it would be nice to have and I think I have a solution for the error reporting but I have to work out how to do it. It might need to wait for the mixin rewrite because that has completely rewritten error handling with greater ease and flexibility for hooking in to it.

ThistleSifter:
Hold on, I've got it. I'll write it up and paste it here tomorrow.

rpatters1:
I'll hold off checking in until seeing your brainstorm.

How do you feel about adding `CreateOkButtonLocalized`, `CreateCancelButtonLocalized`, and `CreateCloseButtonLocalized`. They could either take a string argument as a key (which would break conformity with the other "Create..." apis) or the keys could just be hardcoded to "ok", "cancel", and "close". It's annoying having constantly to add a SetTextLocalized to these methods.

ThistleSifter:
I ended up editing in my brainstorm to the comment above last night. Here's a quick demo anyway:
```lua

local t_num = {"this", "is", false, "a", "test"}
local t_str = {}
t_str.a = "this"
t_str.b = "is"
t_str.c = false
t_str.d = "a"
t_str.e = "test"

...

popup:AddStrings(t_num) -- bad argument #1[3] to 'AddStrings' ... etc

popup:AddStrings("foo", t_str) -- bad argument #2["c"] to 'AddStrings' ... etc
```

&nbsp;
> How do you feel about adding to `FCMCustomWindow` `CreateOkButtonLocalized`, `CreateCancelButtonLocalized`, and `CreateCloseButtonLocalized`.

What about `CreateOkButtonAutoLocalized` and not having a key argument? Could also bundle in some default translations in another library file so that every script author doesn't have to retranslate "OK" unless they wish to.

rpatters1:
I am really puzzled by this statement as to how it is supposed to work:

```
    mixin_helper.assert_argument_type(tostring(i + 1) .. to_key_string(k2), v2, "string", "number", "FCString")
```

I just tested it, and it passes a string to the first argument `mixin_helper.assert_argument_type` that that is expecting a number. And I don't understand it: `2[1]` (for the first item in my array table). This appears to be wrong (though it produces seemingly correct results), but there could be wrinkle of Lua I don't understand here.

ThistleSifter:
Technically it doesn't matter what you pass for the value of `argument_number`, just that it probably should be a number or something that starts with a number, since it should fit in the same form as the standard error message. It's probably worth updating the docs to reflect that, maybe `number | string`, but only use `string` if you have a good reason to. Something like that.

rpatters1:
Okay, here it goes. I'm committing it.

PR #650, opened 2024-01-22
mobdebug breaks build!!

rpatters1:
A reminder to @jwink75 that `require('mobdebug').start()` breaks the build, *even if it is commented out!*.

A question for @asherber: would it be possible to modify the build script to ignore comments. At least single-line comments that start with `--`?

### DIFF
--- a/src/page_format_wizard.lua
+++ b/src/page_format_wizard.lua
@@ -1858,7 +1858,6 @@ local function format_wizard()
   bold_control(score_ctrls.staff_settings)
   bold_control(parts_ctrls.staff_settings)
   bold_control(special_ctrls.staff_settings)
---    require('mobdebug').start()
 --  finale.FCFileSaveAsDialog(finenv.UI()):GetFileName(str)
 end
 


jwink75:
Ach! Sorry!Sent from my iPhoneOn Jan 21, 2024, at 5:48 PM, rpatters1 ***@***.***> wrote:﻿A reminder to @jwink75 that require('mobdebug').start() breaks the build, even if it is commented out!.
A question for @asherber: would it be possible to modify the build script to ignore comments. At least single-line comments that start with --?



You can view, comment on, or merge this pull request online at:
  https://github.com/finale-lua/lua-scripts/pull/650

Commit Summary

  81fdea8  mobdebug breaks build!!


File Changes  (1 file)

  
    M
    src/page_format_wizard.lua
    (1)
  


Patch Links:

  https://github.com/finale-lua/lua-scripts/pull/650.patch
  https://github.com/finale-lua/lua-scripts/pull/650.diff

asherber:
Yes, I can modify the bundler so that it ignores single-line comments.

Also, looking at the bundler code, I am reminded that we added a list of imports to ignore -- currently `luaosutils` and `mime.core`. This is probably how we should handle the other embedded libs like `cjson`, rather than requiring devs to use the `require_embedded` workaround.

I could add `mobdebug` to this list as well, but (a) ignoring comments will take care of this, and (b) I'm a firm believer that the only commented lines in code should be comments -- any debug code should be removed before submitting a PR.

PR #651, opened 2024-01-22
Bundler improvements

asherber:
1. Add `cjson` and `lfs` to bundler exclusions
2. Don't try to import from single-line comments
3. Run the build action (but not the deploy action) for pull requests

### DIFF
--- a/.github/actions/bundle/dist/index.js
+++ b/.github/actions/bundle/dist/index.js
@@ -6810,7 +6810,15 @@ const getImport = (line) => {
     };
 };
 exports.getImport = getImport;
-const ignoreValues = ["luaosutils", "mime.core", "cjson", "lfs"];
+const ignoreValues = [
+    "cjson",
+    "cjson.safe",
+    "lfs",
+    "luaosutils",
+    "mime.core",
+    "socket.core",
+    "socket",
+];
 const commentRegex = /^\s*--/;
 const getAllImports = (file) => {
     const imports = new Set();
--- a/.github/actions/bundle/dist/index.js
+++ b/.github/actions/bundle/dist/index.js
@@ -1,7 +1,7 @@
 /******/ (() => { // webpackBootstrap
 /******/ 	var __webpack_modules__ = ({
 
-/***/ 7936:
+/***/ 1346:
 /***/ (function(__unused_webpack_module, exports, __nccwpck_require__) {
 
 "use strict";
@@ -28,7 +28,7 @@ var __importStar = (this && this.__importStar) || function (mod) {
 Object.defineProperty(exports, "__esModule", ({ value: true }));
 exports.issue = exports.issueCommand = void 0;
 const os = __importStar(__nccwpck_require__(2037));
-const utils_1 = __nccwpck_require__(5781);
+const utils_1 = __nccwpck_require__(9754);
 /**
  * Commands
  *
@@ -100,7 +100,7 @@ function escapeProperty(s) {
 
 /***/ }),
 
-/***/ 3539:
+/***/ 5681:
 /***/ (function(__unused_webpack_module, exports, __nccwpck_require__) {
 
 "use strict";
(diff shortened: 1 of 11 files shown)


asherber:
@rpatters1 Don't merge this yet; I want to run a couple of tests in the morning.

asherber:
Okay, in my test branch I [added some things](https://github.com/finale-lua/lua-scripts/pull/652/files#diff-12e3f3f567de0acd6626185bbfe20371e7acb7225bb1b1e1abbe27bbf00c7c38) that we don't want to break the build -- and the build [passed](https://github.com/finale-lua/lua-scripts/actions/runs/7606430393/job/20712209937?pr=652)

@rpatters1 Are there any other embedded libs besides `"luaosutils", "mime.core", "cjson", "lfs"`?

If not, then this is **ready to merge.**

rpatters1:
This looks great. I also appreciate the addition of the build action on the PRs. That way I (and everyone else) don't have to remember to check for these things. BTW I 100% agree that (in the era of ubiquitous git) comments in production code should be comments.

For completeness we should add `socket` and `socket.core` to the list of bundler exclusions. Normally scripts don't have to require these, but they might be requiring in library scripts or other found code that does it anyway.

Also, can we include in this PR the removal of `utils.require_embedded` from the repo, with changes as needed?

rpatters1:
Another that would be really nice to add to the build action, if it is possible, is a lint check. I would very much like to prevent PRs that contain lint errors. But it would have to have all the approved exceptions that exist in our `recommended_settings.json` file.

asherber:
@rpatters1 In openai.lua, you have `utils.require_embedded("cjson.safe")`. What is that, and should it be added to the list of exclusions?

jwink75:
I have a few questions…1. What will the new method for calling things like luaosutils be under this new setup?2. Does this mean the repo will actually report those lint errors when I try to submit?3. Do you know if it’s possible to configure Zerobrane to show me the lint errors you have Visual Studio set up for?Sent from my iPhoneOn Jan 22, 2024, at 6:14 AM, Aaron Sherber ***@***.***> wrote:﻿
@rpatters1 In openai.lua, you have utils.require_embedded("cjson.safe"). What is that, and should it be added to the list of exclusions?

rpatters1:
Yes, we should add `cjson.safe` to the exclusions. According the [docs](https://kyne.com.au/~mark/software/lua-cjson-manual.html), the "safe" version does not throw parse errors. It instead returns an error message in a second parameter.

rpatters1:
In answer to @jwink75's questions.

1. We will be reverting to using plain old `require` for everything. I will update `rgp_lua.md` if @asherber  doesn't.
2. Ideally the submission process would report lint errors.
3. There might be a lint feature in ZeroBrane. It would be worth investigating.

asherber:
Added `cjson.safe` and removed `require_embedded`. I'll try to work on linting later today.

asherber:
@rpatters1 I think the linting might be tricky and would rather save it for later and merge this stuff. The VSCode extension that handles linting can in theory be run as a standalone app, but in some informal testing I haven't been able to get it to work as advertised. 

There are some prebuilt GH actions that use Luacheck, but I'd need to play around with these to see if I can get things configured correctly to give comparable results to VSCode (or maybe use a Luacheck-based linter in VSCode as well). I think that may all take a little while.

rpatters1:
I'm perfectly fine with postponing lint checking. It looks like you modified the wrong version number in `hotkey_script_palettes.lua`. There is no Lua plugin version `0.70.1`.

If you want me to merge, let me know. Otherwise, with that one little fix, go for it. Thank you again.

asherber:
Ah, thanks for catching that version number! I'll fix it and merge.

asherber:
Hmm, it seems that the logic I added to the workflow yml correctly prevents publishing the site when a PR is opened -- but does _not_ correctly publish the site when a PR is merged! I should be able to fix that with a little testing this afternoon.

asherber:
Fixed via #653

jwink75:
I know Zerobrane does some analysis… I’ll try to do some research into how to configure it.

PR #655, opened 2024-01-23
Linting

asherber:
This adds linting to the build process, using luacheck. 

1. Luacheck is added to the list of recommended extensions for VSCode, so that devs will see locally the same warnings and errors that will be found on GH.
2. The GH action will run on any push, and on PRs to master; any failures will show up in the PR.
3. The `.luacheckrc` in the repo (which is used locally as well as on GH) is configured to ignore formatting issues, as well as a few other things that Robert and I felt were not important.
4. Only files in the `src` directory are linted, and third-party files in there (like `lunajson` and `mobdebug`) are ignored.
5. A dev may decide that a certain warning is acceptable and may be bypassed. For example:
    ```lua

    function do_something(region)
    ...
    end
    ```
    This will generate a warning for the function definition that the `region` parameter shadows the earlier variable with the same name. To suppress the warning, add a comment `-- luacheck: ignore region` on the same line as the function definition. 

The docs for luacheck are [here](https://luacheck.readthedocs.io/en/stable/index.html). In particular, here are the [potential warnings](https://luacheck.readthedocs.io/en/stable/warnings.html) that can be emitted (note that our `.luacheckrc` disables some of these) and instructions for [using comments](https://luacheck.readthedocs.io/en/stable/inline.html) to disable specific warnings. Devs are expected to disable warnings as locally as possible in their code.

There were about 70 warnings already in the codebase. I went through and fixed the ones that were easy to fix, and I added ignore hints to the others. **Please take a look at your scripts** and let me know if any changes are needed.

Suggestions and comments are welcome.


asherber:
@cv-on-hub I don't think this actually affects any of your current scripts, but we wanted to get your eyes on the PR as well.

cv-on-hub:
> we wanted to get your eyes on the PR as well

Thanks @asherber. The mechanics of how this all fits together is beyond my understanding but it looks like enhanced strength for the repo.

cv-on-hub (approved):

asherber:
As I'm looking at some of the potential warnings, I think I'm going to re-enable some of the warnings around unused variables -- these just create clutter and make code harder to read. This is going to make some extra work for me to clean up current instances, so I'm going to set this PR back to draft and will advise when I've pushed the new commit. Feel free to continue the discussion.

asherber:
I take it back. This is a good first step, and we should get it merged. Further cleanup can come later.

jwink75:
For anybody interested: Zerobrane Studio does indeed support Luacheck, it just requires setting a parameter in the configuration file.

jwink75:
I don’t know if anyone else is using Zerobrane, but I did figure out how to configure it to use Luacheck and use the luacheckrc parameters from the ‘linting’ pull request

asherber:
Okay, I have implemented stricter linting, primarily looking for unused variables. In addition, new global variables may only be initialized in the top-level scope. I have adjusted all scripts so that the linting passes.

Please review all of your scripts one more time to make sure that nothing is broken. I'd like to merge this in the next few days.

cc: @cv-on-hub @jwink75 @Nick-Mazuk @rpatters1 @CJGarciaMusic

cv-on-hub:
> new global variables may only be initialized in the top-level scope.

I have only one script using global vars -  `ostinato_maker.lua`. I've revised it leaving just one global, `global_dialog`, which I think is needed for modeless operation. Moved its initialisation to top-level, mimicking Robert's sample script `modeless_dialog.lua`. Submitted in #661 - deploy check successful.

jwink75:
The new rules don’t like me.

jwink75:
So here’s a question…

There are a lot of times where I’ll do something along the lines of:

for key, val in pairs(table) do
	val = variable
end

 Because that’s a quick and easy way of going through a table, rather than figuring out how many entries there are and setting up some index variable that will increment. The linting rules really don’t like the fact that in the above code ‘key’ is unused, so I get a ton of errors.

Is there another way of iterating through a table that the linter likes but retains the simplicity of pairs or ipairs?

jwink75:
It’s even worse than I thought, because I just got an error on:

for i = 1, 5, 1 do

Saying that I was an unused variable… But as my limited understanding of this kind of thing goes, it is NOT unused, it’s just all sort of internal to the loop (initial value is 1, go until it’s 5, increment by 1 each loop).

asherber:
All, sorry if I wasn't clear about this -- I'm not asking you to go through and fix issues, because I've already taken care of that. Please take a look at the copies of your scripts in the [head branch for this PR] and make sure that my altered versions work correctly.

@jwink75 I know that there are a bunch of things in your scripts that the linter doesn't like. Instead of overwhelming the comments here, would you mind either contacting me externally (my email address is in my scripts) or opening a [Discussion](https://github.com/finale-lua/lua-scripts/discussions/landing) if you'd rather keep this discussion on GH?

jwink75:
Happy to move to a discussion, but the link didn't work and I can't seem to find it in the GitHub UI, so forgive me one more comment here...

Aaron, I was testing my harp pedals script and there were a few things that are not working, so I pulled up an older version that is working and I'm going through and trying to address the linting issues myself... Partly as a learning exercise so I don't make those goofs going forward. Like I said above, though, there seem to be a number of things that shouldn't be errors, so I'm wondering how I can handle those better (aside from cheating the system by using luacheck: ignore)

cv-on-hub:
> take a look at the copies of your scripts in the `head branch`

Sorry - I also misunderstood. Your `luacheck` adjustment to that script is fine. I'll delete #661  and wait until the dust has settled for other changes.

asherber:
The discussions tab is right at the top of the GH UI:

image

@jwink75 The harp pedals script has lots of issues; you'll see that I addressed some of the low-hanging fruit but then marked the whole script with an ignore hint when I added the stricter rules. I'm happy do do a pair programming session with you and walk through some of what's going on. Please email me and we can set up a time that works for you.

ThistleSifter:
While I appreciate the convention of using an underscore for unused variables in a `for` iterator, I don't agree with enforcing it in this repo. Aside from being a convention and not a rule, it is a form of programmer's shorthand and shorthand only makes sense if you already know what it means. For those who aren't programmers, but have simply picked up enough knowledge to start writing scripts, it is more intuitive to write them out, easier to understand when returning to it in the future and harmless. We have a variety of contributors with varying backgrounds and experience in programming and I think we need to acknowledge that and work with it.

I'd like the calls to `local mixin = require("library.mixin")` at the top of each mixin file to be left in. It simplifies editing and costs nothing to have them in since the `mixin` library is what calls those files in the first place.

Please tag me in future.

src/library/mixin_helper.lua
```diff
@@ -528,7 +528,7 @@ function mixin_helper.create_custom_window_change_event(...)
 
     local function trigger_func(window, immediate)
         if type(window) == "boolean" and window then
-            for win in event.target_iterator() do
+            for _ in event.target_iterator() do
```
ThistleSifter, line 531: This line was correct as it was. The typo is in lines 533 and 535 below which should reference `win` instead of `window`.

asherber: Thanks, will fix.

asherber:
@ThistleSifter Thanks for your comments, and sorry I didn't tag you earlier.

I agree about supporting contributors of different levels of experience, but I also think it's worthwhile to nudge them towards better code. What I like about the underscore is that it should be the outcome of a virtuous cycle: write a loop, get a lint warning, realize that the variable isn't used, and acknowledge it as such. The lint warning itself is useful because it tells you that you're not using the variable, and maybe you meant to, in which case you can fix your loop body.

Similarly, I'm not crazy about putting those unused `local mixin` declarations back in if they're not actually being used -- but if you feel strongly about it I will (along with an ignore hint to make the linter happy).

jwink75:
> The discussions tab is right at the top of the GH UI:
> 
> image
> 

i checked again in case I was blind or crazy... The discussions button truly is NOT in my UI!
image

cv-on-hub:
Have now checked all of my updated scripts in this PR and they're all fine except for:

- `hairpin_creator.lua`: failing to initialise `y_current` to zero at line 551 results in an error in FCMCustomWindow, as it won’t accept a nil coordinate.
- `ostinato_maker.lua`: does `global_dialog` need to be global, or is that just if Modeless is in conjunction with RetainLuaState?

asherber:
@cv-on-hub 

1. Can you see where the error is thrown? It looks to me like `y_current` is initialized in other places before it is used.
2. Looking at the version of your script in this branch, you reference `global_dialog` inside a function [here](https://github.com/asherber/lua-scripts/blob/70d75ad13e989ec56f9d0cdf6b7e2e432dc9208c/src/ostinato_maker.lua#L319), which implies that the variable needs to be global.

cv-on-hub:
> 1. Can you see where the error is thrown?

1. Yes - you're right. I was testing a slightly different version of the file that needed initialization. This one doesn't.
2. The global_dialog call at [line 319](https://github.com/asherber/lua-scripts/blob/70d75ad13e989ec56f9d0cdf6b7e2e432dc9208c/src/ostinato_maker.lua#L319) presumed that the var was global modelled on all of Robert's old Modeless scripts. But it would also work with a local top-level var, and I see that you've changed that var to top-level local on his scripts as well.

asherber:
@cv-on-hub I did initially change those vars to top-level local, but after discussion with Robert I changed it back to global and added the ignore hint. Either way is fine from a linter point of view. If the var is local then it won't persist across invocations, but it sounds like you may not want that behavior anyway.

rpatters1:
@cv-on-hub The `global_dialog` variables must be global or else they can be collected by lua gc when the script is called again with `RetainLuaState`.

ThistleSifter:
> What I like about the underscore is that it should be the outcome of a virtuous cycle: write a loop, get a lint warning, realize that the variable isn't used, and acknowledge it as such.

Not necessarily. It depends on what's more important to the author: knowing what it is, or knowing whether it's intended to be used. Knowing what it is is more important (you can't decide whether it's needed without first knowing what it is), which is why the underscore only makes sense if you *already know* what the value is and you're sure you won't need the reminder (revisting old code a year or 2 later can be an eye opening experience, at which point it may as well have been written by someone else) and you're sure someone else won't need the hint either. Naming things is the best form of documentation we have, and I don't agree with forcing people to name things a certain way just to convince a linter of their intentions. Code is primarily meant to be read and understood by people, especially the person who wrote it, and if the author finds a certain way easier to understand and it isn't causing bugs or opening a potential pitfall, I don't see the issue.



> but if you feel strongly about it I will (along with an ignore hint to make the linter happy).

Thanks, that would be good.

asherber:
Okay, @ThistleSifter, I put the `mixin` references back.

As for the other issue, I think the linter rule about flagging unused variables is very useful. In fact, it helped bring to light at least two bugs in the library (one of which was the one you called out, above). Once the linter has flagged a variable like this, there's no rule that says you _have_ to fix it by changing it to an underscore, although I think in most cases it's the clearest and most expedient; you could also choose to leave the variable in place and satisfy the linter by adding an ignore hint to that line of code. 

(I actually agree with just about everything you say in your last comment, especially about readability; it just leads me to a slightly different conclusion!)

rpatters1:
I think my stuff is okay. If it isn't, I'll fix it in a future PR.

rpatters1:
I just realized there is at least one situation where we have to keep unused variables, and that is the `tie.lua` library.

Consider this code:

```lua
local layer, start_note, end_note = tie.calc_tie_span(note, true, true)
```

The `layer` variable may never actually be used in the Lua script, but it is the container for `start_note` and `end_note`. So `layer` is very much under the covers being used by the PDK Framework. It is also keeping the `start_note` and `end_note` instances alive. I'm pretty sure it would be acceptiable to change this to

```lua
local _, start_note, end_note = tie.calc_tie_span(note, true, true)
```

But it is something we need to be aware of.

asherber:
@rpatters1 What do you think? This has been open for a week -- time to merge it?

rpatters1:
I've been waiting to merge anything else. lol.

jwink75:
I say we do it… everything I’ve tested seems to work other than the harp wizard, which we commented out and will tackle later. If there are any other problems with my scripts I can always fix later.Sent from my iPhoneOn Jan 30, 2024, at 6:10 PM, rpatters1 ***@***.***> wrote:﻿
I've been waiting to merge anything else. lol.

rpatters1:
Is there any way to force all the backlog of PRs to run the lint check before we try to merge them?

asherber:
Closing and reopening each PR works. I'll take care of it now.

PR #656, opened 2024-01-23
Update pitch_singles_changer.lua

cv-on-hub:
Reverting to `cjson = require("cjson")`


asherber:
Thanks -- I'm actually surprised this made it through the build process!

PR #659, opened 2024-01-25
Update selection_refiner.lua

cv-on-hub:
simpler staff selection using popup menus; simpler key command mapping; some bolding in dialog window.


asherber:
@cv-on-hub Please fix the one [lint warning](https://github.com/finale-lua/lua-scripts/actions/runs/7720650038/job/21045888325?pr=659)

PR #661, opened 2024-01-26
Revise ostinato_maker.lua

cv-on-hub:
Revised for new Linting rules, with just one global, `global_dialog`. Moved its initialisation to top-level, mimicking Robert's sample script `modeless_dialog.lua`. Other minor tweaks, like `script_name` coming from `general_library.lua` and removing trailing dots from plugindef().


asherber:
@cv-on-hub Robert and I discussed this idiom, which is used by 6 of your and his scripts. We agreed that moving these globals to the top-level scope resulted in inelegant code, and so we agreed that cases like this could use luacheck hints to bypass that check. If you look [here](https://github.com/asherber/lua-scripts/blob/c6f6156f436dd56424d484aeb0f1a0fc7dd374cb/src/ostinato_maker.lua#L60), you'll see a hint line that tells the linter not to bother with those specific checks for those specific variables in this file. 

So what I would suggest here is to go back to the version of your script currently in `master` and make any substantive revisions you want -- but don't worry about the linting.

jwink75:
I am using the luacheck hints to bypass all these complaints about the unused variables in pairs and ipairs… but it feels like cheating :P

asherber:
@jwink75 In general, it is cheating, and also unnecessary. Please see my comment in the other PR. I have already fixed all of the issues in your page wizard; your harp pedal script needs further discussion. Please email me or start a discussion thread to get into specifics.

cv-on-hub:
OK so I'll split the difference. Converted global_selection to (local) current_selection and reverted to original script structure. Kept the luacheck hint and those other minor alterations.

asherber:
@cv-on-hub Please merge master into this branch to resolve conflicts

PR #665, opened 2024-01-30
Generate RTFNotes from Notes during build process

asherber:
I basically hooked into the class that injects the `HashURL` so that it also injects `RTFNotes`. And the design is a little more modular now, so we can more easily inject other elements in the future, if needed. 

I confirmed locally that dist files built in this branch differ from master only by the presence of `RTFNotes` (where appropriate) and the lack of whitespace mangling in `plugindef`. 

I'm happy for any suggestions or questions.

PR #666, opened 2024-01-31
Add utility function for displaying script notes

asherber:
It would be better to merge this after #665, in case there are any changes needed here (and so that the repo actually has RTFNotes to display).

The modal dialog will display RTFNotes in a (readonly) text edit control, falling back to Notes. If neither one exists, then no dialog is shown. The caller can specify the dialog caption, height, and width; sensible defaults are provided. 

@cv-on-hub Please see if this meets your needs as a replacement for the custom code you've got behind your `?` buttons. Note that after the other PR is merged all scripts with Notes will automatically get RTFNotes during the build process.

### DIFF
--- a/src/library/utils.lua
+++ b/src/library/utils.lua
@@ -292,4 +292,72 @@ function utils.rethrow_placeholder()
     return "'" .. rethrow_placeholder .. "'"
 end
 
+--[[
+% show_notes_dialog
+
+Displays a modal dialog with the contents of finaleplugin.RFTNotes (if present) or finaleplugin.Notes. If neither one is present, no dialog is shown.
+
+@ caption (string) The caption for the dialog. Defaults to plugin name and version.
+@ width (number) The width in pixels of the edit control. Defaults to 500.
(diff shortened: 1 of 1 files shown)


rpatters1:
One thing to be aware of about this function is that for RTF display it requires RGP Lua 0.68+. (And depending on which APIs you use, possible even 0.71, but I doubt you are using `GetCharacterAtIndex`, which will be replaced in 0.71 with a different API due to its lack of proper UTF-16 support.)

But then, multiline edit boxes only work on Mac. I assume that's what @cv-on-hub was using to show info before I added the text editor control. So any script that shows info in a multiline edit box must use FCCtrlTextEditor for cross platform compatibility and should therefore require 0.68 minimum.

rpatters1:
Eventually, it may make sense to expand this window to include all the functionality and display of the JW Lua info dialog. (You access that from the Explorer tab.) That will save me the trouble of doing it in RGP Lua.

But we can leave that as a future enhancement.

PR #674, opened 2024-02-01
Update utils.lua

cv-on-hub:
The default font size in `utils.show_notes_dialog()` is illegible for me so I propose a fourth optional input parameter, `fontsize`.

### DIFF
--- a/src/library/utils.lua
+++ b/src/library/utils.lua
@@ -300,14 +300,16 @@ Displays a modal dialog with the contents of finaleplugin.RFTNotes (if present) 
 @ caption (string) The caption for the dialog. Defaults to plugin name and version.
 @ width (number) The width in pixels of the edit control. Defaults to 500.
 @ height (number) The height inpixels of the edit control. Defaults to 350.
+@ fontsize (number) Font size in points. Defaults to 10.
 ]]
-function utils.show_notes_dialog(caption, width, height)
+function utils.show_notes_dialog(caption, width, height, fontsize)
     if not finaleplugin.RTFNotes and not finaleplugin.Notes then
         return
     end
     
     width = width or 500
     height = height or 350
+    fontsize = fontsize or 10
     
     if not caption then
         caption = plugindef()
@@ -350,7 +352,7 @@ function utils.show_notes_dialog(caption, width, height)
             else
                 local edit_font = finale.FCFontInfo()
                 edit_font.Name = "Arial"
-                edit_font.Size = 10
+                edit_font.Size = fontsize
                 edit_text:SetFont(edit_font)
                 edit_text:SetText(notes_str)
             end


asherber:
I think the goal is to find something that works for most people. For any given font size, whether it's a function default or user-specified, there are going to be some people who find it too large and some people who find it too small. Some of this has to do with the resolution of the user's monitor, some with the user's eyesight, and some (I think) with OS differences.

I don't like the idea of making the font size in this dialog developer-selectable, because I think it will lead to different authors picking different sizes, leading to different problems.

On my (fairly vanilla) Windows computer, Arial 10 is larger than the default font used in dialog boxes and other Lua dialogs:

image

image

image

@cv-on-hub Are you on Windows or Mac (I think Mac)? Can you post screenshots? If Arial 10 is smaller than the font in system dialogs, then I'm in favor of this function choosing a different font size based on OS. (The sample script for the text control by @rpatters1 does exactly that.)

asherber:
Oh, I forgot something: For RTF, the font size is specified in the RTF itself, which is created during the build process. It's possible to do OS sniffing at the time of display and make adjustments to the RTF, but I'd rather avoid that if possible by finding a single default font size that is reasonably good on both OSes.

asherber:
And adjusting the RTF at display time is pretty much a lost cause if the dev has supplied his own RTF.

cv-on-hub:
Yes - I'm on Mac. I attach a combined pic of the standard Mac text options, which should all be 1-to-1 by pixel. 10-point  Arial looks too small but Monaco on Mac is easier to read at small sizes than Arial. Robert's text_control sample suggests 9-point Consolas for Win, 11-pt Monaco for Mac.  So maybe use those choices here at either 9-11 or 10-12 point size. 

All mac texts

cv-on-hub:
In terms of the RTF extraction to RTFNotes, my raw `pandoc` output doesn't specify font size, so are you supplying a template with font size? If you omit that spec won't it follow the size of the display window?

Still leaving the tiny problem of the literal tab spacing.

asherber:
1. To my eyes, the 10pt Arial in your screenshot looks slightly smaller than the Mac alert font, which is slightly smaller again than the standard dialog font. So that's different than on Win, where Arial 10 is slightly larger than those fonts.
2. Yes, I'm supplying a template with font size. If I leave that out, then the control picks a default size, which looks to be 12pt on Win. To my eyes, that's too large. (Frankly, even 10pt is larger than I would personally like; I'd prefer 9.) Here's what 12pt looks like on Win:
image
3. I am hesitant to do OS sniffing at runtime and change the font size. For one thing, it wouldn't work for user-supplied RTF (which I don't think we'll have much of, but still). For another, I would likely have to adjust heading sizes as well. It's just all a little janky.
4. I would suggest not trying to get tabs to work. The Markdown info I've seen doesn't really have a role for the tab character, and in fact it often gets condensed to a space (as it does in HTML). I see what you're doing in your screenshot, and it looks very nice -- but I think there are other ways of presenting the info clearly. (It really wants to be a table, but tables have their own issues.)

If 10pt is too small on a Mac, then maybe 11pt is a reasonable compromise for both platforms -- larger than ideal on Win, smaller than ideal on Mac. Here's what 11pt looks like on Win:

image

I'd love to hear other opinions about how this looks.

As for detailed formatting, my suggestion would be to keep in mind that we're trying to make all of this work within a basic framework that needs to give reasonable results when the Markdown is converted to HTML (for the website) and when it's converted to RTF (for info display). I think what we have now offers much more than what we had before, but it's not meant to be a detailed layout engine.

cv-on-hub:
> it's not meant to be a detailed layout engine.

True. And I've discovered that &#9; converts neatly to \\tab in RTF which works perfectly in my INFO dialogs and won't look horrible in whatever Vercel does to it. (I don't need many of them). Which means I won't bother with custom RTF and there's one less cause for concern. 

11pt Arial is borderline for me on Mac, even though too big on Win. Monaco looks better but being a monospaced "code" font has no bold.

11pt arial

jwink75:
11pt on Windows seems reasonable to me

asherber:
I may have a workable solution for this, but it'll be a couple of days before I have time to work on it.

rpatters1:
About tabs, the control has a tabwidth setting. Pandoc may have an option to set tabwidths as well. Pandoc has an extension framework that allows you to write extensions (in Lua of all things), but I haven't tried to use it.

asherber:
I strongly advise not mucking with tabs.

IMO, the goal of the Notes "ecosystem" is to legibly present information in three ways:

1. The Notes field should be human-readable.
2. The Notes field should be transformable to readable HTML for display on the website.
3. The Notes field should be transformable to reasonable RTF that can be displayed in a dialog.

Using `&#9;` subverts both 1 and 2: It is not human-friendly in the raw Notes text, and when transformed to HTML it renders as whitespace and collapses with other whitespace around it (i.e., it doesn't function as a tab stop).

cv-on-hub:
> Using `&#9;` subverts both 1 and 2

So I guess if there are cases where I feel strongly enough about tab alignment I can revert to my current system of a standalone `info_notes` string rendered in a standard `AlertInfo()` window. The system is simple enough to fathom and future editors can replace or ignore it. (I care more about word alignment than bold/italic!)

asherber:
Script authors are free to do what they like. Personally, I have a strong allergy to duplicated blocks of code or text that need to be manually kept in sync. My own approach would be to see how I can clearly present the information, given the building blocks that the library provides; to me, overall clarity and ease of use are more important than typographical details. YMMV.

cv-on-hub:
I'm not entirely convinced that this:
```
**w** &#9; flip [copy articulations]  
**e** &#9; flip [copy expressions]  
**r** &#9; flip [copy smartshapes]
```

is significantly more inhuman than this:
```
**w** - flip [copy articulations]  
**e** - flip [copy expressions]  
**r** - flip [copy smartshapes] 
```

asherber:
I suspect that HTML entities are pretty opaque to people with less development experience. Kind of looks like cursing! 😄 

Just off the top of my head, I think that to bold a single character like that, it might be more readable with underscores rather than asterisks. And if the ragged margin of the second column bothers you, maybe stacking them might be more effective:

```
__w__
flip [copy articulations]

__e__
flip [copy expressions]
```

asherber:
Please take a look at #676 as an alternative.

cv-on-hub:
As @asherber says, this is replaced by #676

PR #675, opened 2024-02-01
Linting tweaks

asherber:
1. Change recommended VS Code extension
2. Put back unused function params with underscore (now that we have luacheck versions straightened out)
3. Delete debugging file


rpatters1 (approved):
lgtm

src/tuplet_chooser.lua
```diff
@@ -3,7 +3,7 @@ function plugindef()
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "https://carlvine.com/lua/"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "v0.70"
+    finaleplugin.Version = "v0.70.1"
```
rpatters1, line 6: This is the wrong version number that crept into this script. I think it should be MinJWLuaVersion, but the author will need to confirm.

rpatters1 (changes requested):

src/tuplet_chooser.lua
asherber: The initial commit of this file had both `Version` and `MinJWLuaVersion` as 0.62, and `Version` has been incremented in each subsequent commit.

@cv-on-hub Can you confirm that the plugin version number is correct? (I added the `.1` because I restored a function parameter that I had removed in the first linting commit.)

cv-on-hub: The version I released on 2023/11/15 was `finaleplugin.Version = "v0.70"`. 
So your update to "v0.70.1" is sensible.
`finaleplugin.MinJWLuaVersion = 0.62` is correct.

rpatters1 (approved):

PR #676, opened 2024-02-03
Teach utils.show_notes_dialog() to use mac/win sizes

asherber:
### tl;dr

The `RTFNotes` generated by the build process now allows for different font sizes on Win and Mac. You can use [this standalone script](https://gist.github.com/asherber/85cfe0fb44a650fe316cb162f1de9697) to test the functionality.

### explanation

While the build process will continue to embed in `RTFNotes` the default font sizes for Windows, it will also add to the RTF a comment string containing a JSON fragment with substitutions to make for Mac. At display time, the `show_notes_dialog()` determines whether to make these substitutions based on the current OS.

The hope is that we can agree on default sizes for each OS that work reasonably for everyone, so that `RTFNotes` works transparently without additional effort. (Current defaults are 9pt for Win and 12pt for Mac.) Individual script authors can still choose to include their own RTF with their scripts. If developers are working on Mac, they can also leave those font sizes embedded and change the comment string to provide the substitutions for Win by setting the `os` element to `win`.

This is intended to supersede #674

### DIFF
--- a/.github/actions/bundle/dist/custom_template.rtf
+++ b/.github/actions/bundle/dist/custom_template.rtf
@@ -2,8 +2,8 @@
         {\rtf1\ansi\deff0{\fonttbl{\f0 \fswiss Helvetica;}{\f1 \fmodern Courier New;}}
         {\colortbl;\red255\green0\blue0;\red0\green0\blue255;}
         \widowctrl\hyphauto
-        \f0\fs20
-        \f1\fs20
+        \$basefont$
+        {\info{\comment $fontsizes$}}
         $body$
         }
     
--- a/.github/actions/bundle/dist/index.js
+++ b/.github/actions/bundle/dist/index.js
@@ -6987,14 +6987,39 @@ const getRTFNotes = (input) => {
     const notesRegex = /(?<=finaleplugin.Notes = \[\[).*(?=\]\])/ius;
     const match = input.match(notesRegex);
     if (match) {
-        const notes = (0, dedent_js_1.default)(match[0]);
+        let notes = (0, dedent_js_1.default)(match[0]);
+        notes = notes.replace(/^#{4,}/gm, '###');
         const templateFile = path_1.default.join(__dirname, "custom_template.rtf");
-        const args = ['-f', 'markdown', '-t', 'rtf', '-s', `--template=${templateFile}`];
+        /**
+         * Default font sizes are:
+         *         p   h1  h2  h3
+         * pandoc  24  36  32  28
+         * win     18  26  23  20
+         * mac     24  32  29  26
+         *
+         * The win p size is injected as a variable into the call to pandoc, and the
+         * other win sizes are replaced below. The mac sizes are injected, as a JSON
+         * fragment, into the RTF comment field; utils.show_notes_dialog() can use
+         * this to perform replacements.
+         */
+        const args = `
+            -f markdown 
+            -t rtf 
+            --standalone 
+            --template=${templateFile}
+            -V basefont=fs18
+            -V fontsizes="os":"mac","fs18":"fs24","fs26":"fs32","fs23":"fs29","fs20":"fs26"
+        `.trim().split(/\s+/);
         const pandocResult = (0, child_process_1.spawnSync)('pandoc', args, { input: notes, encoding: 'utf-8' });
         if (!pandocResult.error) {
-            result = pandocResult.stdout.replace(/fs28/g, 'fs24')
-                .replace(/fs32/g, 'fs28')
-                .replace(/fs36/g, 'fs32');
+            result = pandocResult.stdout.replace(/(?<=\\)fs\d\d/g, function (match) {
+                switch (match) {
+                    case 'fs36': return 'fs26';
+                    case 'fs32': return 'fs23';
+                    case 'fs28': return 'fs20';
+                    default: return match;
+                }
+            });
             result = `    finaleplugin.RTFNotes = [[${result}]]`;
         }
     }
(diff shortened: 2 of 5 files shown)


rpatters1:
This looks good to me. Let's wait until we hear from @cv-on-hub to merge.

cv-on-hub:
That works well for me on Mac - congratulations and thanks!

Although it's assumed that `RTFNotes` will always be there (in `dist`), they WON'T be in the user's local script if they want to use the system default. In which case your `dedent()` routine does nothing useful to display Markdown `Notes` in their place. A couple of simple `string.gsub()`s work better. I'm working on one - not quite ready yet, but this is a step forward:
```
local notes = finaleplugin.Notes:gsub("^%s+", "")
     :gsub("[ \t]*\n[ \t]*(%S)", " %1"):gsub(" *\n *", "\n\n")
```

asherber:
I don't understand what you're saying about the Markdown in Notes. `dedent` will remove leading indentation, and the function will then display the Markdown _as text_ -- it doesn't make an attempt to render it, but one of the points of Markdown is that it's reasonably human-readable on its own. Assuming that Notes looks readable (albeit indented) in the source file, it should display fine in a text control or dialog. Are you seeing something different?

cv-on-hub:
Why not just add a few simple checks, beyond leading whitespace, to at least remove unintended line breaks? Not a "render" but at least closer to readable "simple" Markdown contents. One line of code. `RTFNotes` can conceivably fail so why not have a slightly more appealing backup?

asherber:
How do we determine whether a line break is unintended? Can you give an example?

cv-on-hub:
The only intentional line breaks are (1) line ends in two spaces; (2) blank line between lines.

All other lines ending in one or no spaces should render jointly as a complete paragraph.

asherber:
Are you seeing cases where lines improperly fail to render as a paragraph? Can you post an example?

Are you thinking that someone might type in a block of text and accidentally leave two spaces at the end of a line? And doesn't that really have to do with translation to RTF/HTML, not with just displaying the Notes as text?

I think I'm not understanding the problem you're trying to avoid.

cv-on-hub:
Text we put into `finaleplugin.Notes` is rendered on finalelua.com with redundant lines breaks removed. Below is the difference in some "simple" formatted markdown, using your `dedent()` method (top) next to mine. It's simple cosmetics, but not without utility. Your method also assumes regular indenting throughout which can never be assumed in user-supplied markdown.

your md
my md

cv-on-hub:
I might be missing something (it happens) but I thought this would be neater and more general for "dedenting" RTFNotes ...

`notes = finaleplugin.Notes:gsub("^%s+", ""):gsub("[ \t]*\n[ \t]*(%S)", "\n%1")`

cv-on-hub:
If you're prepared to contemplate this, I attach here a working script that demonstrates "basic" (non-styled) Markdown formatting that observes paragraph joining, forced newlines (double spaces) and basic lists (line begins with hyphen), with all other whitespace trimmed.  180 code characters in six `gsub`s.
[test2.lua.txt](https://github.com/finale-lua/lua-scripts/files/14155504/test2.lua.txt)

asherber:
Okay, now I understand what you're saying about the Markdown display. I've thought about it a bit, and my inclination is to leave it alone. Here's my process:

- Markdown is text, with a set of conventions that govern how that text can be rendered as HTML. But it's still just text, and it should look presentable _as text_ -- for example, to anyone looking at the source code for a script.
- On the website, individual lines of Markdown are concatenated into paragraphs by the rendering process.
- I don't want to be in the business of rendering Markdown in Lua. Rendering is complicated, which is why there are libraries to do it -- and I think adding a Markdown library to this repo (or to RGP Lua) is way out of scope. In fact, this is why I added the generation of RTF to the build process, so that in Lua we can display something prettier without directly worrying about the rendering.
- Given that, I don't want to be in the business of _partially_ rendering Markdown in Lua (e.g., rendering paragraphs but not styles). That's sort of where we started with all of this -- the website was only rendering a subset of the Markdown, so I added a real Markdown library. If you're only doing a partial rendering, then the question comes up of why _this_ renders but not _that_, and it's a slippery slope.
- Nobody is forcing script authors to use `show_notes_dialog()`. Many scripts don't have notes, many have only a brief description that's not worth displaying in an info dialog. The ones who want to make use of it will need to add a button for it and will presumably test the display before submitting the script to the repo, and if it doesn't look right, they'll figure out how to adjust the text appropriately.

asherber:
And responses to your other items:

- In this PR, I actually do have `edit_font.Size = finenv.UI():IsOnWindows() and 9 or 12` -- I caught the omission after I prepared the standalone script in Gist. 😄 
- The point of `dedent` is not to remove all leading whitespace, but just to remove from all lines however much whitespace appears before the first line. (It's actually a simplified version of a method that handles more complicated cases.) The reason that we shouldn't remove all leading whitespace is:
```lua
finaleplugin.Notes = [[
    Here's a sample function:

    ```
    local function sample()
        do_something()
    end
    ```
]]
```

If we remove all leading whitespace, then the function loses its native indentation. You're right that this assumes that the text uses simple block indentation to begin with; I think that's a reasonable assumption, based on what I've seen, and as I noted above, script authors will presumably test out the display before committing their scripts.

cv-on-hub:
> I don't want to be in the business of rendering Markdown

I figured as much. Obviously I disagree, and for a certain subset of Lua authors offering a partial render when no typeface variants are available is better than none and I find much bald Markdown considerably inhuman. We only use Markdown in the first place because it's "required" by the website. (Plaintext script comments get to use ASCII-art!)

But not to worry - it's only the backup system. This revision with OS-customised fontsize for both RTF and plain text is admirable.

cv-on-hub:
(FWIW: I quibble, but my `:gsubs` produce much the same output as `pandoc -f markdown -t plain`, though would surely fail in more complex cases.)

asherber:
It sounds like @cv-on-hub and I have reached a state of "agree to disagree". If I'm reading his next to last comment correctly, it sounds like he does not object to this PR (even if he wishes it behaved somewhat differently). 

@rpatters1 If this sounds right to you, then I think we can go ahead and merge.

rpatters1:
It's fine with me.

cv-on-hub:
And fine with me.

PR #678, opened 2024-02-05
Don't take dependent actions in forks

asherber:
This fixes an issue that seems to have been going on for a while, unrelated to recent changes made to the GH actions for this repo.

The actions in the `deploy.yml` file are set to run on any push to `master` -- but the ones that require the secret token from this repo (checkout docs, commit to this repo, commit to docs) fail on forks because they don't have the token. This PR changes `deploy.yml` so that these actions only run when pushing to `master` _in this repo_, not in a fork. The build actions that don't require a token will still run in forks.

cc: @jwink75


rpatters1:
Yay! Do it.

PR #679, opened 2024-02-05
Switch from lunajson to cjson

asherber:
Looks like my document options script was the last one to use `lunajson`, so I replaced it with `cjson` and also removed the `lunajson` files from the repo.


rpatters1 (approved):
lgtm

PR #684, opened 2024-02-08
Update utils.lua

cv-on-hub:
Does this help? Adding optional `parent` input to `show_notes_dialog()`.

### DIFF
--- a/src/library/utils.lua
+++ b/src/library/utils.lua
@@ -306,6 +306,9 @@ function utils.show_notes_dialog(caption, width, height, parent)
     if not finaleplugin.RTFNotes and not finaleplugin.Notes then
         return
     end
+    if parent then
+        assert(parent.ExecuteModal, "argument 4 must be nil or an instance of FCResourceWindow")
+    end
 
     local function dedent(input)
         local first_line_indent = input:match("^(%s*)")
@@ -353,9 +356,6 @@ function utils.show_notes_dialog(caption, width, height, parent)
 
         width = width or 500
         height = height or 350
-        if (parent and not parent.ExecuteModal) then
-            parent = nil -- probably not a valid window object
-        end
         
         local dlg = finale.FCCustomLuaWindow()
         dlg:SetTitle(finale.FCString(caption))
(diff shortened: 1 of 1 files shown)


src/library/utils.lua
```diff
@@ -352,6 +353,9 @@ function utils.show_notes_dialog(caption, width, height)
 
         width = width or 500
         height = height or 350
+        if (parent and not parent.ExecuteModal) then
+            parent = nil -- probably not a valid window object
+        end
```
rpatters1, line 358: Passing a `parent` that is not a window is a program bug and should therefore throw. You have two choices:

- let the call to `ExecuteModal` throw if it's the wrong type
- use `assert`, e.g.:

```lua
assert(parent and not parent.ExecuteModal, "argument 4 must be nil or an instance of FCResourceWindow")
```

rpatters1 (changes requested):

src/library/utils.lua
ThistleSifter: `assert` throws the error where it is called. For something like argument validation, I usually find it more helpful to have the error be thrown a level higher, to where the current function was called. If you do want to do that, you can follow this proforma:

```lua
if argument_is_bad then
    error("error message", 2)
end
```

cv-on-hub: Very good. Amended. This now throws a helpful error message back at the right part of the calling script.

asherber:
Can you please summarize the issue this is intended to solve?

rpatters1:
I can summarize it. The issue is that when you launch a modal window from another window, that window needs to be specified as the parent. This is of particular importance on Windows. On Windows, if you don't properly specify the parent, the modal window ends up modal with respect to the Finale window but not modal with respect to the window that launched it. That is, you will be able to access and interact with the window that launched it unless you specify the launching window as parent.

asherber:
@rpatters1 Particularly on Win, but the param can be passed on both platforms with no ill effect?

I'd like to suggest changing the order of params to the function--a breaking change to a function no one is using yet. 😄 It sounds like `parent` will be used for most calls, and the other three will be used less often (defaults should generally be fine). So I think `parent` should be moved to the head of the param list, so that callers don't need to pass three nils every time.

rpatters1:
It actually has benefit on macOS as well: it centers the new window on the parent if there is one.

I'm okay with a breaking change in the order of parameters.  @cv-on-hub will need to add corrections in the PR in every script where it is called.

cv-on-hub:
Good suggestions. I've checked `/src/` and no scripts currently use `show_notes_dialog` so changing parameter order looks benign. I've also added  `:gsub("%.%.%.", "")` to the default `caption`.

rpatters1 (approved):

PR #685, opened 2024-02-08
Update FCMCtrlStatic.lua

cv-on-hub:
I realise @ThistleSifter is changing things right now, but noticed this error and thought it might continue unnoticed.


rpatters1:
I was going to include this in the localization PR, but let's go ahead and fix it.

PR #687, opened 2024-02-09
RTF fix 

### LINKED ISSUE #686: [BUG] RTFNotes eats AdditionalMenuOptions et al
**Describe the bug**
When the bundler builds the `RTFNotes` from `Notes` it erroneously eats the `AdditionalMenuOptions`, `AdditionalUndoText`, and `AdditionalDescriptions` if they follow `Notes`, which they often do. 

**To Reproduce**
Examine any `dist` version of a script with `AdditionalMenuOptions`. `beam_selected_region.lua` is an example.

@asherber

asherber:
Fixes #686 

I had a greedy regex where a non-greedy one was wanted, and I didn't have a test case to cover this situation. When I committed the initial code, I verified visually that the only changes in dist were that `RTFNotes` got added -- but I didn't look closely enough to see that the field contained more than it should have. All better now.

### DIFF
--- a/.github/actions/bundle/dist/index.js
+++ b/.github/actions/bundle/dist/index.js
@@ -6984,7 +6984,7 @@ const getHashURL = (name) => {
 };
 const getRTFNotes = (input) => {
     let result = '';
-    const notesRegex = /(?<=finaleplugin.Notes = \[\[).*(?=\]\])/ius;
+    const notesRegex = /(?<=finaleplugin.Notes = \[\[).*?(?=\]\])/ius;
     const match = input.match(notesRegex);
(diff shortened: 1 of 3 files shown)


rpatters1:
My regex experience is limited to using the ones ChatGPT tells me to. If you're satisfied that it works, let's merge it.

asherber:
You're familiar with the old chestnut?

>Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.

This change passes the eye test and also passes the test case I added to the suite. And the RTF for `beam_selected_region.lua` looks correct. Merging.

Issue #692, opened 2024-02-11
[BUG] Localized name not displaying correctly


asherber:
For the scripts that newly take advantage of localization options, returning `t.menu, t.menu, t.desc` doesn't seem to have the desired effect. In my Lua menu, I'm seeing the name of the script file (e.g., `baseline_move_reset.lua`) instead of the expected text. 

Win 11, Fin 27.4.1, RGPLua 0.70. For some reason, I seem to be unable to take a screenshot inside of Finale at the moment. I'm using finale_lua_menus.txt to organize things under Plug-ins | Lua, but I see the same behavior when I delete that file.

cc: @rpatters1

rpatters1:
Do any of the additional menu items appear?

rpatters1:
OK, I have reproduced it. Will explore later today.

rpatters1:
This issue is that RGP Lua pre-0.71 is search for this line of code:

```lua
function plugindef()
```

I have fixed this issue by modifying the localized `plugindef` function definitions as follows:

```lua
function plugindef(locale)
--[[
-- This comment allows RGP Lua pre-0.71 to find the plugindef function
function plugindef()
--]]
```

As you can see, this allows any version of RGP Lua to find a valid compilable Lua function. (It also works on JW Lua, which appears only to search for `function plugindef`, like 0.71 does now.)

The only reason I am not sure it will work is I need to be sure the bundler does no strip out these comments.

rpatters1:
Maybe we could skip the stripping of comments for the script itself, and only do it for libraries. Or maybe we could actually inject this comment into each plugindef function.

asherber:
I think it's likely that the bundler will strip that comment, but I'm not at my computer. And injecting it would solve the issue for dist scripts but not src.

asherber:
The bundler does strip that comment, but it's easy to tell it not to strip comments from `plugindef()`. Is that the approach you'd like to take?

In terms of aesthetics, would JW/RGPLua be happy with just a single-line comment, or does it need to be formatted just that way? Does `function` need to be flush left?

rpatters1:
It has to be a multiline comment, but the function statement could be preceded by whitespace, e.g.

```lua
--[[
   function plugindef()
--]]
```

I think it looks better with no whitespace, but ymmv.

[...]

PR #693, opened 2024-02-11
Don't strip plugindef comments

asherber:
Related to #692 

I also added a commit that removes "boilerplate" comments from existing scripts. I can put these back if you want, though they seem unnecessary to me.

@rpatters1 You'll need to add the new lines to your localized scripts, either in this PR or one after it.


rpatters1:
Looks good to me. The boilerplate was put there by JW Lua when you used the dialog box to create a new script.

I have a PR ready to go.

Issue #691, opened 2024-02-10
Rest Slider modeless


rpatters1:
The Rest Slider would be way more useful if it were modeless. Is there any reason it can't be? Also, either way it should show the selected region somewhere.

@cv-on-hub

cv-on-hub:
For once I agree with you about modeless. I just realised yesterday how much more useful it would be that way. I also think it should replace `rest_offsets.lua` which offers less functionality less elegantly. Will submit a PR incorporating modeless, selection confirmation and deletion of  `rest_offsets.lua`.

asherber:
Please don't get rid of Rest Offsets--I find it much more usable than the slider.

cv-on-hub:
> Please don't get rid of Rest Offsets

Good to know. It shall stay.

rpatters1:
As an aside, I admire your ingenuity in using a single edit control to trigger UI events. It has drawbacks though. It doesn't work very well with Windows. That's because Windows allows you to tab to any control, and once the edit control doesn't have focus, the keystrokes don't work any more. This happens, for example, if you move the slider.

There are a number of other issues with the windows version. The slide covers up half the next line and there are some alignment issues between the text fields and the buttons.

Your approach with keystrokes also fails on macOS if you turn on keyboard navigation in your Mac. You can enable this mode in System Preferences->Keyboard->Shortcuts. At the bottom there is an option "Use keyboard navigation to move focus between controls." When this option is enabled, you can tab between all the controls, not just the text fields. There is a combo keystroke to toggle this mode, but I don't recall offhand what it is.

There isn't an obvious or standard way to accomplish exactly what you want. Keyboard navigation is supported on both OSes just as I have described. It might be a better long term option to use it rather than trying to invent our own that seems a bit hackish if well executed on the Mac half (when the Mac is in the right mode). Stealing characters from an edit box is a snake in the grass waiting to bite some future user of this repo.

jwink75:
lol I feel like that last was aimed at me, intentional or not :) I was *TOTALLY * planning on reverse engineering some of Carl’s work to see just how he handled those keyboard shortcuts!

rpatters1:
On Windows it seems like could almost be user-hostile. At best it's not very effective. The situation is slightly less bad on macOS because the keyboard navigation option does not appear to be that well known in the user community. (I didn't know about it until 12-18 months ago.)

cv-on-hub:
> On Windows it seems like could almost be user-hostile

[...]

PR #696, opened 2024-02-13
Update rest_slider.lua

cv-on-hub:
Converting to `modeless` operation, in response to #691, required major surgery. Produces many `undo_block` entries on slider action but can find no way to avoid this while retaining continuous document screen updating.

### DIFF
--- a/src/rest_slider.lua
+++ b/src/rest_slider.lua
@@ -1,87 +1,76 @@
 function plugindef()
-    finaleplugin.RequireSelection = true
+    finaleplugin.RequireSelection = false
+    finaleplugin.HandlesUndo = true
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "https://carlvine.com/lua/"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "0.15"
-    finaleplugin.Date = "2023/12/15"
+    finaleplugin.Version = "0.21"
+    finaleplugin.Date = "2024/02/13"
     finaleplugin.CategoryTags = "Rests, Selection"
-    finaleplugin.MinJWLuaVersion = 0.68
+    finaleplugin.MinJWLuaVersion = 0.70
     finaleplugin.Notes = [[
         Slide rests up and down on the nominated layer with continuous visual feedback. 
         This was designed especially to help align rests midway 
         between staves with cross-staff notes. 
-        The "mid-staff above" and "mid-staff below" buttons achieve this with one click. 
-        Cancel the script to leave rests unchanged.
+        The _Mid-Staff Above_ and _Mid-Staff Below_ buttons achieve this with one click. 
+        Cancel the script to leave rests unchanged. 
 
-        "Reset Zero" sets nil offset. 
-        Note that with transposing instruments this is NOT the middle 
-        of the staff if "Display in Concert Pitch" is selected. 
-        In those instances use "Floating Rests" to return them to 
+        _Reset Zero_ sets nil offset. 
+        Note that on transposing instruments this is NOT the middle 
+        of the staff if _Display in Concert Pitch_ is selected. 
+        In those instances use _Floating Rests_ to return them to 
         their virgin state where the only offset is that set at 
-        Document → Document Options → Layers → Adjust Floating Rests by...
+        _Document_ → _Document Options_ → _Layers_ → _Adjust Floating Rests by..._
 
         At startup all rests in the chosen layer are moved to the 
-        same offset as the first rest in the selection. 
+        same offset as the first rest on that layer in the selection. 
         Layer numbers can be changed "on the fly" to help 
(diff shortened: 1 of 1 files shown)


cv-on-hub:
> what about using `RegisterHandleActivate`?

Great idea. I'd tried many other options but not the `Activate` handler. But unless I'm doing it wrong the _Undo Stack_ is very picky and doing this fails to return the active tool to `Selection` at the end of the process. Same as every other failed option that doesn't track every single change that's interpreted as an __Edit__ change. 

Less important, but a single "wrapping" __undo__ can't display the current (changing) score selection or layer number. I also wonder whether closing the dialog on __OK__ would be better (though then resembling `Modal` operation) but I can't find the magic combination to achieve that with `OkButtonCanClose`.

rpatters1:
You can make the OK button close by resetting the `OkButtonCanClose` option inside the InitWindow handler. The reason it's a bit difficult is that the `mixin` library assumes you do `not` want to close on OK and sets it accordingly inside `RunModeless`.

I think the choice boils down to whether multiple undo entries is more annoying or modal operation is more annoying. Even if you reset `OkButtonCanClose` you can't count on the OK button with a modeless. In fact the who reason for the modeless is to be able to operate on multiple selection regions without closing the window. Speaking for myself, I prefer the modeless dialog, but others may feel differently.

There might be a way to notify when the slider button is released. I'll add a note to myself to investigate it. It wouldn't help if you were using arrow keys to move it, but there might be a way if you are using the mouse to move it. Either way, it will have to wait for another RGP Lua release.

cv-on-hub:
> the `mixin` library assumes you do _not_ want to close on OK

Ah yes - I remember this issue from ages ago (early foray into Modeless under `mixin`). Having initially agreed with your preference for Modeless in this case, now that it's up and running my preference has reverted to Modal, and that's the version I'll be using. It's cute to change the selection while the dialog is on-screen, but it only takes a handful of milliseconds to close a Modal, change selection and reopen the Modal (using KM for instance). I really hate the ambiguity of window focus in Modeless, especially after activating `show_notes_dialog()`.

So this version can go on the __repo__ if you like, along with the plethora of __undo__ events. I use (of course) the key equivalents (+/- with fast key repeats) to change values so simply tracking the slider-thumb won't help. Have been discussing this offline with @ThistleSifter and probably run out of other workarounds for a less wasteful __undo__ list.

There's been some good feedback on the original Modal version on FB (well ... one positive review!). Not sure it's a good idea to post __both__ _Modal_ and _Modeless_ versions.

rpatters1:
GIven the ambiguity, I think an option would preferable. It's literally only the difference between calling RunModeless and ExecuteModal (plus updating the undo handling.)

One of the things I dislike most about the modal version is that the selected region vanishes when the dialog opens.

FWIW: it looks like adding a mouse tracking started/stopped notification is going to be fairly easy. That will considerably reduce the number of undo entries with modeless.

jwink75:
+1 vote for an option! Maybe just a simple checkbox whose state is retained. For what it's worth, Carl, I fully intend to make one for the Page Format Wizard so that it doesn’t automatically switch units… I happen to like that feature but totally recognize that others don’t :)

cv-on-hub:
> I think an option would preferable.

Are you suggesting posting both? I'd rename this one `rest_slider_modeless`. (In this case the conversion to `modeless` required jumping through multiple tricky hoops that I wouldn't replicate in a `modal` version).

jwink75:
Couldn’t it be done in the same script? Something like:

if run_modeless_check:GetCheck() == 1 then
	finenv.RegisterModelessDialog(dialog)
	dialog:ShowModeless()
else
	dialog:ExecuteModal(nil) == finale.EXECMODAL_OK
end

rpatters1:
I'm suggestion one script with the option to be modal or modeless. The difference between the two (with mixins) is diminimous.

rpatters1:
With mixins it is simpler that that, @jwink75:

```lua
if modeless_option then
	dialog:RunModeless()
else
	dialog:ExecuteModal()
end
```

There will also need to be code to suppress undo handling if it is modal.

cv-on-hub:
OK. A few more hoops to jump through converting back to modal (... all that `timer` stuff) but not too arduous. __OPTIONAL__ modeless follows. 

PS @jwink75: Page Format Wizard is magical and I have no problem changing Measurement Units - it was the fact that changing units in some areas changed the VALUES for data in a different area!

rpatters1:
Looks excellent to me. When 0.72 comes out we can add mouse tracking events and reduce the number of modeless undo events. But that's gotta wait for 0.72.

FYI: there is currently no continuous update on Windows. Just the final value when the mouse is released. That will change with 0.72 so that both report continuously during mouse scrolling.

rpatters1:
Let me know if you are ready, and I'll merge it. (Probably later tonight.)

cv-on-hub:
> Let me know if you are ready, and I'll merge it.

Great. Just a few tidying things and a few less conditionals. All ready to go.

PR #698, opened 2024-02-15
Rewrite metadata parser in Lua

### LINKED ISSUE #695: [BUG] Localized scripts not showing up correctly on website
When scripts take advantage of the new localization features, the metadata extractor is not able to grab things correctly for use on the website.

Although #689 refers specifically to the bundler, it may be that solving this current issue requires rewriting the metadata extractor in Lua, since it's no longer a matter of simply parsing out strings from the `.lua` file.

cc: @rpatters1

asherber:
Fixes #695 

The main part of the action is still in JS. It sends the `plugindef()` function and a "parser" function to the Lua interpreter, which calls `plugindef()` and sends back a JSON version of its return values and the `finaleplugin` properties. Then we do some additional manipulation in JS and save the JSON.

My division of labor is basically that Lua is only invoked to interpret `plugindef()` and return the metadata; anything else (e.g., date formatting, forcing two decimal places on Min/MaxJWLuaVersion, etc.) is done back in JS.

I added some details about the process to the action's README, and I tried to make sure that the code is readable (and commented appropriately if something is unclear). I'm happy for suggestions if anything can be made clearer.

I confirmed through manual diffing that the output of the rewritten action matches the output of the current action. (Actually, there are a few differences -- places where the current action is buggy!)

### DIFF
--- a/.github/actions/get-metadata/README.md
+++ b/.github/actions/get-metadata/README.md
@@ -44,3 +44,10 @@ This is the author data from the script. It contains the following properties:
 - Return the captured data as a JSON object
 - Perform some manipulation on the JSON (dates, string arrays, etc.)
 - Build an array of the returned JSON and save to a file
+
+## Caution
+
+`MinJWLuaVersion` and `MaxJWLuaVersion` Min/MaxJWLuaVersion are defined as numbers in Lua,
+but they're expected to have two decimal places -- v0.70 is not v0.7. A function in `parse.ts`
+ensures that the values will display with two decimal places. If there's eveer a case where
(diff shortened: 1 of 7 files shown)


asherber:
(Because of the large-scale changes, I find it easier to digest the diff in Unified rather than Split format.)

rpatters1:
Look fantastic. The only suggestion I have is to prominently mention somewhere that the string conversion assumes 2 decimal places and that if there is ever a case where there are more or fewer, we will have to revisit this.

asherber:
Thanks, I'll make that clearer.

PR #699, opened 2024-02-16
Update hotkey_script_palettes.lua

cv-on-hub:
script_name from general_library; localised functions(); amended all secondary (modal) dialog foci; trimmed redundant vars; added Info button to “Configure” palettes; updated Notes for RTF.

### DIFF
--- a/src/hotkey_script_palettes.lua
+++ b/src/hotkey_script_palettes.lua
@@ -3,9 +3,9 @@ function plugindef()
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "https://carlvine.com/lua/"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "0.65"
+    finaleplugin.Version = "0.66"
     finaleplugin.LoadLuaOSUtils = true
-    finaleplugin.Date = "2024/02/08"
+    finaleplugin.Date = "2024/02/23"
     finaleplugin.CategoryTags = "Menu, Utilities"
     finaleplugin.MinJWLuaVersion = 0.70
     finaleplugin.Notes = [[ 
@@ -16,7 +16,7 @@ function plugindef()
 
         The _Hotkey Palette_ principle is demonstrated expertly by 
         Nick Mazuk on YouTube 
-        ([www.youtube.com/@nickmazuk](https://www.youtube.com/@nickmazuk)). 
+        ([www.youtube.com/@nickmazuk ](https://www.youtube.com/@nickmazuk)). 
         Scripts are grouped into primary categories like _Intervals_, _Layers_, 
         _Notes & Chords_, _Measure Items_ and so on as a set of palettes triggered by keystroke. 
         Each primary palette calls up a second palette containing scripts in related areas, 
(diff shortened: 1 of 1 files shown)


cv-on-hub:
I think this branch is now ready to merge.

src/hotkey_script_palettes.lua
```diff
@@ -3,83 +3,48 @@ function plugindef()
...
         easily configurable palettes (dialog windows) organised by type of activity 
         and triggered by simple "hotkey" keystrokes.
 
-        The "Hotkey Palette" principle is demonstrated expertly by Nick Mazuk at 
-        [https://www.youtube.com/@nickmazuk]. 
-        Scripts are grouped into primary categories like "Intervals", "Layers", 
-        "Notes & Chords", "Measure Items" and so on as a set of palettes triggered by keystroke. 
-        Each primary palette calls up a second palette containg scripts in related areas, 
+        The _Hotkey Palette_ principle is demonstrated expertly by 
+        Nick Mazuk [on YouTube](https://www.youtube.com/@nickmazuk).
```
asherber, line 18: Note that links in Markdown like this don't render in RTF as you might expect -- appears to be partly a function of how pandoc translates them and partly a function of how the native controls render the RTF (at least on Windows). As a compromise solution, I would recommend removing the brackets around the target text, so that the URL simply displays in parens.

I can send a screenshot later on of what this looks like on Windows.

cv-on-hub: > I can send a screenshot later on of what this looks like on Windows.

Yes please. The RTF looks fine through `pandoc` and through 'show_notes_dialog' on Mac, though sometimes needs a _hard_ space to force a gap after the hyperlink.

asherber: Well, I take it back -- your info window appears to visually render just fine, though I remember there being issues with links when I was testing the feature.

**However,** the links are not clickable, and they don't even display the URL on hover, which makes them both practically useless and uninformative.

image

(As a side note, looking at all of the text you have is making me realize that a little extra line spacing might be useful, at least on Windows. Does the text appear a bit cramped on Mac as well?)

cv-on-hub: > the links are not clickable, and they don't even display the URL

Well that's disappointing, because they both click and hover in Mac. But worth losing to honour platform equality. Line spacing is OK on Mac - in keeping with the font size, though a tad wider wouldn't hurt. 

RTF on Mac

asherber: @cv-on-hub Thanks for the screenshot. I'll open another thread if I can implement line spacing, so we can share how it looks on our respective platforms.

@rpatters1 Any thoughts about settings on the Windows control that would improve the hyperlink experience?

rpatters1:
Both OS controls should correctly render line spacing and other paragraph formatting that is built into the RTF. In addition, starting with RGP Lua 0.71, there is a [SetLineSpacing](https://pdk.finalelua.com/class_f_c_ctrl_text_editor.html#aa1b7bae0862b03434c24675581f3cb0a) method. Note the platform differences in how it works.

asherber:
I'm going to try adding the line spacing directly to the RTF, because I think the behavior may be more predictable. But the bigger issue on Windows is that links don't seem to be clickable. What's the name of the native control?

rpatters1:
On WIndows it's Rich Edit.

rpatters1:
I should add, it's WinAPI RIch Edit, not .Net.

rpatters1:
I made a relatively simply change to the Windows control, with a tip from @asherber. It now behaves the same on Win as on Mac, including changing the cursor. This will be available in 0.72 of RGP Lua.

rpatters1:
Also available will be the URL tooltips. These already exist on Mac, but (at least on my system) the delay before they appear is a bit long. On Win (with my changes) they will appear right away.

cv-on-hub:
> Also available will be the URL tooltips.

Fantastic. (And yes - they are sluggish on Mac). So should I reinstate the hyperlinks in anticipation?

PR #702, opened 2024-02-16
Update gracenote_slash.lua

cv-on-hub:
Many small changes but also establishing that my local `master` is workable again.`script_name` from general_library; added single note slash option; added “restore defaults” option, keystroke filtering and key commands; allow configuration without music selection; upgraded Notes for `utils.show_notes_dialog`; sundry minor improvements.


cv-on-hub:
PS: have left RTF hyperlinks intact in `plugin.Notes` until options explored with @asherber.

PR #705, opened 2024-02-21
Add `assert_table_argument_type` to `library.mixin_helper`

ThistleSifter:
I need table argument validation for something I'm working on, so I split it out from the multi string proxy function into its own function. Also took the opportunity to fix up some old indenting horrors from the autoformatter.

I'd appreciate it if someone could test this by making sure any multi string proxy method still works and throws errors as expected.


rpatters1:
It's working for me.

rpatters1:
Is this ready to be merged?

ThistleSifter:
Yes, it's ready.

PR #709, opened 2024-02-27
Update rest_slider.lua

cv-on-hub:
Fixed error when modelss->modal transition left an unclosed UndoBlock (active Tool left undefined!); combined the three data initialisation routines; neatened selection ID; allowed for empty staff names; modal->modeless transition is re-entrant.


rpatters1:
I'm merging this now, but you may be interested in using the new `RegisterMouseTracking...` methods in `FCCustomLuaWindow`. These could reduce the number of Undo blocks when running modeless.

cv-on-hub:
> you may be interested in using the new `RegisterMouseTracking...` methods

Yes - I saw this in 0.72 and will check it out.

PR #710, opened 2024-03-01
Update ostinato_maker.lua

cv-on-hub:
Converted to modal-modeless User Option; `plugin.Notes` changed to `utils.show_notes_dialog()`; `script_name` now derived from `general_library` instead of literal string.


cv-on-hub:
I've found some other selection region anomalies in this script. Pulling this PR until I work out a reliable solution.

PR #712, opened 2024-03-09
Create rest_minimiser.lua

cv-on-hub:
New script to delete any rests in the selection that serve no musical function.


cv-on-hub:
> Have you looked at `FCNoteEntryCell.ReduceEntries`?

I didn't know such a thing existed, and see that you just created it in RGP 0.68. Could reduce this script to a couple of lines! Thanks Robert ... will examine and simplify. (I had also never really noticed `FCNoteEntryCell` before - looks incredibly useful).

rpatters1:
The [version history](https://robertgpatterson.com/-fininfo/-rgplua/vershist.html) is usually worth perusing after each release. :-)

rpatters1:
FWIW: `FCNoteEntryCell` is what `eachentry` and `eachentrysaved` use under the covers. You might find it interesting to look at the [how those functions work](https://github.com/finale-lua/lua-source/blob/master/built-in-functions/rgplua_built_in_functions.lua).

We also added a `FCXNoteEntryCell` to the `mixin` library that allows you to cherry pick entries from an `eachentry` loop for later processing without risk of crashing Finale. (The built-in version has that risk.)

cv-on-hub:
> There is a lot of overlap with this script

This is an understatement of delightfully mammoth proportions. That script is gone and replaced with 30 terse line under a new name: `rhythm_reducer.lua`.

PR #715, opened 2024-03-23
Update pitch_changer.lua

cv-on-hub:
The last version of this script failed to correctly identify accidentals in all key signatures. The pitch-matching algorithm has been improved, but also a __Modeless__ option has been added. Also: `script_name` from `general_library` not literal; upgraded Notes for Markdown using `utils.show_notes_dialog`; sundry other improvements.


cv-on-hub:
Sorry - forgot that the MODELESS option was already subject to PR #713.

PR #721, opened 2024-04-09
Update rest_slider.lua

cv-on-hub:
Implemented `RegisterMouseTracking` in RGPLua 0.72 to reduce redundant UndoBlock registrations. Hurt my head at first, but then proved to be really simple and effective. Have left `HandlesUndo` false for Modal use, since the single default UndoBlock remains most useful in that mode.


cv-on-hub:
PS: many thanks @rpatters1 for the `slider_control.lua` sample file. Thanks to that I got this up and running in less than an hour.

PR #727, opened 2024-04-21
Create music_spacing_hack.lua

cv-on-hub:
This may not be suitable for the REPO. There's a couple of Music Spacing options that I change frequently for different spacing scenarios. This little hack uses a couple of hotkeys to do that very quickly, without a mouse and without navigating the whole Document → Document Options → Music Spacing → Avoid Collision of → menu/dialog system.


cv-on-hub:
Too many updates here. Submitting a fresh new unified PR.

PR #739, opened 2024-05-20
Four scripts finaleplugin.Notes update

cv-on-hub:
`finaleplugin.Notes` in these 4 scripts has been updated for better __Markdown__ rendering and __RTF__ display using `utils.show_notes_dialog()`. If they use config.get_user_settings() then derive `script_name` from `general_library` instead of literal string. Some other code amendments here and there.


rpatters1:
Rather than requiring a character value to be entered by number, you could display `FCUI:CreateSymbolDialog` to allow the user to choose it by the glyph. The `key_modes_manage.lua` script contains an example. This could be in addition to typing it in. I'm not going to hold up the merge for this, but it is something to consider for the future.

cv-on-hub:
> you could display `FCUI:CreateSymbolDialog`

Great idea. Will look into it.

PR #743, opened 2024-06-02
Improve inject() in bundler

asherber:
@rpatters1 Not sure if you wanted to review this.

The `inject()` function in the bunder, which inserts `HashURL` and `RTFNotes`, was assuming that there would only be one line in `plugindef()` that starts with `return` and one line that starts with `end`. This is not necessarily true, since a line of text in `Notes`, for example, could start with either of these. (And the `Notes` in the script I recently added did!)

This PR changes `inject()` to get the last matching token. There's still an edge case here if `plugindef()` does not include a `return` statement (since it's not technically required) but does have a line starting with `return` in `Notes`. But since this repo seems to have an unofficial requirement for a proper `plugindef()` function with a `return`, I'm not worried about this.

### DIFF
--- a/.github/actions/bundle/dist/index.js
+++ b/.github/actions/bundle/dist/index.js
@@ -6975,17 +6975,22 @@ const injectExtras = (name, contents) => {
     return parts.prolog + parts.plugindef + parts.epilog;
 };
 exports.injectExtras = injectExtras;
-const inject = (contents, injection) => {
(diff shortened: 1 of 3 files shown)


asherber:
BTW, I ran the bundler locally, and there were no changes to the generated dist files, as expected.

rpatters1:
I'm cool with it if it works.

PR #744, opened 2024-06-02
Create articulation_expression_swap

MuseCraft-Studio:
Replaces the selected articulation with the selected expression (or vice versa) for the full document if a region is not selected.

### DIFF
--- a/src/articulation_expression_swap.lua
+++ b/src/articulation_expression_swap.lua
@@ -1,0 +1,207 @@
+function plugindef()
+    finaleplugin.RequireSelection = false
+    finaleplugin.Author = "CJ Garcia"
+    finaleplugin.Copyright = "2024 MuseCraft Studio"
+    finaleplugin.Version = "1.0"
+    finaleplugin.Date = "5/27/2024"
+    finaleplugin.RevisionNotes = [[
+        May 27, 2024: Script work began with idea from Burt Goldstein
+        May 28, 2024: Version 1.0
+        ]]
+    finaleplugin.CategoryTags = "Articulation, Expression"
+    return "Articulation and Expression Swap", "Articulation and Expression Swap",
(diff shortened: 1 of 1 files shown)


rpatters1:
- This repo uses snake-case for function names as well as variable names.
- You should check for 0 on return from DisplayArticulationDialog and DisplayExpressionDialog. 0 means the user canceled out of the dialog.
- I'm guessing you only want to allow text expressions. That means you need to check the 2nd return value of DisplayExpressionDialog and reject the result if it is true. Otherwise, you need to track the 2nd return value through your code so that you actually retrieve the appropriate shape expression. (Remember that text and shape expressions are numbered separately starting from 1.)
- Be aware that both DisplayArticulationDialog and DisplayExpressionDialog allow the user to modify the articulation and expressions lists, respectively. Consider placing those calls in their own Undo block separate from the `run` function.
- Since you are using the string-valued initializer for `FCString` there should be an appropriate `MinJWLuaVersion`.

cv-on-hub:
Hi @MuseCraft-Studio. Hope you don't mind but I needed some coding recreation and have implemented some of @rpatters1 's suggestions to your script. (But not isolated __Display[X]Dialog__ in a separate Undo block). And done a bit of streamlining. Feel free, of course, to ignore this completely! 

[articulation_expression_swap.lua.txt](https://github.com/user-attachments/files/16061117/articulation_expression_swap.lua.txt)

rpatters1:
What is the status of this PR? Will there be updates?

MuseCraft-Studio:
I got the script updated with all of the feedback, but after I couldn't find comparable examples of the undo block for the articulation/expression dialog, I got a bit distracted with other things. I'm happy to just push Carl's version (which also doesn't have the undo blocks)

rpatters1:
I'd say push the changes. The separate undo block is optional.

PR #746, opened 2024-06-02
Create string_harmonics.lua

cv-on-hub:
Create diamond noteheads on the top note of dyads identified as viable string harmonics. I use this script __a lot__ and just noticed that its exact functionality is not available in the Repo. It is only provided in the full commercial version of TGTools, not the inbuilt version.

### DIFF
--- a/src/string_harmonics.lua
+++ b/src/string_harmonics.lua
@@ -3,33 +3,42 @@ function plugindef()
     finaleplugin.Author = "Carl Vine"
     finaleplugin.AuthorURL = "https://carlvine.com/lua/"
     finaleplugin.Copyright = "https://creativecommons.org/licenses/by/4.0/"
-    finaleplugin.Version = "0.62"
-    finaleplugin.Date = "2024/06/02"
+    finaleplugin.Version = "0.63"
+    finaleplugin.Date = "2024/07/21"
     finaleplugin.Notes = [[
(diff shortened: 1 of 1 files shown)


rpatters1:
You do not need the `allowable` table. My recommendation is to change the `allowed` definition as follows:

```lua
local allowed =
{
   [3] = 2,
   [4] = 2,
   [5] = 3,
   [7] = 4,
   --etc.
}
```
Then you can simply compare to `if allowed[midi_diff]`. But in addition, the values in the key value pairs are the displacement_diff values between the two notes, so the actual code would be:

```lua
local displacement_diff = highest.Displacement - lowest.Displacement
if allowed[midi_diff] and allowed[midi_diff] = displacement diff then
   .
   .
   .
end
```

This prevents augmented/diminished intervals from creating false positives.

With this change, I would skip mention of MIDI notes in the plugin notes and instead provide information about when to choose this script over the existing string harmonics scripts in the repo. Also maybe more specific menu item text to differentiate it from them.

FWIW: the script doesn't "create" a diamond notehead, it changes the existing top notehead to a diamond. The use of "create" confused me at first, because that's what the existing string harmonic scripts do: create the diamond notehead.

cv-on-hub:
That's a very cool solution. Revised script follows.

PR #747, opened 2024-06-17
Create slur_changer.lua

cv-on-hub:
Change the characteristics of every slur in the current selection. This functionality is buried within JWChange, but I thought was worth bringing into the light.


cv-on-hub:
In the latest revision I've noticed that the values for __SS_ONSTATE__ and __SS_OFFSTATE__ appear to be inverted in the `FCSmartShape` _Class Reference_. Finale seems to expect __SS_ONSTATE__ to be __0__, not __1__.

PR #748, opened 2024-06-21
Update music_spacing_hack.lua

cv-on-hub:
Added the __Manual Positioning__ options. Other minor tweaks.


rpatters1:
This seems like a pretty useful script. I would modify the name and description not to use the word "hack". Also, modify the notes to be a general description of what it does and how to use it rather than the personal references that are there now.

FWIW: I maintain a separate repo of personal quick-and-dirty scripts that augment my specific workflows. Not all scripts need to be maintained here. But this one seems appropriate to me.

cv-on-hub:
And so it shall be. I also keep some separate "private" scripts, as well as heavily tailored versions of other Repo scripts (including a couple of your Transposition scripts!).  Happy that this one seems generally useful. Revision follows.

PR #758, opened 2024-08-03
Minor Code Improvements

cv-on-hub:
Sundry minor code improvements to three __CV__ scripts.


rpatters1:
In my opinion, nagging for a selection when there is none is user-abusive behavior. Instead, it should silently either do nothing or process the whole document, whichever is the more appropriate behavior. The other transposition scripts choose the first option (silently do nothing), which I think is the appropriate choice for transposition.

I'm going to go ahead and merge this, but I hope you will consider this for future changes.

cv-on-hub:
> nagging for a selection when there is none is user-abusive behavior

Fair comment. I'll remove that in the next update.

PR #759, opened 2024-08-05
Update rest_slider.lua

cv-on-hub:
This script was plagued by ambiguous behaviour when modeless and the selected region or active layer changed. Has driven me crazy for weeks. Seems to all lead back to `RegisterMouseTracking()` and the unusual way that this script continuously updates data in the score. `MouseTracking` just had to go. (This necessitated a major restructure of the whole script).


rpatters1:
I would like to better understand the issue with mouse tracking, irrespective of whether this script has a modeless option. I will merge this for now, but a more detailed explanation, with code samples, would be helpful, if it isn't too much trouble.

rpatters1:
P.S. Feel free to respond privately.

PR #766, opened 2024-10-01
Update musicxml_massage_export.lua

cv-on-hub:
So I have a folder with 50 MusicXML export files and I thought it would be easier to add folder scanning to this script than to massage them individually. It wasn't, but at least it was more interesting. I tried hard to not break anything in the original script.

### DIFF
--- a/src/musicxml_massage_export.lua
+++ b/src/musicxml_massage_export.lua
@@ -4,13 +4,13 @@ function plugindef()
     finaleplugin.NoStore = true
     finaleplugin.Author = "Robert Patterson (folder scanning added by Carl Vine)"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "1.0.5"
+    finaleplugin.Version = "1.0.6"
     finaleplugin.Date = "October 2, 2024"
     finaleplugin.LoadLuaOSUtils = true
     finaleplugin.CategoryTags = "Document"
     finaleplugin.MinJWLuaVersion = 0.74
     finaleplugin.AdditionalMenuOptions = [[
-        Massage MusicXML Single File
+        Massage MusicXML Single File...
     ]]
     finaleplugin.AdditionalUndoText = [[
         Massage MusicXML Single File
@@ -38,7 +38,7 @@ function plugindef()
         Due to a limitation in the xml parser, all xml processing instructions are removed. These are metadata that neither
         Dorico nor MuseScore use, so their removal should not affect importing into those programs.
     ]]
-    return "Massage MusicXML Folder",
+    return "Massage MusicXML Folder...",
         "Massage MusicXML Folder",
         "Massage a folder of MusicXML files to improve importing to Dorico and MuseScore."
 end
--- a/src/musicxml_massage_export.lua
+++ b/src/musicxml_massage_export.lua
@@ -2,10 +2,11 @@ function plugindef()
     finaleplugin.RequireDocument = false
     finaleplugin.RequireSelection = false
     finaleplugin.NoStore = true
-    finaleplugin.Author = "Robert Patterson"
+    finaleplugin.Author = "Robert Patterson (folder scanning added by Carl Vine)"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "1.0.1"
-    finaleplugin.Date = "September 25, 2024"
+    finaleplugin.Version = "1.0.4"
+    finaleplugin.Date = "October 1, 2024"
+    finaleplugin.LoadLuaOSUtils = true
     finaleplugin.CategoryTags = "Document"
     finaleplugin.MinJWLuaVersion = 0.74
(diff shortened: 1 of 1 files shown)


rpatters1:
This was definitely on my to-do list for this script. Might I suggest:

- use `luafilesystem` instead of `osutils`. I added `luafilesystem` as a built-in library a few versions back. ChatGPT knows how to write code for `luafilesystem` (aka `lfs`) so it should not be a big deal to change it. That will make it platform-independent and also easily allow to:
- make it recursive into all subfolders, at least optionally
- I think I'd rather see single file or folder as separate menu options rather than a dialog, but I'm willing to be convinced otherwise. It depends on whether we need the dialog for other options (which I'd like to avoid, honestly.)

Beyond the scope of this PR, I have some further improvements in mind. I've discovered another deficiency in the Finale export that would be easy to fix. That is that floating rests in multiple layers get pegged to their current vertical positions in the xml. This wreaks havoc on the MuseScore import, at least. As I see it we could

- revert all rests to floating as an option in the script. (In the xml, this is simply a matter of removing the relevant sub-nodes of the `rest` node.)
- require the original final file to be open while we process the musicxml (which can be done by a Lua script) and then only revert the rests that are actually floating. (This would eliminate the need for an option in the script.)
- something else?

One of the reasons I would really like to see this implemented with `lfs` is that it also facilitates unzipping compressed `mxl` files and processing them as well. That is the other future enhancement I'd like to see.

Anyhow, please let me know what you think about all this. I'm curious which way you are going post-Finale, too.

rpatters1:
P.S. All you need to do to get `lfs` is:

```lua
local lfs = require('lfs')
```

Search for `lfs` on the [doc page](https://www.finalelua.com/docs/rgp-lua) for more info.

ThistleSifter:
> * require the original Finale file to be open while we process the musicxml (which can be done by a Lua script) and then only revert the rests that are actually floating. (This would eliminate the need for an option in the script.)

I like this option and it seems more future-proof, in case other strange issues are found. But would it be possible for the script to handle the musicxml export process too? I think it would be tidier to have the script handle the whole process from start to finish. "Export the musicxml beforehand but keep the original file open before running this script" seems rather kludge-y.

rpatters1:
The script could execute the menu option, but it would have no control over what the menu option does. It would depend on how the export options are set as to whether it would be a batch process, and there would be no connection between that processing and the processing of the resultant xml files.

At one point I looked into possibly reproducing the whole musicxml export in Lua, but that's a ridiculously huge task. It's not worth the effort. The open-source [Dolet plugin](https://github.com/MakeMusicInc/DoletSibelius) for Sibelius could be a guide for doing it, but it is 26.5k lines of code. (!)

rpatters1:
>Export the musicxml beforehand but keep the original file open before running this script" seems rather kludge-y.

I agree. I think a better approach is to only allow the script to run when do docs are open and then open and close each Finale doc as needed. The goal is to keep it batchable.

ThistleSifter:
> The script could execute the menu option, but it would have no control over what the menu option does.

Ah, oh well.


> I agree. I think a better approach is to only allow the script to run when no docs are open and then open and close each Finale doc as needed. The goal is to keep it batchable.

That sounds better. Would it expect the Finale file to be in the same directory and with the same name?

cv-on-hub:
I didn't know about __luafilesystem__. So easy to use!

What about three menu choices:

- Massage MusicXML single file
- Massage MusicXML single file (unfloat rests)
- Massage Folder of MusicXML

"Folder" still needs a dialog offering the use of a different directory, hence the option to NOT rename those files, plus the option to unfloat rests.

rpatters1:
Let's leave unfloating rests to a different PR. I don't see any reason to keep it separate if we are going to open the FInale files alongside, but I don't want to hold this up for it.

We might instead have a separate Configuration menu item which includes all the options. Don't some of your scripts work this way?

cv-on-hub:
A separate __Configuration__ menu is an option, but this remains problematic on Mac (at least on Sonoma+) because the standard file/directory OPEN dialogue doesn't have any __title__ and I can't see any way to alert the User to "_Now Choose a File_" or "_Now Choose a Folder_" or "_Choose a Different Output Folder_". This is why I've inserted "_Choose a Different Folder_" __AFTER__ "_Massage a Whole Folder_". 

Many of our users still don't like editing the source script, and think this script is more of a __single use__ item not warranting stored options.

rpatters1:
That's way too many options. (As you said, it's a 1-off.) It should work just like the MusicXML export does. If you batch them, they go in the same folder as the originals. One menu option to do a single file. One to do a folder. If absolutely necessary, one to do configure options, but I'm hoping we won't need options.

cv-on-hub:
No problem. I'll rework it with a second menu and remove the separate folder option. (Which I'll keep just for me!) The user chooses either single file or single directory.

rpatters1:
This looks excellent. Thank you. I only ask that you add an ellipsis to the menu text, since both options open a dialog.

cv-on-hub:
Sure thing.

rpatters1:
Thank you!

PR #769, opened 2024-10-05
Update musicxml_massage_export.lua

cv-on-hub:
Oops! I had copied `finaleplugin.ScriptGroupName` across as a placeholder and neglected to change the value to "MusicXML". So in `finalelua.com` this script took the place of "Staff Explode".

### DIFF
--- a/src/musicxml_massage_export.lua
+++ b/src/musicxml_massage_export.lua
@@ -4,8 +4,8 @@ function plugindef()
     finaleplugin.NoStore = true
     finaleplugin.Author = "Robert Patterson (folder scanning added by Carl Vine)"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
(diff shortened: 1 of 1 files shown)


rpatters1:
I'm in the midst of a major upgrade of the MusicXML massager script. I've found a number of bugs in the original because I did not fully understand the MusicXML format. (I still don't, but I understand it a lot better. 🤣 )

Anyway, I'm changing it to leave a log file of what it has done rather than a big alert message. My question is, do y'all think it is more useful to produce a single logfile with all the results or a logfile per xml file processed?

cv-on-hub:
I didn't think the alert message was particularly useful but appreciated the log file. If it's part of a whole-folder extraction then a single composite log file would be much better IMHO!

rpatters1:
As an aside, one thing I realized about both this and the doc options exporter is that neither needs Finale to be authorized to work.

PR #770, opened 2024-10-08
Massage MusicXML v2

rpatters1:
Bug fixes:

- Now correctly skips all notes in a chord so that a chord does not end up with the first note under the ottava and the remaining notes not under it. (This can cause havoc on a Dorico import.)
- Path name text encoding issues are resolved. (This was mainly a Windows issue.)

Enhancements:

- Refactored UI
- Folder search now recursively searches subfolders.
- Added a pass for removing "real" whole rests under fermatas.
- Added a pass for refloating rests that are floating in the original Finale document. (This requires the Finale document to be in the same folder with the `.musicxml` file and have the same file name.)

I am leaving support for compressed `.mxl` files to a future version.

I will wait for comments from @cv-on-hub  and @ThistleSifter before merging this, at least for a while.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ mobdebug.lua
 
 src/**/personal*
 dist/**/personal*
+personal/*
 
 .vscode/settings.json
 */**/.vscode
--- a/src/document_options_to_musescore.lua
+++ b/src/document_options_to_musescore.lua
@@ -4,8 +4,8 @@ function plugindef()
     finaleplugin.NoStore = true
     finaleplugin.Author = "Robert Patterson"
     finaleplugin.Copyright = "CC0 https://creativecommons.org/publicdomain/zero/1.0/"
-    finaleplugin.Version = "1.0"
-    finaleplugin.Date = "October 2, 2024"
+    finaleplugin.Version = "1.0.1"
+    finaleplugin.Date = "October 6, 2024"
     finaleplugin.CategoryTags = "Document"
     finaleplugin.MinJWLuaVersion = 0.75
     finaleplugin.Notes = [[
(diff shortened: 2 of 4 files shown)


cv-on-hub:
This script is excellent. (At least once I worked out that the new `library/utils.lua` revision was required!) Especially updating filenames while they're being processed. 

Just two suggestions ...
Add blank line to log file before each file is processed so Line 477 becomes
&nbsp; &nbsp; &nbsp; `log_message("\n\n***** START FILE PROCESSING *****")`

and make the __Close__ button respond to keystroke, which might be achieved most simply:
```
--dialog:CreateCancelButton("cancel")
dialog:CreateOkButton():SetText("Close")
  ...
--self:GetControl("cancel"):SetText("Close")
```

cv-on-hub:
Oh yes - also need to add `self:SetOkButtonCanClose(true)` inside `dialog:RegisterInitWindow`
(Clicking __OK__ appears to abort any remaining process the same as clicking __Cancel__)

rpatters1:
I can certainly do that about the Ok button, but are you not concerned that making it respond to the keyboard makes it too easy to accidentally abort the job? I kind of like that idea that you have to take the trouble to click the button.

cv-on-hub:
> I kind of like that idea that you have to take the trouble to click the button.

There is some truth there and it's not a matter of grave concern!

cv-on-hub:
PS: also noted that this iteration seems noticeably faster than version 1.0.

rpatters1:
It turns out the Esc key cancels the job. Maybe that requires a little more intention than the Enter key.

rpatters1:
I don't see how it could be faster. It is making three passes on the xml whereas the first iteration just made one. *And* it is opening a Finale document if needed. All of that should make it (marginally) slower. (Or a lot slower if the Finale doc is big: just opening it is what takes the most time.)

cv-on-hub:
I can't comment on the `MuseScore` or `utils` scripts but `massage` works, and looks good to me.

PR #772, opened 2024-10-14
Zip utilities

rpatters1:
Provides a start on zip utilities.

Current functions permit extracting `enigmaxml` from `musx` files. While there is not yet any script in the repo calling it, I have some private ones calling it. I anticipate we could add a script that calls it to the repo.

A future enhancement may be adding support for `.mxl` compressed files to the MusicXML Massager.


src/library/ziputils.lua
```diff
@@ -0,0 +1,225 @@
...
+% calc_gunzip_command
+
+Returns the platform-dependent command to gunzip a file. It can be passed
+to `luaosutils.process.execute`, which will then return the text directly.
+
+
+@ archive_path (string) platform-encoded path of source gzip archive.
+: (string) platform-encoded command string to execute.
+]]
+function ziputils.calc_gunzip_command(archive_path)
```
asherber, line 118: @rpatters1 You might consider using native Powershell commands instead of requiring 7z. Powershell can already deal directly with zip files using [Expand-Archive](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.archive/expand-archive?view=powershell-5.1). It takes a [few lines of code](https://securitytidbits.wordpress.com/2017/04/14/powershell-and-gzip-compression/comment-page-1/) to deal with gzip.

rpatters1:
If you would like to submit a PR, I would welcome it. I don't particularly like the 7z dependency. My Win experience predates PowerShell, so I never think to use it.

asherber:
What shell does `process.execute` use by default on Win? I note that your Mac commands are straighforward, but the Win commands are all prefixed with `cmd /c`, which almost implies that the commands are actually executing under Powershell.

asherber:
Oh, I see, you're calling `CreateProcessW`. Okay, I can work with that.

rpatters1:
Sorry, life happened.

Yes, it's using `CreateProcessW`. The reason I added it to luaosutils was to not have a cmd window popup in Windows, which is what happens with `os.execute`. (It also returns the output text, which `os.execute` doesn't.)

PR #775, opened 2024-10-18
Refactor ziputils

rpatters1:
- refactor `ziputils` into `ziputils` and `enigmaxml` libraries
- refactor both to use `client` library for managing dependencies like `luaosutils`
- refactor PowerShell commands to work with `io.popen` as well as `luaosutils.process.execute`
- PowerShell script for `calc_gunzip_command` now returns raw decompressed bytes in `stdout`.
- other minor changes as part of refactor

I wonder is the reason `"%s"` works now (and `& { %s }` doesn't) is because all the commands are on a single line now. (This is to accommodate `io.popen` which apparently doesn't work with embedded newlines.)

### DIFF
--- a/docs/rgp-lua/finenv-properties.md
+++ b/docs/rgp-lua/finenv-properties.md
@@ -82,6 +82,18 @@ end
 
 ---
 
+#### EmbeddedLuaOSUtils\* (read-only property)
+
+If this property is true, it signifies that the `luaosutils` library is embedded in this version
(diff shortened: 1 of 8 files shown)


asherber:
I don't think that's the issue, but I'll try some tests later, just for my own edification.

rpatters1:
FWIW: When I change it to `& { %s }`, the command emits the powershell help page.

rpatters1:
It works the same irrespective of calling `luaosutils` or `io.popen`. That is, the current version works and the `& { %s }` version emits the help page.

PR #783, opened 2025-03-03
Don't run workflow on an autopublish

asherber:


asherber:
Also bump dependencies, per Dependabot

